PR #1, opened 2015-01-22
fix method visibility

tenderlove:
`create_entity` and `rename_keys` should not be inside the
metaprogramming loop (since they don't need to be metaprogrammed).
Also, having `protected` inside the loop will impact the visibility of
methods on the next run


abonas:
@tenderlove  thanks!!

PR #2, opened 2015-01-22
use `send` instead of getting the method and calling it

tenderlove:
`send` will automatically call the method for us, we don't need to get
the method object


abonas:
@tenderlove good to know :)
thanks!!

PR #3, opened 2015-01-22
remove explicit local variable assignment

tenderlove:
conditionals return a value in Ruby, so we can omit the explicit local
variable assignment.


abonas:
thanks!

tenderlove:
No problem!! :smile:

PR #4, opened 2015-01-22
switch to functional style iteration

tenderlove:
`each_with_object` will iterate over a collection with a particular
object, then return that object.  That means we can manipulate a hash
and return it without assignments

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -119,15 +119,13 @@ module Kubeclient
 
    public
    def get_all_entities
-      result_hash = {}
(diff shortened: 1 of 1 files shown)


abonas:
thanks. what is the benefit of this change versus the original one?
(have to admit that to me it's less readable)

tenderlove:
@abonas usually in Ruby we try to stick to functional style when possible.  In this case `each_with_object` will return the hash that we passed to it, which means we don't have to explicitly declare the hash at the top of the method, then explicitly return at the bottom.  Essentially this buys us a more concise method definition.

PR #5, opened 2015-02-09
add support for watching entities

simon3z:


test/kubeclient_test.rb
```diff
@@ -3,6 +3,7 @@
 require 'webmock/minitest'
 require './lib/kubeclient'
 
+TEST_FILE_PATH = File.dirname(__FILE__)
```
abonas: why is this needed? how are you invoking tests?

simon3z: I am invoking it with:

    rake test

but obviously I don't want to embed fixed paths such as:

    json_watch_stream = File.new('test/watch_stream.json')

I prefer to discover the path and use:

    json_watch_stream = File.new(File.join(TEST_FILE_PATH, 'watch_stream.json'))

so it's possible to invoke rake from any path and there's no assumption on where the code, Rakefile and tests are.

lib/kubeclient.rb
```diff
@@ -8,9 +9,36 @@
 require 'kubeclient/replication_controller'
 require 'kubeclient/entity_list'
 require 'kubeclient/kube_exception'
-
+require 'kubeclient/watch'
 
 module Kubeclient
+
+  # HTTP Stream used to watch changes on entities
+  class WatchStream
```
abonas: afaik it's considered best practice to put one class per file - that is to put WatchStream in a separate file.
But I'd like to hear other opinions as well. @tenderlove ? @Fryguy ?

abonas:
can you also update the readme with usage example for watch?

lib/kubeclient.rb
simon3z: Many modules in the stdlib contain multiple classes per file. In rails instead it's discouraged/prohibited because it's breaking its own autoloading mechanism.

In our case I think it boils down to whether it belongs there or not logically: in my opinion it's part of the rest-api client engine that currently reside in that file.

Fryguy: Yes, it is very common practice for one class per file, generally in a subdirectory structure that matches the underscorized full namespace.  Then, usually a top file requires it.

So, Kubeclient::WatchStream generally lives in lib/kubeclient/watch_stream.rb and in lib/kubeclient.rb you would `require 'kubeclient/watch_stream'`

PR #11, opened 2015-02-13
Add Travis support and badges

Fryguy:
@abonas You will need to enable the hooks for Travis, Code Climate, and Gemnasium for those things to auto-run on push.  The Travis and Gemnasium badges won't work until you get those set up, which I can't do, since this repo is owned by you.  I can show you how if you need help.  I think I have a document on a wiki somwhere with the instructions that I can send you.


Fryguy:
Also, @abonas would you like the ManageIQ bot to monitor this repo?

abonas:
@Fryguy , thanks a lot for this. I am adding @simon3z here because afaik he also started to do some travis related work.

simon3z:
@Fryguy thanks for the badges, I was looking forward to add them as well. If you don't mind I'd prefer using the travis ci that I prepared last friday here (I just sent the PR):

https://github.com/abonas/kubeclient/pull/13

.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: ruby
+rvm:
+  - "1.9.3"
```
abonas: @Fryguy , @simon3z 
I think we'll have to limit the ruby version for this gem to be  2.0 and up (not 1.9.3) due to a change introduced in this pr:
https://github.com/abonas/kubeclient/pull/05/files#diff-78cbcd17c7ae3f02081ec1ce253bb73bR14
ruby 1.9.3 assumes path, not uri and hence running with ruby 1.9.3 fails with "NoMethodError: undefined method `empty?' for #<URI::HTTP:0x00000003054da0>"
see the diffs:
http://ruby-doc.org//stdlib-1.9.3/libdoc/net/http/rdoc/Net/HTTPGenericRequest.html#method-c-new
http://ruby-doc.org//stdlib-2.0/libdoc/net/http/rdoc/Net/HTTPGenericRequest.html#method-c-new

abonas:
> g



> would you like the ManageIQ bot to monitor this repo?

@Fryguy , what does the bot provide? in other words, what's the benefit to be monitored by it?

README.md
```diff
@@ -95,4 +100,4 @@ This client is tested with Minitest.
 Please run all tests before submitting a Pull Request, and add new tests for new functionality.
 
 Running tests: <br>
-`rake test`
\ No newline at end of file
+`rake test`
```
abonas: what's different in the last line (103) - why does it show up as a diff? I am looking at it and it looks the same to me :)

.travis.yml
abonas: I pushed a change to gemspec limiting ruby to 2.0 and up, and released a new gem version 0.1.2 to include that change as well all watches/rubocop/events changes

Fryguy:
> what does the bot provide? in other words, what's the benefit to be monitored by it?

@abonas you can see here: https://github.com/ManageIQ/miq_bot/blob/master/README.md

You may not want all of that, so I might have to update the bot to allow me to "choose" specific pieces.

README.md
Fryguy: My editor makes sure the last line of a file ends in a carriage return.

.travis.yml
Fryguy: Ok...thanks!

PR #12, opened 2015-02-13
DRY up exception handling

Fryguy:
As per Code Climate: F in Kubeclient::Client - Identical code found in five :resbody nodes (mass*5 = 475)

@abonas Please review...I couldn't run the specs because they are not working on master, but this is clean replace of the exception handling logic.

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -33,6 +33,12 @@ module Kubeclient
       RestClient::Resource.new(@api_endpoint)
     end
 
+    def handling_kube_exception
+      yield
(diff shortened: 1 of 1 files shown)


abonas:
@Fryguy what exception do you get when running tests on master? for me, 'rake test' passes

abonas:
I checked this locally and it runs fine, so I am merging it.

Fryguy:
@abonas  I was using Ruby 2.1, and that fails due to some scoping issues.  I think @simon3z fixed them in his Travis PR #13

PR #13, opened 2015-02-15
build: add travis continuous integration support

simon3z:


simon3z:
@abonas tomorrow I'll add more info on why I made some of the changes. That's why I haven't sent the PR on friday.

lib/kubeclient.rb
```diff
@@ -47,6 +47,8 @@ def get_resource_name(entity)
       end
     end
 
+    public
```
abonas: why the changes in this file are relevant for a travis ci pr?

simon3z: this fixes the get_/delete_/etc. visibility in ruby 2.1 (define_method scope is protected (2.1) vs public (2.0)... so I had to move this back here)

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: ruby
+rvm:
+  - 2.0
+  - 2.1
```
Fryguy: Be sure to use strings for these.  Travis treats "2.0" to mean "the latest of Ruby 2.0" whereas using a float value does not do that.

.travis.yml
```diff
@@ -0,0 +1,8 @@
...
+sudo: false
+cache: bundler
+script:
+  - bundle exec rake test rubocop
```
Fryguy: why is this needed?  I think you can remove the script section entirely.

lib/kubeclient.rb
Fryguy: This is what was causing me to fail locally on Ruby 2.1

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: ruby
+rvm:
+  - 2.0
+  - 2.1
+sudo: false
+cache: bundler
```
Fryguy: I totally forgot about these 2...definitely good to have those in there.

.travis.yml
simon3z: It seems fine to me:
https://travis-ci.org/simon3z/kubeclient/builds/50911341

ruby 2.0.0p598 (2014-11-13 revision 48408) [x86_64-linux]

Fryguy: If you look closely it says 2 instead of 2.0 for the  Ruby version.  This
is because of the String thing in the  yml.

.travis.yml
simon3z: Because travis was running only "test" and not "rubocop" (at least there's not trace of it in the console).

I'll fix the rake default target to execute rubocop as well.

.travis.yml
simon3z: If it's using latest 2.0, isn't it just cosmetic?

simon3z:
@abonas @Fryguy this should be ok now.

PR #15, opened 2015-02-16
Extract dynamically defined method content to real methods

Fryguy:
This is done to reduce the complexity score and to find where the real
issues may lie.

Flog scores:

Before:
```
   254.0: flog total
    16.9: flog/method average

   154.1: Kubeclient::Client#none
```

After:
```
   241.3: flog total
    11.5: flog/method average

    65.9: Kubeclient::Client#none
    27.1: Kubeclient::WatchStream#each     lib/kubeclient/watch_stream.rb:11
    23.1: main#none
    20.0: Kubeclient::Client#get_all_for_entity lib/kubeclient.rb:42
    13.6: Kubeclient::Client#update_for_entity lib/kubeclient.rb:96
```

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -48,7 +48,7 @@ module Kubeclient
       raise KubeException.new(e.http_code, JSON.parse(e.response)['message'])
     end
 
-    def get_all_for_entity(entity)
+    def get_entities(entity)
       # TODO: labels support
       # TODO: namespace support?
       response = handling_kube_exception do
@@ -66,13 +66,13 @@ module Kubeclient
       collection = EntityList.new(entity, resource_version)
 
       result['items'].each do |item|
-        collection.push(create_entity(item, entity))
+        collection.push(new_entity(item, entity))
       end
 
       collection
     end
 
-    def watch_all_for_entity(entity, resource_version = nil)
+    def watch_entities(entity, resource_version = nil)
       resource_name = get_resource_name(entity.to_s)
 
       uri = api_endpoint.merge(
@@ -93,21 +93,21 @@ module Kubeclient
       WatchStream.new(uri, options)
     end
 
-    def get_for_entity(entity, id)
+    def get_entity(entity, id)
       response = handling_kube_exception do
         rest_client[get_resource_name(entity) + "/#{id}"].get
       end
       result = JSON.parse(response)
-      create_entity(result, entity)
+      new_entity(result, entity)
     end
 
-    def delete_for_entity(entity, id)
+    def delete_entity(entity, id)
       handling_kube_exception do
         rest_client[get_resource_name(entity) + "/#{id}"].delete
       end
     end
 
-    def create_for_entity(entity, entity_config)
+    def create_entity(entity, entity_config)
       # to_hash should be called because of issue #9 in recursive open
       # struct
       hash = entity_config.to_hash
@@ -116,7 +116,7 @@ module Kubeclient
       end
     end
 
-    def update_for_entity(entity, entity_config)
+    def update_entity(entity, entity_config)
       id = entity_config.id
       # to_hash should be called because of issue #9 in recursive open
       # struct
@@ -131,7 +131,7 @@ module Kubeclient
 
     protected
 
-    def create_entity(hash, entity)
+    def new_entity(hash, entity)
       entity.classify.constantize.new(hash)
     end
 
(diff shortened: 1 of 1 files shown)


lib/kubeclient.rb
```diff
@@ -39,6 +39,73 @@ def handling_kube_exception
       raise KubeException.new(e.http_code, JSON.parse(e.response)['message'])
     end
 
+    def get_all_for_entity(entity)
```
abonas: method name is confusing imho

abonas:
@Fryguy can you please explain why does it reduce complexity? 
I am definitely supporting simplification of the longer methods, no doubt. But I don't see much difference in the majority of the methods, they were 3-5 lines anyhow, so why is this considered a simplification?

Fryguy:
@abonas complexity is basically defined by the number of code paths.  Generally the higher the complexity,  the less readable and the harder to test.

Most complexity algorithms consider calling a method low complexity.   It doesn't matter how complex the method actually is.   By defining a method,  you are creating a line of separation.   On the other side of the line,  you can test in isolation, and with a good method name make the calling code much more readable. 

So in this case,  by extracting out the guts of the methods,  we are

- isolating the code to a method which can be tested in isolation (and able to be analysed for complexity in isolation)
- separating the metaprogramming part from the "real" code part,  which helps readability
- there is a scope attached to each of those blocks previously, that is wider than is needed by the method.  That is, the blocks have access to variables they don't need from the class scope,  and can even modify them.  By isolating to a method, that scoping is eliminated. 

See also  http://blog.codeclimate.com/blog/2013/08/07/deciphering-ruby-code-metrics/

Fryguy:
Rebased the PR.

lib/kubeclient.rb
Fryguy: Yeah...do you have a better idea?  Maybe just `get_entities` to match the previous `get_pods`, `get_nodes`, etc.?

Fryguy:
And this is why I don't think rubocop should be part of pass/fail for the tests.

Fryguy:
@abonas Updated the code and renamed the extracted methods as you requested.

abonas:
thanks @Fryguy .
I have 2 suggestions to make this even more clear :)
1. wherever there's a "#{method_name}", it sounds imho more correct to call it "entity_name", because it's already part of defining a method name string and it has a prefix anyway.
2. wherever the newly defined methods have a parameter named entity, I suggest to rename 'entity' to 'entity_type'. because right now for example "get_entities(entity)" doesn't look clear, or "create_entity(entity, entity_config)" also makes wonder why both parameters are needed.
other than that it looks great.

Can you please also explain where/how do you run that flog score from? I probably missed that in between zillion checks and validations that were added recently.

abonas:
@Fryguy just making sure you saw my comments in this PR...

Fryguy:
Oh thanks @abonas.  Somehow I did miss your response.

Those are great suggestions...I'll push an update.  To keep it consistent I might also rename the constant ENTITIES to ENTITY_TYPES, as well as the corresponding variables like `ENTITY_TYPES.each do |entity_type|`.  Is that ok with you?

To run flog...

```
gem install flog
flog lib
```

`flog lib` runs flog against all Ruby files in the lib directory.

Fryguy:
@abonas Updated.  I renamed the ENTITIES constant to ENTITY_TYPES.  If you are not ok with that I'll undo it, but I actually think it's all clearer now.

abonas:
> ENTITIES to ENTITY_TYPES, as well as the corresponding variables like ENTITY_TYPES.each do |entity_type|. Is that ok with you?

sounds great

Fryguy:
Great, then this should be ready to go.

lib/kubeclient.rb
```diff
@@ -17,7 +17,7 @@ module Kubeclient
   # Kubernetes Client
   class Client
     attr_reader :api_endpoint
-    ENTITIES = %w(Pod Service ReplicationController Node Event Endpoint)
+    ENTITY_TYPES = %w(Pod Service ReplicationController Node Event Endpoint)
```
simon3z: We recently added Namespaces, is it missing now?

abonas: > We recently added Namespaces, is it missing now?

this PR is up to date to yesterday, namespaces were added today. it supposed to rebase with np on top of namespaces PR imho.

PR #16, opened 2015-02-17
client: add support for kubernetes endpoints

simon3z:


abonas:
according to this: https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/client/endpoints.go
there is no "delete" method.
is there any additional info / doc on endpoints that you are familiar with?

simon3z:
I haven't looked for any info/doc because I quickly found the information I was looking for (services => pods relation). I also sent a patch to kubernetes to make that relationship more evident:

https://github.com/GoogleCloudPlatform/kubernetes/pull/4482

abonas:
> I haven't looked for any info/doc because I quickly found the information I was looking for (services => pods relation). I also sent a patch to kubernetes to make that relationship more evident:
> 
> GoogleCloudPlatform/kubernetes#4482

I don't understand which actions can be available on endpoints api besides GET. 
Delete seems like not supported, 
Create - not clear why/how can someone create it not via service/pod configuration. 
If there are actions that are not supported, imho it's better to consider not to expose them in kubeclient as well.

simon3z:
I think we should adopt the same philosophy we use for other entities. We don't duplicate the validation logic in our code, we let it fail from the server side.

If someone attempts to create/delete an endpoint the relevant error will be returned by the server. Moreover If in the future these operations will be added in kubernetes we won't have to update our code.

abonas:
> I think we should adopt the same philosophy we use for other entities. We don't duplicate the validation logic in our code, we let it fail from the server side.
> 
> If someone attempts to create/delete an endpoint the relevant error will be returned by the server. Moreover If in the future these operations will be added in kubernetes we won't have to update our code.

the rest of the entities do expose all actions, and generally speaking I tend not to expose unsupported actions in client side, so it can give an added value and extra info to users since there is no api documentation on k8s side. but I'll merge this to not stall the rest of the PRs.

PR #17, opened 2015-02-17
Add support for watching multiple entities at once

simon3z:


lib/kubeclient.rb
```diff
@@ -18,9 +18,9 @@ class Client
     attr_reader :api_endpoint
     ENTITIES = %w(Pod Service ReplicationController Node Event)
 
-    def initialize(api_endpoint, version)
-      api_endpoint += '/' unless api_endpoint.end_with? '/'
-      @api_endpoint = api_endpoint + version
+    def initialize(uri, version)
```
abonas: why isn't this a separate PR?

simon3z:
I'll do this by incremental steps.

lib/kubeclient.rb
```diff
@@ -138,5 +129,51 @@ def all_entities
         result_hash[key_name] = send(method_name)
       end
     end
+
+    def watch(entities)
+      watch_endpoint = api_endpoint + '/watch/'
```
Fryguy: Prefer using URI, if you can...

lib/kubeclient.rb
```diff
@@ -138,5 +129,51 @@ def all_entities
...
+      options = { inactivity_timeout: 0, keepalive: true }
+
+      EventMachine.run do
+        multi_request = EventMachine::MultiRequest.new
+
+        entities.each do |entity, version|
```
Fryguy: So is each entity a separate connection?

lib/kubeclient.rb
```diff
@@ -138,5 +129,51 @@ def all_entities
...
+
+          unless version.nil?
+            uri.query = URI.encode_www_form('resourceVersion' => version)
+          end
+
+          request = EventMachine::HttpRequest.new(uri, options).get
+
+          request.errback do
+            EventMachine.stop
+            fail request.error
```
Fryguy: Prefer raise over fail

lib/kubeclient.rb
```diff
@@ -138,5 +129,51 @@ def all_entities
...
+                request.response_header.status,
+                request.response_header.http_reason
+              )
+            end
+          end
+
+          buffer = ''
+          request.stream do |chunk|
+            if request.response_header.successful?
+              buffer += chunk
```
Fryguy: Prefer << over += for string concatenation

lib/kubeclient.rb
```diff
@@ -138,5 +129,51 @@ def all_entities
...
+              while (notice = buffer.slice!(/.+\n/))
```
Fryguy: If multiple messages come through,  this might miss one,  I think, since the regex is greedy

test/watch_test.rb
```diff
@@ -13,28 +13,36 @@ def test_watch_pod_success
...
                  status: 200)
 
     client = Kubeclient::Client.new 'http://localhost:8080/api/', 'v1beta3'
-    client.watch_pods.with_index do |message, index|
-      assert_instance_of(WatchNotice, message)
-      assert_equal(expected[index]['type'], message.type)
-      assert_equal('Pod', message.object.kind)
-      assert_equal('php', message.object.metadata.name)
+
+    watch_enum = Enumerator.new do |x, y|
```
Fryguy: I don't understand using enumerator here... Why not a normal array?

Fryguy:
Doh I just noticed you closed this.   It seemed small enough for a PR to me so not sure why you closed it.

simon3z:
@Fryguy the problem is that em-http-request has terrible ssl support. At the moment ssl support is more important than multi-watch. I'm evaluating typhoeus as an alternative.

PR #26, opened 2015-03-01
add minimal support for labels

kazegusuri:
This supports only for `get` verb such as `get_pods`, `get_replication_controllers` and so on.

The usage is like following:
```
client = Kubeclient::Client.new 'http://localhost:8080/api/', 'v1beta1'
pods = client.get_pods(labels: 'app=redis-master')
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,10 @@ Examples:
 <br>
 `pods = client.get_pods`
 <br>
+You can get entities which have specific labels by specifying input parameter named `labels`: <br>
+`pods = client.get_pods(labels: 'name=redis-master')` <br>
+You can specify multiple labels and that returns entities which have both labels:  <br>
+`pods = client.get_pods(labels: 'name=redis-master,app=redis')`
 
 2. Get a specific node (and respectively: get_service "service id" , get_pod "pod id" , get_replication_controller "rc id" )
 <br>
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -53,11 +53,13 @@ module Kubeclient
       raise KubeException.new(e.http_code, JSON.parse(e.response)['message'])
     end
 
-    def get_entities(entity_type)
-      # TODO: labels support
+    def get_entities(entity_type, options)
+      params = {}
+      params['labels'] = options[:labels] if options[:labels]
+
       # TODO: namespace support?
       response = handling_kube_exception do
-        rest_client[get_resource_name(entity_type)].get # nil, labels
+        rest_client[get_resource_name(entity_type)].get(params: params)
       end
 
       result = JSON.parse(response)
@@ -165,8 +167,8 @@ module Kubeclient
       entity_name_plural = entity_name.pluralize
 
       # get all entities of a type e.g. get_nodes, get_pods, etc.
-      define_method("get_#{entity_name_plural}") do
-        get_entities(entity_type)
+      define_method("get_#{entity_name_plural}") do |options = {}|
+        get_entities(entity_type, options)
       end
 
       # watch all entities of a type e.g. watch_nodes, watch_pods, etc.


abonas:
Hi @kazegusuri, thanks for your contribution.
Couple of comments/questions on this PR:

1. the usage should be documented in README.md file, so it should be part of this PR too, the usage comment above is useful but it won't be seen by most of the users.

2. The example above talks about filtering by one label, but the parameter name is 'labels' in plural. How can a user filter by more than one label in the same request? Do you have any example?

3. It's great that you've added tests, can you please explain what are they testing? There aren't any assert statements there, and the label foo=bar used in those tests is not used in any of the synthetic inputs in the project, so what are they testing/validating?

4. Travis build failed as it shows above, please check what are the errors/validations that failed and fix as necessary.

kazegusuri:
> the usage should be documented in README.md file, so it should be part of this PR too, the usage comment above is useful but it won't be seen by most of the users.

updated.

> The example above talks about filtering by one label, but the parameter name is 'labels' in plural. How can a user filter by more than one label in the same request? Do you have any example?

Just specify multiple key/values with comma sperated like `name=foo,app=bar`. README.md is also updated about multiple labels.

> It's great that you've added tests, can you please explain what are they testing? There aren't any assert statements there, and the label foo=bar used in those tests is not used in any of the synthetic inputs in the project, so what are they testing/validating?

Filtering by labels are done by apiserver, so all we have to do is just pass labels as a query parameter.
The test checks whether the query parameter has the labels. There is no assertion in the test, but it properly fails if the query parameter is wrong.

> Travis build failed as it shows above, please check what are the errors/validations that failed and fix as necessary.

My first commit failed because of some rubocop rules. I fixed it, so now it's right.

README.md
```diff
@@ -105,6 +105,13 @@ It is possible to interrupt the watcher from another thread with:
 <br>
 `watcher.finish` <br>
 
+8. Get entities filtered by labels <br>
```
abonas: I think this section is great, I think it suits better to be part of item 1 in the readme where get_all_X is explained, rather then being number 8. it will increase the chances of clarity and it's more relevant there as it's an additional option for GET rather than a separate action.

abonas:
> There is no assertion in the test, but it properly fails if the query parameter is wrong.

Please explain this - what do you mean by a wrong query parameter (labe instead of labels? wrong format in the value?) , and how does it fail the test? which error? I don't understand how this tests the functionality ( I agree that the actual filtering is done on k8s side hence there's no point to test it on kubeclient side).

abonas:
Hi @kazegusuri  have you seen the comments/are you planning to resubmit the patch?

kazegusuri:
Sorry for late. I have no time for now, I will take a look at this weekend.

README.md
kazegusuri: Done. Please take a look.

kazegusuri:
> the value?) , and how does it fail the test? which error? I don't understand how this tests the functionality ( I agree that the actual filtering is done on k8s side hence there's no point to test it on kubeclient side).

It is possible to break the behavior of the client when someone change a sourcecode.  For example, if rewrite to `param` as a parameter name instead of `params` in [this line](https://github.com/abonas/kubeclient/pull/26/files#diff-c8c777a3697ea654c558a52712390632R81), the unit test fails and displays following messages:

```
  1) Error:
ServiceTest#test_get_with_labels:
WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET http://localhost:8080/api/v1beta1/services with headers {'Accept'=>'*/*; q=0.5, application/xml', 'Accept-Encoding'=>'gzip, deflate', 'Param'=>'{"labels"=>"foo=bar"}', 'User-Agent'=>'Ruby'}

You can stub this request with the following snippet:

stub_request(:get, "http://localhost:8080/api/v1beta1/services").
  with(:headers => {'Accept'=>'*/*; q=0.5, application/xml', 'Accept-Encoding'=>'gzip, deflate', 'Param'=>'{"labels"=>"foo=bar"}', 'User-Agent'=>'Ruby'}).
  to_return(:status => 200, :body => "", :headers => {})

registered request stubs:

stub_request(:get, "/\/services/ with query params {"labels"=>"foo=bar"}")

============================================================
    /home/sabo/work/client/vendor/bundle/ruby/2.1.0/gems/webmock-1.20.4/lib/webmock/http_lib_adapters/net_http.rb:114:in `request'
    /home/sabo/work/client/vendor/bundle/ruby/2.1.0/gems/rest-client-1.7.3/lib/restclient/request.rb:270:in `net_http_do_request'
    /home/sabo/work/client/vendor/bundle/ruby/2.1.0/gems/rest-client-1.7.3/lib/restclient/request.rb:418:in `block in transmit'
    /home/sabo/work/client/vendor/bundle/ruby/2.1.0/gems/webmock-1.20.4/lib/webmock/http_lib_adapters/net_http.rb:123:in `start_without_connect'
    /home/sabo/work/client/vendor/bundle/ruby/2.1.0/gems/webmock-1.20.4/lib/webmock/http_lib_adapters/net_http.rb:150:in `start'
    /home/sabo/work/client/vendor/bundle/ruby/2.1.0/gems/rest-client-1.7.3/lib/restclient/request.rb:413:in `transmit'
    /home/sabo/work/client/vendor/bundle/ruby/2.1.0/gems/rest-client-1.7.3/lib/restclient/request.rb:176:in `execute'
    /home/sabo/work/client/vendor/bundle/ruby/2.1.0/gems/rest-client-1.7.3/lib/restclient/request.rb:41:in `execute'
    /home/sabo/work/client/vendor/bundle/ruby/2.1.0/gems/rest-client-1.7.3/lib/restclient/resource.rb:51:in `get'
    /home/sabo/work/client/lib/kubeclient.rb:85:in `block (3 levels) in <class:Client>'
    /home/sabo/work/client/lib/kubeclient.rb:46:in `handling_kube_exception'
    /home/sabo/work/client/lib/kubeclient.rb:84:in `block (2 levels) in <class:Client>'
    /home/sabo/work/client/test/service_test.rb:40:in `test_get_with_labels'
```

Actually I'm not sure this should be tested, so I will happy to remove the unit test if you think no need.

kazegusuri:
Rebased to HEAD.

abonas:
> Actually I'm not sure this should be tested, so I will happy to remove the unit test if you think no need.

it sounds that it tests the server rather than the client, whether the server disregards a param it is not familiar with or gives an error, but since the tests here are unitests at the moment with no live env, imho testing this is not relevant.

so I recommend to remove the tests.
cc @simon3z

kazegusuri:
Removed tests. Please take a look.

abonas:
> Removed tests. Please take a look.

looks great

PR #29, opened 2015-03-04
define entity class definitions dynamically

abonas:


abonas:
@simon3z @Fryguy  what do you think of this change? I think it looks cleaner and more flexible looking forward

simon3z:
LGTM

Fryguy:
LGTM.   In the future this might go away so you have a place to hang action methods like an update method in a more OO way,  but for now its good

abonas:
@Fryguy can you elaborate more about what you wrote above re action methods/or an example?
(I'll merge this regardless, it doesn't block)

Fryguy:
I just noticed a bug here.   It should be Kubeclient. const_set.   Otherwise this puts the classes in the top level scope in stead of in the Kubeclient namespace.  You can also just call const_set directly with no receiver, since you are in that scope.

Fryguy:
Regarding the action method comment,  consider the current usage... 

```ruby 
pod = client.get_pod(id)
client.update_pod(pod, :name => "foo") 
```

Which is very procedural/functional like ... versus 

```ruby 
pod = client.get_pod(id)
pod.update(:name =>" foo") 
```

Or perhaps 

```ruby 
pod = Pod.find(id)
pod.update(:name =>" foo") 
```

Which is more OO and more Rubyish .

PR #32, opened 2015-03-05
add test helper

abonas:


abonas:
cc @Fryguy

test/kubeclient_test.rb
```diff
@@ -1,7 +1,4 @@
-require 'minitest/autorun'
-require 'json'
-require 'webmock/minitest'
-require './lib/kubeclient'
+require './test/test_helper'
```
Fryguy: No need for the ". /"

Fryguy:
Looks good to me other than the very minor comment.   :+1:

test/kubeclient_test.rb
abonas: @Fryguy 
it doesn't work without.
I tried "test/test_helper" or just "test_helper"
when doing rake test it gets "cannot load such file -- test/test_helper" error

Fryguy: Interesting, so here's what I think is happening.  The `require` method doesn't actually work with relative paths.  That relative path is relative to your current working directory, so the reason is works is because you are running the tests from the root.  If you want to use relative paths from the file, use `require_relative`, which would make these `require_relative './test_helper'` or even, I think just `require_relative 'test_helper'`.

However, you shouldn't need all that.  The test directory should be on the load path, and you should just be able to do `require 'test_helper'` all by itself.  I think the reason it's not working is because of how the [test task is defined in the Rakefile](https://github.com/abonas/kubeclient/blob/master/Rakefile#L7-L9), which should be using a Rake::TestTask, like so:

```ruby
Rake::TestTask.new do |t|
  t.libs << "test"
end
```

Then in the test files you can just do `require 'test_helper'`

abonas: > Rake::TestTask.new do |t| t.libs << "test" end

@Fryguy thanks for all the help.
I tried to do the above in Rakefile, and switched to use require 'test_helper' in files, however now when running rake test there are 0 tests running.
(0 runs, 0 assertions, 0 failures, 0 errors, 0 skips)
I'm checking Rake docs to see what's missing...

Fryguy: See http://ruby-doc.org//stdlib-2.0/libdoc/rake/rdoc/Rake/TestTask.html .  You may need to set the testfiles...I thought that was done automatically, but maybe not?

Fryguy: Yeah, it *should* default properly... See https://github.com/ruby/ruby/blob/v2_0_0_643/lib/rake/testtask.rb#L90

abonas: > Yeah, it should default properly... See https://github.com/ruby/ruby/blob/v2_0_0_643/lib/rake/testtask.rb#L90

found the problem...  with the default rakefile configuration, test files names must start with "test", in kubeclient they ended with "test", hence renaming solved the problem...

abonas:
@Fryguy @simon3z pls have a look on the last commit before I merge.

Fryguy:
Looks great! :+1:

Fryguy:
I'll wait until this is in before fixing #33

PR #33, opened 2015-03-09
Ensure dynamic classes live in the Kubeclient namespace.

Fryguy:
@abonas Please review.

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -7,6 +7,7 @@ require 'kubeclient/kube_exception'
 require 'kubeclient/watch_notice'
 require 'kubeclient/watch_stream'
 
+# A Ruby client for Kubernetes REST api.
 module Kubeclient
   ENTITY_TYPES = %w(Pod Service ReplicationController Node Event Endpoint
                     Namespace)
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -8,21 +8,21 @@ require 'kubeclient/watch_notice'
 require 'kubeclient/watch_stream'
 
 module Kubeclient
+  ENTITY_TYPES = %w(Pod Service ReplicationController Node Event Endpoint
+                    Namespace)
+
+  # Dynamically creating classes definitions (class Pod, class Service, etc.),
+  # The classes are extending RecursiveOpenStruct.
+  # This cancels the need to define the classes
+  # manually on every new entity addition,
+  # and especially since currently the class body is empty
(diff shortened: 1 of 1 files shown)


Fryguy:
Technically #29 implemented the dynamic classes properly, since the original classes were not namespaced.  However, a gem *should* namespace its own classes.

abonas:
@Fryguy looks great to me, thanks a lot for the patch. 
Just a small question - is it OK not mentioning the module name when accessing ENTITY_TYPES inside Kubeclient class, now that it moved out of the class? I know it works, but is it a good/safe practice?

I think rubocop found a minor thing, pls fix it so it will pass.

Fryguy:
Do you have any suggestions for the module comment...I was thinking of just taking what's in the README.  Incidentally, I think a round of documentation might be a good idea at some point...would help flesh out http://www.rubydoc.info/gems/kubeclient

Fryguy:
> Just a small question - is it OK not mentioning the module name when accessing ENTITY_TYPES inside Kubeclient class, now that it moved out of the class? I know it works, but is it a good/safe practice?

Yes it is fine.  It is accessible by scope, so it works, and it is not uncommon to do it this way.

Fryguy:
well that's weird...I updated the comment and the tests are now failing?!  even locally that happens...I think the rakefile needs an update...I'll do that in a separate PR.

For now, I'll try to make the tests pass here.

abonas:
> Do you have any suggestions for the module comment...I was thinking of just taking what's in the README. Incidentally, I think a round of documentation might be a good idea at some point...would help flesh out http://www.rubydoc.info/gems/kubeclient

I'm always pro documentation, what do you have in mind/ what did you mean by fleshing out http://www.rubydoc.info/gems/kubeclient? README is detailed enough, no? unless I misunderstood you.

Fryguy:
Well, documentation at the class level and method level would be helpful to people reading the code.  rubydoc.info reads those docs and generates nice pages.

abonas:
> Well, documentation at the class level and method level would be helpful to people reading the code. rubydoc.info reads those docs and generates nice pages.

yes, sure, any reference to a good example/convention of doing that?
are you planning to fix this current PR so it can be merged?

Fryguy:
Yeah working on it... The are a lot of namespacing oddities it turns out.

abonas:
> Yeah working on it... The are a lot of namespacing oddities it turns out.
> …

ok, pls share when you understand what's going on there :)

Fryguy:
Closing in favor on #39

PR #37, opened 2015-03-16
switch EntityList from an Array subclass to a delegate

tenderlove:
subclassing Array in Ruby will cause problems down the line, so lets use
a delegate class now

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -70,13 +70,11 @@ module Kubeclient
           result.fetch('metadata', {}).fetch('resourceVersion', nil)
       end
 
-      collection = EntityList.new(entity_type, resource_version)
-
-      result['items'].each do |item|
-        collection.push(new_entity(item, entity_type))
+      collection = result['items'].map do |item|
+        new_entity(item, entity_type)
       end
 
-      collection
+      EntityList.new(entity_type, resource_version, collection)
     end
 
     def watch_entities(entity_type, resource_version = nil)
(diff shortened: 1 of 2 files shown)


abonas:
> subclassing Array in Ruby will cause problems down the line, so lets use
> a delegate class now

@tenderlove thanks of the patch.
I'd appreciate an explanation on what kind of problems can the current impl cause to get a better picture on this patch. thanks.

tenderlove:
The subclass won't always behave the way you want it to.  The C implementation will bypass certain methods.  For example:

```ruby
class Foo < Array
  def initialize
    puts "called"
    super
  end
end

x = Foo.new
y = x.slice(0, 1)

p [x.class, x.object_id]
p [x.class, y.object_id]
```

This code allocates two instances of `Foo`, but the constructor `initialize`, is only called once.  This means you can't set up your object correctly in the case that someone calls `slice` on it.  It's much easier to start out with a restrictive API and add methods as they are needed than to deprecate methods.

tenderlove:
[Here's a blog post with more info](http://words.steveklabnik.com/beware-subclassing-ruby-core-classes)

Fryguy:
This is, personally, the One Thing&trade; I hate about Ruby.  Completely breaks the Ruby-is-pure-OO concept. :frowning:

abonas:
> Here's a blog post with more info

very informative, thanks a lot

abonas:
> This is, personally, the One Thing™ I hate about Ruby. Completely breaks the Ruby-is-pure-OO concept. :frowning:

+1

PR #39, opened 2015-03-16
move dynamically defined classes under `Kubeclient`

tenderlove:
We shouldn't define classes in the top level namespace, since other
people may want to use those names.  This patch moves dynamically
defined constants under `Kubeclient` so that the library is friently
with other code bases.

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -12,16 +12,14 @@ module Kubeclient
   class Client
     attr_reader :api_endpoint
 
-    ENTITY_TYPES = %w(Pod Service ReplicationController Node Event Endpoint
-                      Namespace)
-
     # Dynamically creating classes definitions (class Pod, class Service, etc.),
     # The classes are extending RecursiveOpenStruct.
     # This cancels the need to define the classes
     # manually on every new entity addition,
     # and especially since currently the class body is empty
-    ENTITY_TYPES.each do |entity_type|
-      Object.const_set(entity_type, Class.new(RecursiveOpenStruct))
+    ENTITY_TYPES = %w(Pod Service ReplicationController Node Event Endpoint
+                      Namespace).map do |et|
+      [Kubeclient.const_set(et, Class.new(RecursiveOpenStruct)), et]
     end
 
     def initialize(uri, version)
@@ -53,7 +51,7 @@ module Kubeclient
       raise KubeException.new(e.http_code, JSON.parse(e.response)['message'])
     end
 
-    def get_entities(entity_type, options)
+    def get_entities(entity_type, klass, options)
       params = {}
       params['labels'] = options[:labels] if options[:labels]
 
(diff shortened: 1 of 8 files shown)


Fryguy:
@tenderlove This is what I was doing in #33 , but you seem to have finished it before me, so closing that one.

tenderlove:
@Fryguy oops, sorry about that.  I should have read the other PRs.

Fryguy:
No problem! :smiley_cat:

lib/kubeclient.rb
```diff
@@ -195,7 +193,7 @@ def ssl_options(client_cert: nil, client_key: nil, ca_file: nil,
     end
 
     def all_entities
-      ENTITY_TYPES.each_with_object({}) do |entity_type, result_hash|
+      ENTITY_TYPES.each_with_object({}) do |(_, entity_type), result_hash|
```
abonas: Why there was a need to add here this part? (_, entity_type)

test/test_kubeclient.rb
```diff
@@ -106,11 +106,11 @@ def test_get_all
     assert_instance_of(EntityList, result['pod'])
```
abonas: Does it make sense to add also Kubeclient:: before EntityList now? Its scope should also be limited to Kubeclient, but the only difference is that it was not created dynamically

lib/kubeclient.rb
```diff
@@ -12,16 +12,14 @@ module Kubeclient
...
     # Dynamically creating classes definitions (class Pod, class Service, etc.),
     # The classes are extending RecursiveOpenStruct.
     # This cancels the need to define the classes
     # manually on every new entity addition,
     # and especially since currently the class body is empty
-    ENTITY_TYPES.each do |entity_type|
-      Object.const_set(entity_type, Class.new(RecursiveOpenStruct))
+    ENTITY_TYPES = %w(Pod Service ReplicationController Node Event Endpoint
+                      Namespace).map do |et|
+      [Kubeclient.const_set(et, Class.new(RecursiveOpenStruct)), et]
```
abonas: basically this converts the ENTITY_TYPES array to an array of arrays where each entry is a class->entity type, right?
I am just curious (is this a best practice?) why did you choose to implement it like this, and not for example as a hash that holds key entity type to its klass as a value?
thanks

tenderlove: No reason.  We could do a hash too, but I didn't see a place where we needed to do a hash lookup.

test/test_kubeclient.rb
tenderlove: Yes, I didn't change it yet because I sent the other PR.  But now that it's merged, I can fix this.

lib/kubeclient.rb
tenderlove: `ENTITY_TYPES` is a list of lists.  `(_, entity_type)` deconstructs the list.  `_` is the class we made above, and `entity_type` is a string.

abonas:
@tenderlove , thanks for the commits, there was a tiny rubocop issue, I fixed it in a separate PR. all good!

PR #42, opened 2015-03-18
Fix #38 add default path and version

### LINKED ISSUE #38: Add defaults when kubeclient is initialized
Based on:
https://github.com/ManageIQ/manageiq/pull/2156#issuecomment-81717931

Add defaults if they weren't passed from client:
1. /api suffix
2. apiversion

abonas:


abonas:
@simon3z  / @Fryguy   pls review.
@Fryguy  -  this is a first step to pass less info on k8s provider side in manageiq when initializing the client

lib/kubeclient.rb
```diff
@@ -22,8 +22,12 @@ class Client
       [Kubeclient.const_set(et, Class.new(RecursiveOpenStruct)), et]
     end
 
-    def initialize(uri, version)
+    def initialize(uri, version = 'v1beta3')
       @api_endpoint = (uri.is_a? URI) ? uri : URI.parse(uri)
+
+      unless @api_endpoint.path.include? 'api'
+        @api_endpoint.merge!(File.join(@api_endpoint.path, 'api'))
```
tenderlove: You probably don't want this because `File.join` will use the default file separator which is `\` on Windows.  I would just use string interpolation.

abonas: > You probably don't want this because File.join will use the default file separator which is \ on Windows. I would just use string interpolation.

good point, thanks !
I assume the same goes for line 31 too (which is not part of this patch though)

BTW from looking at the documentation of the method it's really unclear...

Fryguy: Actually since you have a URI object,  perhaps you can just use it,  which
would be the most accurate.

abonas: > Actually since you have a URI object, perhaps you can just use it, which would be the most accurate.

what do you mean?
user could just pass host:port as part of the uri, without "/api" and without the version.
in the case "/api" and version weren't passed, they need to be appended to the path in uri, no?

Fryguy: I guess I meant I don't think you should append.  I think you should only
default to /api if they didn't pass a path at all.   If they passed a
path,  then that is the caller stating they know better, and they have
declared a different endpoint.

abonas: > I guess I meant I don't think you should append. I think you should only default to /api if they didn't pass a path at all. If they passed a path, then that is the caller stating they know better, and they have declared a different endpoint.

ok now I get it... set path to "/api" if path is empty. makes sense!

test/test_kubeclient.rb
```diff
@@ -20,6 +20,37 @@ def test_json
     assert_equal(expected, JSON.parse(JSON.dump(our_object.to_h)))
   end
 
+  def test_pass_uri
+    stub_request(:get, /\/services/)
+      .to_return(body: open_test_json_file('entity_list_b1.json'),
+                 status: 200)
+    uri = URI::HTTP.build(port: 8080)
```
simon3z: No need for lines 24-26 if your assert is the only assert is line 33

PR #45, opened 2015-03-20
Remove support of v1beta1

abonas:
Removed support of v1beta1 due to its deprecation soon on k8s side.
Deletion of all v1beta1 related tests, replacing them with v1beta3
examples.
Updated README as well.


abonas:
@simon3z please review

Issue #40, opened 2015-03-16
Add a "connection check" method on api root


abonas:
Need to add a connection validation call on the root of the api
Based on https://github.com/ManageIQ/manageiq/pull/2156#issuecomment-81778779

PR #47, opened 2015-03-21
Add method to retrieve the api versions from server

abonas:


abonas:
@Fryguy this patch can solve #40 IMO.

abonas:
@simon3z / @Fryguy pls review

lib/kubeclient.rb
```diff
@@ -202,5 +202,11 @@ def all_entities
         result_hash[key_name] = send(method_name)
       end
     end
+
+    def api_versions_on_server
```
simon3z: Probably "server_api_versions" is shorter/nicer.

abonas: > Probably "server_api_versions" is shorter/nicer.

:+1:

lib/kubeclient.rb
```diff
@@ -202,5 +202,11 @@ def all_entities
...
+      versions_endpoint = URI.parse(@api_endpoint.to_s)
+      versions_endpoint.path.slice! @api_version
+      RestClient::Resource.new(versions_endpoint.to_s).get
```
simon3z: There are too many conversions in these three lines  (and a string manipulation), I feel like we should keep the api_endpoint (as uri) and api_version separate and join them only before a rest query. Best case here would be to do just:

    RestClient::Resource.new(@api_endpoint).get

lib/kubeclient.rb
```diff
@@ -202,5 +202,11 @@ def all_entities
         result_hash[key_name] = send(method_name)
       end
     end
+
+    def server_api_versions
+      versions_endpoint = URI.parse(@api_endpoint.to_s)
+      versions_endpoint.path.slice! @api_version
+      RestClient::Resource.new(versions_endpoint.to_s).get
```
Fryguy: 2 things

- I don't think using .slice! on an method on URI#path is a good idea.  I recommend using `versions_endpoint.path = versions_endpoint.path.slice(@api_version)`.
- Should this be wrapped in `handling_kube_exception`
- I'm not sure on the name of the method.  I was thinking just `.api`.  I'm concerned that if in the future they add other stuff besides "versions" to the api endpoint, then this method name has to change, or will be inaccurate.  ".api" implies that it's the data from the api endpoint.

abonas: > I don't think using .slice! on an method on URI#path is a good idea. I recommend using versions_endpoint.path = versions_endpoint.path.slice(@api_version).

pls explain why isn't it a good idea?

> Should this be wrapped in handling_kube_exception

sure
> I'm not sure on the name of the method. I was thinking just .api. I'm concerned that if in the future they add other stuff besides "versions" to the api endpoint, then this method name has to change, or will be inaccurate. ".api" implies that it's the data from the api endpoint.

OK

Fryguy:
>> I don't think using .slice! on an method on URI#path is a good idea. I recommend using versions_endpoint.path = versions_endpoint.path.slice(@api_version).

> pls explain why isn't it a good idea?

The slice! method modifies the receiver directly, so it's modifying the actual String object instead of creating a new one.  In this case the URI instance "owns" the string, so it's not a good practice to modify it out from under them as you don't know how it's being used under the covers.  For example, if under the covers it's being used as a Hash key for something, then this code would break that.  In general, if you received the String from someone else, then don't modify it, but if you made it yourself, it's fair game.

abonas:
@Fryguy let me know what you think on the latest version of this...

lib/kubeclient.rb
```diff
@@ -24,9 +24,9 @@ class Client
 
     def initialize(uri, version = 'v1beta3')
       @api_endpoint = (uri.is_a? URI) ? uri : URI.parse(uri)
-      @api_endpoint.path = '/api' if @api_endpoint.path.empty?
-      if @api_endpoint.path.end_with? '/'
-        @api_endpoint.path = @api_endpoint.path.chop
+      @api_endpoint.path = '/api/' if @api_endpoint.path.empty?
+      unless @api_endpoint.path.end_with? '/'
+        @api_endpoint.path = @api_endpoint.path + '/'
```
Fryguy: Would prefer interpolation of the addition, but it's minor.

lib/kubeclient.rb
```diff
@@ -40,7 +40,7 @@ def rest_client
           ssl_client_cert:  @ssl_options[:client_cert],
           ssl_client_key:   @ssl_options[:client_key]
         }
-        endpoint_with_version = @api_endpoint.merge(@api_endpoint.path + '/' \
+        endpoint_with_version = @api_endpoint.merge(@api_endpoint.path \
```
Fryguy: The '\' line continuation is not necessary

Fryguy: In fact, I would use interpolation here as well.  `+` on String is generally avoided.

lib/kubeclient.rb
```diff
@@ -86,7 +86,7 @@ def get_entities(entity_type, klass, options)
     def watch_entities(entity_type, resource_version = nil)
       resource_name = get_resource_name(entity_type.to_s)
 
-      uri = @api_endpoint.merge(@api_endpoint.path + '/' + @api_version \
+      uri = @api_endpoint.merge(@api_endpoint.path + @api_version \
```
Fryguy: Same on both comments

Fryguy:
All minor comments.  Looks great!

lib/kubeclient.rb
abonas: > The '\' line continuation is not necessary

actually it is necessary, otherwise I found that rubocop complains, but if I move up the + , I can remove the line continuation and then it works fine.

abonas: > In fact, I would use interpolation here as well. + on String is generally avoided.

done

Fryguy:
Looks good

abonas:
> Looks good

merging, ok?

Fryguy:
Go for it

PR #52, opened 2015-03-25
refactor to keep version separate from uri

abonas:


abonas:
@Fryguy / @simon3z please review
@simon3z  this is following our discussion on the versions api patch.
if this patch seems the right direction, I'll refactor #47 accordingly

lib/kubeclient.rb
```diff
@@ -24,17 +24,14 @@ class Client
 
     def initialize(uri, version = 'v1beta3')
       @api_endpoint = (uri.is_a? URI) ? uri : URI.parse(uri)
-
       @api_endpoint.path = '/api' if @api_endpoint.path.empty?
-      @api_endpoint.merge!(File.join(@api_endpoint.path, version))
-      # version flag is needed to take care of the differences between
-      # versions
+      if @api_endpoint.path.end_with? '/'
+        @api_endpoint.path = @api_endpoint.path.chop
```
simon3z: I think it's possible to shorten it with:

 @api_endpoint.path.chop!

abonas: > I think it's possible to shorten it with:
> 
> @api_endpoint.path.chop!

yes, but @Fryguy  advised against a similar approach here:
https://github.com/abonas/kubeclient/pull/47#discussion_r27026824

lib/kubeclient.rb
```diff
@@ -43,11 +40,15 @@ def rest_client
           ssl_client_cert:  @ssl_options[:client_cert],
           ssl_client_key:   @ssl_options[:client_key]
         }
-
-        RestClient::Resource.new(@api_endpoint, options)
+        endpoint_with_version = URI.parse(@api_endpoint.to_s)
+        endpoint_with_version.path = endpoint_with_version.path << '/' \
+                                     << @api_version
```
simon3z: This can be simplified (removing the URI parse part) with something like:

    endpoint_with_version = @api_endpoint.merge(@api_endpoint.path + '/' + @api_version)

Same thing below.

abonas: > This can be simplified (removing the URI parse part) with something like:
> 
> endpoint_with_version = @api_endpoint.merge(@api_endpoint.path + '/' + @api_version)
> 
> Same thing below.

sure

lib/kubeclient.rb
```diff
@@ -43,11 +40,14 @@ def rest_client
           ssl_client_cert:  @ssl_options[:client_cert],
           ssl_client_key:   @ssl_options[:client_key]
         }
-
-        RestClient::Resource.new(@api_endpoint, options)
+        endpoint_with_version = @api_endpoint.merge(@api_endpoint.path + '/' \
+                                                    + @api_version)
```
simon3z: (Commenting on this nit only because I am already commenting below)

I prefer to leave the operator at the end of the line:

     endpoint_with_version = @api_endpoint.merge(@api_endpoint.path + '/' +
                                                 @api_version)

And I think you can spare the "\" at the end.

lib/kubeclient.rb
```diff
@@ -86,8 +86,8 @@ def get_entities(entity_type, klass, options)
     def watch_entities(entity_type, resource_version = nil)
       resource_name = get_resource_name(entity_type.to_s)
 
-      uri = api_endpoint.merge(
-        File.join(api_endpoint.path, 'watch', resource_name))
+      uri = @api_endpoint.merge(@api_endpoint.path + '/' + @api_version + '/' \
+                                + 'watch' + '/' + resource_name)
```
simon3z: Concatenating `'/' + 'watch' + '/'` is unnecessary. You can just do:

    uri = @api_endpoint.merge(@api_endpoint.path + '/' + @api_version +
                              '/watch/' + resource_name)

(Same for '\' as before, it shouldn't be required)

lib/kubeclient.rb
abonas: > (Commenting on this nit only because I am already commenting below)
> 
> I prefer to leave the operator at the end of the line:
> 
>  endpoint_with_version = @api_endpoint.merge(@api_endpoint.path + '/' +
>                                              @api_version)
> 
> And I think you can spare the "\" at the end.

rubocop has issues with this approach

lib/kubeclient.rb
abonas: > Concatenating '/' + 'watch' + '/' is unnecessary. You can just do:
> 
> uri = @api_endpoint.merge(@api_endpoint.path + '/' + @api_version +
>                           '/watch/' + resource_name)
sure

> (Same for '\' as before, it shouldn't be required)

PR #54, opened 2015-03-25
Add Kubeclient#api_valid?

Fryguy:
~~@abonas Please review.  I built this on top of #47, so it includes your commit.  If you change anything in there I will have to rebase, but that's ok.  So, only need to review the second commit.~~

With Kubeclient having a simple `api_valid?` method, then I can call that directly in https://github.com/ManageIQ/manageiq/pull/2156 without any extra logic.  Also, that allows you to change the validation method if `.api` gets unwieldy or is invalid in the future.

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -211,5 +211,10 @@ module Kubeclient
       end
       JSON.parse(response)
     end
+
+    def api_valid?
+      result = api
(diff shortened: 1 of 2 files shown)


Fryguy:
sure, I can change it to .api_valid?

Fryguy:
@abonas Updated to your latest changes and also renamed to api_valid?

Fryguy:
I think this Fixes #40

PR #62, opened 2015-04-07
fix new rubocop 0.30 offenses

abonas:


lib/kubeclient/common.rb
```diff
@@ -67,8 +67,8 @@ def rest_client
             ssl_client_cert: @ssl_options[:client_cert],
             ssl_client_key: @ssl_options[:client_key]
           }
-          endpoint_with_version = @api_endpoint.merge(@api_endpoint.path + '/' \
-                                                      + @api_version)
+          endpoint_with_version = @api_endpoint.merge(@api_endpoint.path + '/' +
+                                                      @api_version)
```
simon3z: I think we preferred `"#{@api_endpoint.path}/#{@api_version}"` in other parts.

Same below as well.

abonas: > I think we preferred "#{@api_endpoint.path}/#{@api_version}" in other parts.
> 
> Same below as well.

right, but it is not related to what this patch fixes.

simon3z: :+1:

abonas: > I think we preferred "#{@api_endpoint.path}/#{@api_version}" in other parts.
> 
> Same below as well.

and it causes a "line too long" offence...

PR #69, opened 2015-04-24
Fixed name of label selector parameter

stephenjlevy:
The parameter name for the label selector for get entities is "labelSelector", not "label_selector".
See http://kubernetes.io/third_party/swagger-ui/#!/v1beta3/listPod


abonas:
@stephenjlevy , thanks!
iirc, this is the second time the labels parameter named has changed in the past months...

PR #70, opened 2015-04-26
fix #51 add namespace to url of GET single entity calls

### LINKED ISSUE #51: GET a single entity must include namespace in the url
This doesn't work:
http://localhost:8080/api/v1beta3/services/kubernetes

This works:
http://localhost:8080/api/v1beta3/namespaces/default/services/kubernetes

The above is NOT relevant for nodes that are continued to be retrieved without namespace in the url.

abonas:


abonas:
cc @simon3z

PR #73, opened 2015-05-04
Fix #61 add namespace to url when delete an entity (except node, namespa...

### LINKED ISSUE #61: DELETE on v1beta3 must include namespace in the url

abonas:
...ce)


abonas:
cc @simon3z

PR #74, opened 2015-05-04
Add assertRequest checks to tests

abonas:


abonas:
cc @simon3z

test/test_kubeclient.rb
```diff
@@ -150,6 +150,10 @@ def test_entity_list
     assert_equal(2, services.size)
     assert_instance_of(Kubeclient::Service, services[0])
     assert_instance_of(Kubeclient::Service, services[1])
+
+    assert_requested :get,
```
simon3z: Minor: I feel like we should decide if we want parenthesis or not. The rest of the `asserts` are using parenthesis. Same for the other below.

abonas: > Minor: I feel like we should decide if we want parenthesis or not. The rest of the asserts are using parenthesis. Same for the other below.

yep. I'd go with parenthesis actually when rethinking this. just seems more "standard".

PR #75, opened 2015-05-05
client: improve api_valid check with version

simon3z:


test/test_kubeclient.rb
```diff
@@ -92,13 +92,31 @@ def test_api_valid
...
+
+  def test_api_valid_with_invalid_version
+    stub_request(:get, 'http://localhost:8080/api')
+      .to_return(status: 200, body: open_test_json_file('versions_list.json'))
+
+    client = Kubeclient::Client.new 'http://localhost:8080/api/', 'foobar1'
+    refute client.api_valid?
+  end
+
+  def test_api_valid_with_missing_version
```
abonas: the method name is not very clear until looking at the code.
I'd say it's more missing_versions_block kind of thing :)
missing_version implies to me that there's one specific version missing from the response...

test/test_kubeclient.rb
```diff
@@ -92,13 +92,31 @@ def test_api_valid
     stub_request(:get, 'http://localhost:8080/api')
       .to_return(status: 200, body: open_test_json_file('versions_list.json'))
 
+    %w(v1beta1 v1beta2 v1beta3).each do |version|
+      client = Kubeclient::Client.new 'http://localhost:8080/api/', version
+      assert client.api_valid?
+    end
+  end
+
```
abonas: I think it's also worth to add a test which checks that the logic works well with the default version (when user didn't pass a version)

simon3z:
@abonas this should be ok now.

test/test_kubeclient.rb
```diff
@@ -92,13 +92,33 @@ def test_api_valid
     stub_request(:get, 'http://localhost:8080/api')
       .to_return(status: 200, body: open_test_json_file('versions_list.json'))
 
+    args = ['http://localhost:8080/api/']
+
+    [nil, 'v1beta1', 'v1beta2', 'v1beta3'].each do |version|
+      client = Kubeclient::Client.new(*(version ? args + [version] : args))
```
abonas: that looks nice - what is this * syntax? :)

PR #83, opened 2015-05-18
Add HTTP basic authentication support

oschreib:
Add support for http basic authentication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,12 +64,14 @@ For testing and development purpose you can disable the ssl check with:
 client.ssl_options(verify_ssl: OpenSSL::SSL::VERIFY_NONE)
 ```
 
-If you are using bearer tokens for authentication as described
-[here](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/authentication.md) then you can specify the
-bearer token to use for authentication:
+If you are using basic or bearer authentication as described
+[here](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/authentication.md) then you can specify one
+of the following:
 
 ```ruby
 client.bearer_token('MDExMWJkMjItOWY1Ny00OGM5LWJlNDEtMjBiMzgxODkxYzYz')
+or
+client.basic_auth('username', 'password')
 ```
 
 If you are running your app using kubeclient inside a Kubernetes cluster, then you can have a bearer token file
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -71,7 +71,9 @@ module Kubeclient
           verify_ssl: @ssl_options[:verify_ssl],
           ssl_client_cert: @ssl_options[:client_cert],
           ssl_client_key: @ssl_options[:client_key],
-          bearer_token: @bearer_token
+          bearer_token: @bearer_token,
+          user: @basic_auth_user,
+          password: @basic_auth_password
         }
         RestClient::Resource.new(@api_endpoint.merge(path).to_s, options)
       end
@@ -100,7 +102,9 @@ module Kubeclient
           verify_mode: @ssl_options[:verify_ssl],
           cert: @ssl_options[:client_cert],
           key: @ssl_options[:client_key],
-          bearer_token: @bearer_token
+          bearer_token: @bearer_token,
+          basic_auth_user: @basic_auth_user,
+          basic_auth_password: @basic_auth_password
         }
 
         WatchStream.new(uri, options)
@@ -223,6 +227,11 @@ module Kubeclient
         RestClient.add_before_execution_proc do |req|
           req['authorization'] = "Bearer #{@bearer_token}"
         end
+      end
+
+      def basic_auth(user, password)
+        @basic_auth_user = user
+        @basic_auth_password = password
       end
     end
   end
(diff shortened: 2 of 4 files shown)


oschreib:
cc @simon3z  @abonas

oschreib:
The reason this patch fails test is actually the bearer_token tests.
Since bearer token test use the RestClient.add_before_execution_proc , it adds the authorization header to each and every request, which conflicts with the basic auth.

The best solution I have in mind is to reset this (RestClient.reset_before_execution_procs) after each bearer test.

jimmidyson:
@oschreib That sounds fine to me.

oschreib:
@jimmidyson actually, the issue is much deeper.
Since you're using RestClient.add_before_execution_proc , it adds the header to every request made by ANY RestClient instance. This means that if someone has two client objects, nothing will work for him (for instance, if I have two k8s providers in miq, both of them using bearer token, only one of them will work).

jimmidyson:
@oschreib Fair point. Works for my use case because I only have one k8s provider to talk to. Seems to be a limitation in the RestClient that it's a singleton  - there's no way to create multiple instances of it AFAIK.

lib/kubeclient/common.rb
```diff
@@ -71,11 +71,18 @@ def create_rest_client(path = nil)
...
           ssl_client_key: @ssl_options[:client_key],
-          bearer_token: @bearer_token
+          bearer_token: @bearer_token,
+          user: @basic_auth_user,
+          password: @basic_auth_password
         }
         RestClient::Resource.new(@api_endpoint.merge(path).to_s, options)
       end
 
+      def update_rest_client
```
simon3z: Move this to another PR (if any). We'll decide on how to deal with this/these bugs somewhere else.

lib/kubeclient/common.rb
```diff
@@ -224,6 +233,12 @@ def bearer_token(bearer_token)
           req['authorization'] = "Bearer #{@bearer_token}"
         end
       end
+
+      def basic_auth(user, password)
+        @basic_auth_user = user
+        @basic_auth_password = password
+        update_rest_client
```
simon3z: Don't call `update_rest_client`. How to address the bug is another discussion. Currently we have the same problem with ssl.

simon3z:
:+1: this is ok for me, how to handle changes of ssl/auth should be discussed separately (in a proper issue)

README.md
```diff
@@ -64,12 +64,14 @@ For testing and development purpose you can disable the ssl check with:
 client.ssl_options(verify_ssl: OpenSSL::SSL::VERIFY_NONE)
 ```
 
-If you are using bearer tokens for authentication as described
-[here](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/authentication.md) then you can specify the
-bearer token to use for authentication:
+If you are using basic or bearer authentication as described
```
abonas: Please s/basic or bearer authentication/basic authentication or bearer tokens

lib/kubeclient/watch_stream.rb
```diff
@@ -36,6 +32,19 @@ def each
...
       end
 
+      def generate_request
+        request = Net::HTTP::Get.new(@uri)
+        if @options[:basic_auth_user] && @options[:basic_auth_password]
+          request.basic_auth @options[:basic_auth_user],
+                             @options[:basic_auth_password]
+        end
+
+        if @options[:bearer_token]
```
abonas: does this imply you might have both basic_auth and bearer token at the same time?

test/test_kubeclient.rb
```diff
@@ -261,6 +262,35 @@ def test_api_bearer_token_failure
...
+      .to_return(body: open_test_json_file('pod_list_b3.json'),
+                 status: 200)
+
+    client = Kubeclient::Client.new 'http://localhost:8080/api/'
+    client.basic_auth('username', 'password')
+
+    pods = client.get_pods
+
+    assert_equal('Pod', pods.kind)
+    assert_equal(1, pods.size)
```
abonas: what about assert_requested validation?

simon3z:
@abonas I think we can merge this as-is (unless you think there's something major to change), so @oschreib will be more free to tackle #85 (which will include some large refactoring). Meanwhile we can deliver `basic_auth` in manageiq.

abonas:
> @abonas I think we can merge this as-is (unless you think there's something major to change), so @oschreib will be more free to tackle #85 (which will include some large refactoring). Meanwhile we can deliver basic_auth in manageiq.

yes, that's fine by me, as long as:
1. the review comments above are addressed/responded
2. we won't release a new gem version until #85 is also resolved.

simon3z:
> yes, that's fine by me, as long as:
> 1. the review comments above are addressed/responded
> 2. we won't release a new gem version until #85 is also resolved.

The idea was to release a new gem before #85 (which as all refactoring is low-priority and therefore may take some time).

abonas:
> The idea was to release a new gem before #85 (which as all refactoring is low-priority and therefore may take some time).

ok, we can release twice if it will help making progress.

README.md
abonas: > Please s/basic or bearer authentication/basic authentication or bearer tokens

I'll fix it in a separate patch

lib/kubeclient/watch_stream.rb
abonas: > does this imply you might have both basic_auth and bearer token at the same time?

should be fixed as part of #85 refactor

PR #89, opened 2015-05-28
Send bearer token only on relevant instance

### LINKED ISSUE #84: Using Bearer token impacts any RestClient instance
As bearer token is using RestClient.add_before_execution_proc to inject the "authorization" HTTP header, it can affect any other RestClient instance created.
For instance, having two separate KubeClient instances, with separate bearer token inside each one, will result in a conflict, as the last bearer token will be injected to both of the objects.

p.s  - we might want to get https://github.com/rest-client/rest-client/pull/137 merged, which will be perfect for that exact matter.

oschreib:
Currently, when setting a new bearer token to a client instance, it affects
any living instance, and might break their authentication support.
This patch fixes that issue.
Fixes #84

### DIFF
--- a/test/test_kubeclient.rb
+++ b/test/test_kubeclient.rb
@@ -232,6 +232,21 @@ class KubeClientTest < MiniTest::Test
     assert_instance_of(Kubeclient::Namespace, result['namespace'][0])
   end
 
+  def test_api_bearer_token_with_params_success
+    stub_request(:get, 'http://localhost:8080/api/v1beta3/pods?labelSelector=name=redis-master')
+      .with(headers: { Authorization: 'Bearer valid_token' })
+      .to_return(body: open_test_json_file('pod_list_b3.json'),
+                 status: 200)
+
+    client = Kubeclient::Client.new 'http://localhost:8080/api/'
+    client.bearer_token('valid_token')
+
+    pods = client.get_pods(label_selector: 'name=redis-master')
+
+    assert_equal('Pod', pods.kind)
+    assert_equal(1, pods.size)
+  end
+
   def test_api_bearer_token_success
     stub_request(:get, 'http://localhost:8080/api/v1beta3/pods')
       .with(headers: { Authorization: 'Bearer valid_token' })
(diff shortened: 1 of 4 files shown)


oschreib:
cc @simon3z @abonas

simon3z:
Let's keep this here until we understand if it's worth having a class encapsulating RestClient so `get`, `put`, `delete` would be transparent. @oschreib you probably need to address the watches as well and IIRC also `api`/`api_valid`.

abonas:
> Let's keep this here until we understand if it's worth having a class encapsulating RestClient so get, put, delete would be transparent. @oschreib you probably need to address the watches as well and IIRC also api/api_valid.

true.
@oschreib :
1. you don't *have* to call the param "additional_headers" (you can, but you don't have to) - I find it somewhat long without real necessity. but it's a nit.

2. tests please - verify that url, parameters and headers are passed and are all in the right places. (iirc there are some tests for them separately, but not together)

abonas:
@oschreib please check whether the other dependent gem (openshift_client) should be updated as well.

jimmidyson:
Looked at this earlier, looks like a much better approach than I took. I did think that we were going to refactor stuff like bearer_token, etc into constructor params?

What happens if basic auth & bearer token are both specified as they use the same Authorization header?

jimmidyson:
What's happening with this PR? Would like to see it merged unless there's anything else to add?

oschreib:
@jimmidyson I'm now working on the miq side of this patch, once it will be ready, we will merge them both. I hope it will be today.

jimmidyson:
@oschreib Can't we merge this before the miq stuff? miq uses released gem right so merging this here shouldn't affect miq dev until next release & upgrade of miq dependency?

test/test_kubeclient.rb
```diff
@@ -232,6 +232,21 @@ def test_get_all
...
+      .to_return(body: open_test_json_file('pod_list_b3.json'),
+                 status: 200)
+
+    client = Kubeclient::Client.new 'http://localhost:8080/api/'
+    client.bearer_token('valid_token')
+
+    pods = client.get_pods(label_selector: 'name=redis-master')
+
+    assert_equal('Pod', pods.kind)
+    assert_equal(1, pods.size)
```
abonas: please add assert_requested that indeed the request looks exactly as expected (and executed once, while at it :) )

abonas:
@oschreib , other than a test related comment , :+1: 
please address that and I'll merge.

PR #94, opened 2015-06-04
Fixes #93: Remove old API versions, add v1

### LINKED ISSUE #93: Drop API versions v1beta1/2 & add v1
Only v1beta3 & v1 are supported now.

jimmidyson:


abonas:
@simon3z , while reviewing this PR,  I'm looking at the below file, https://github.com/abonas/kubeclient/blob/master/test/json/event_list_b3.json
and it seems that though it works with v1beta3, the events format is v1beta1 - is that right?
does that file need an update? if v1beta1 is removed, then this input is probably no longer valid.

jimmidyson:
There's no validation right now in this kube client so the structure of what is returned is simply converted to an OpenStruct object. Should probably update the format of the file, but it doesn't affect any tests.

abonas:
> There's no validation right now in this kube client so the structure of what is returned is simply converted to an OpenStruct object. Should probably update the format of the file, but it doesn't affect any tests.

yes, all the above is correct :) . but the question was to @simon3z to make sure we're not making wrong assumptions on the objects structure elsewhere where this client is used.

iterion:
Any chance we could get this merged? v1 was the default in the latest release so new users of kubernetes wouldn't be able to use this even though it's really helpful!

abonas:
@iterion, I can certainly merge this, but this patch affects only the tests. 
The client can connect today to v1 api.
For example:
client = Kubeclient::Client.new 'http://yourmasterhost:8080/api/', 'v1'

iterion:
Ah, sorry, and thanks for the heads up! I had been using the gem with v1beta3 and saw this PR and assumed that v1 was unsupported.

jimmidyson:
I guess it's time to make v1 the default?

abonas:
@iterion, I haven't tested the full implications of working against v1. But the client does connect and retrieve entities just fine at least. Can you elaborate how/why do you use the kubeclient? just trying to get an idea of the gem's users use cases.

@jimmidyson, a bit too early for making it default - k8s 1.0 is not out yet, so I'd wait to get closer to the release. and there's also a need to add tests for v1 in kubeclient to make sure all is fine.

iterion:
@abonas If I run into any issues I'll be sure to send a PR. Currently I'm using it in a few separate utilities, the main use case is a simple utility that finds a service, extracts the ingress hostname, and configures DNS on AWS. I'm currently working on a script that will set up the full stack of our application within a namespace. After that, i'll probably create some utilities to help manage application deploys.

PR #95, opened 2015-06-07
Update KubeException to inherit StandardError

oschreib:
Currently, KubeException inherits from Exception, which is bad because
the default "rescue" behaviour is to catch only exception which are
StandardException.


oschreib:
cc @abonas @simon3z

PR #96, opened 2015-06-17
Add namespace support for get_entities

simonswine:
Hi @abonas, 

thank your for you Kubernetes Ruby API. This small change adds namespace support for entities. 

Tests and rubocop is fine after the change...

Cheers

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -115,9 +115,9 @@ module Kubeclient
           params['params'] = { labelSelector: options[:label_selector] }
         end
 
-        # TODO: namespace support?
+        ns_prefix = build_namespace_prefix(options[:namespace])
         response = handle_exception do
-          rest_client[resource_name(entity_type)]
+          rest_client[ns_prefix + resource_name(entity_type)]
           .get(params.merge(@headers))
         end
 
--- a/test/json/entity_list_b3_mynamespace.json
+++ b/test/json/entity_list_b3_mynamespace.json
@@ -1,0 +1,74 @@
+{
+    "kind": "ServiceList",
+    "apiVersion": "v1beta3",
+    "metadata": {
+        "selfLink": "/api/v1beta3/namespaces/mynamespace/services",
+        "resourceVersion": "475878"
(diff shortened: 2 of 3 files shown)


abonas:
@simonswine , please add tests to get this change covered - with and without namespace.
And thanks for this patch!

simonswine:
Ok now with a tests for it. 

Sorry I was looking into your tests, as soon as I noticed your not using rspec I decided not to write a test for it.. :laughing:

test/json/entity_list_b3_mynamespace.json
```diff
@@ -0,0 +1,56 @@
+{
```
abonas: is this a real output from k8s? kubernetes-ro and kubernetes are internal k8s services out of the box, belonging to default namespace.

simonswine: @abonas The latest commit contains now a real world example (v1beta3)

lib/kubeclient/common.rb
```diff
@@ -115,9 +115,9 @@ def get_entities(entity_type, klass, options)
           params['params'] = { labelSelector: options[:label_selector] }
         end
 
-        # TODO: namespace support?
+        ns_prefix = build_namespace_prefix(options[:namespace])
```
simon3z: Both `:label_selector` (maybe to be renamed to `:selector` depending on kubernetes terminology) and `namespace` should be additional arguments (as in `get_entity`).

jimmidyson: FYI - in Kubernetes there are label selectors to filter resources & field selectors to specify what fields to return.

fabiokr:
:+1: I'm also needing this functionality. Anything holding on this PR that I can help with?

abonas:
@fabiokr, the PR was not revisited after a review round, hence it was not merged. 
I am adding a similar functionality in https://github.com/abonas/kubeclient/pull/105
It is also awaiting review from @simon3z

abonas:
implemented in https://github.com/abonas/kubeclient/pull/105

PR #97, opened 2015-06-18
Bugfix: update of a entity not working as the name is not read correctly

simonswine:
When I retrieve an entitiy via get_entity method, I cannot update it, as the name in the url is empty. I tested a service with API v1beta3.

I fixed this behavior and tried to add tests for update_entity.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -173,7 +173,7 @@ module Kubeclient
       end
 
       def update_entity(entity_type, entity_config)
-        name = entity_config.name
+        name = entity_config.metadata.name
         # to_hash should be called because of issue #9 in recursive open
         # struct
         hash = entity_config.to_hash
--- a/test/test_service.rb
+++ b/test/test_service.rb
@@ -109,4 +109,28 @@ class TestService < MiniTest::Test
                      'http://localhost:8080/api/v1beta3/namespaces/development/services/redis-slave',
                      times: 1)
   end
+
+  def test_update_service
+    entity = 'service'
+    object = Kubeclient::Service.new
+    name = 'my_service'
+    object.metadata = {
+      'name' => name
+    }
+
+    stub_request(:put, %r{/services/#{name}})\
+      .to_return(
+        body: open_test_json_file('service_update_b3.json'),
+        status: 200
+      )
+
+    client = Kubeclient::Client.new 'http://localhost:8080/api/', 'v1beta3'
+    client.update_entity entity, object
+
+    assert_requested(
+      :put,
+      "http://localhost:8080/api/v1beta3/services/#{name}",
+      times: 1
+    )
+  end
 end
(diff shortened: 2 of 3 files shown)


lib/kubeclient/common.rb
```diff
@@ -173,7 +173,7 @@ def create_entity(entity_type, entity_config, klass)
       end
 
       def update_entity(entity_type, entity_config)
-        name = entity_config.name
+        name = entity_config.name || entity_config.metadata.name
```
abonas: no need to do the "or" part. it's always now part of metadata. v1beta1 is not supported anyway...
so you can just replace the old statement with the new one.

test/test_common.rb
```diff
@@ -0,0 +1,48 @@
+require 'test_helper'
+
+def open_test_json_file(name)
+  File.new(File.join(File.dirname(__FILE__), 'json', name))
+end
+
+# Kubernetes client entity tests
+class KubeClientTest < MiniTest::Test
+  def test_update_entity_name
```
abonas: this belongs better in test_service.rb

test/test_common.rb
```diff
@@ -0,0 +1,48 @@
...
+    entity = 'service'
+    object = Kubeclient::Service.new
+    object.name = 'my_service'
+    object.metadata = {}
+
+    stub_request(:put, %r{/services/#{object.name}})\
+      .to_return(body: '{}', status: 200)
```
abonas: does an update return an empty body?

abonas:
@simonswine, 
Hi, are you planning to revisit this patch?

simonswine:
Hi @abonas,

Thanks for reminding me, this commit should address all of your concerns...

Cheers... simonswine

lib/kubeclient/common.rb
```diff
@@ -173,7 +173,7 @@ def create_entity(entity_type, entity_config, klass)
       end
 
       def update_entity(entity_type, entity_config)
-        name = entity_config.name
+        name = entity_config.metadata.name
```
abonas: :+1:

test/test_service.rb
```diff
@@ -109,4 +109,28 @@ def test_get_service
                      'http://localhost:8080/api/v1beta3/namespaces/development/services/redis-slave',
                      times: 1)
   end
+
+  def test_update_service
```
abonas: 1. what field is this test updating ?
2. this example will not work on a real k8s server as there are many required fields that are not specified. (for example - resourceVersion)

simonswine: I won't continue here, sorry but I think your idea of testing is not how it's meant to be. The test cases will never run against a real k8s server. Just merge this, it's better than the current behavior with the wrong name variable and has absolute no testing of updates at all.

Hope you can understand my point here....

Cheers

abonas: > Hope you can understand my point here....

yes I do. looking forward I am having thoughts to redo the tests to be more "live", because the current status leads to problems like this of a "half baked" test, and as the library and its user base grows, it's very important to allow contributors to safely and quickly introduce changes.

PR #98, opened 2015-06-19
Allow bearer token file for auth

jimmidyson:


jimmidyson:
This also enabled the use of service account provided tokens.

lib/kubeclient.rb
```diff
@@ -13,6 +13,9 @@ module Kubeclient
   class Client < Common::Client
     attr_reader :api_endpoint
 
+    SERVICE_ACCOUNT_TOKEN_FILE =
```
abonas: Please do not hardcode this in the client. Passing the path to file to the client should do the work.
Hardcoding it means duplicating server side configuration/logic, is easily breakable by server side changes and will require more maintenance.

lib/kubeclient.rb
```diff
@@ -52,10 +52,15 @@ def initialize(uri,
...
-
-      bearer_token(auth_options[:bearer_token]) if auth_options[:bearer_token]
+      if auth_options[:user]
+        @basic_auth_user = auth_options[:user]
+        @basic_auth_password = auth_options[:password]
+      elsif auth_options[:bearer_token]
+        bearer_token(auth_options[:bearer_token])
+      elsif auth_options[:bearer_token_file] &&
+            File.file(auth_options[:bearer_token_file])
+        bearer_token(File.read(auth_options[:bearer_token_file]))
```
abonas: what about non existing files etc.? no need for exception handling here?

abonas:
@simon3z, @oschreib  please review.

lib/kubeclient.rb
jimmidyson: It's just a convenience so that users don't have to remember the service token path. It will only be used if passed in to the client constructor, not used by default.

lib/kubeclient.rb
jimmidyson: Will add exception handling. Should this fail if files don't exist or just log error & continue?

lib/kubeclient.rb
abonas: > It's just a convenience so that users don't have to remember the service token path. It will only be used if passed in to the client constructor, not used by default.

I understand, but it still means this constant is hardcoded in the client, and I prefer to avoid maintaining it there - it means duplicating logic and chasing changes in the server side. 
Whoever uses the client can and should pass the path to the token "from the outside".

lib/kubeclient.rb
abonas: > Will add exception handling. Should this fail if files don't exist or just log error & continue?

what is the behavior when you connect to k8s when the token doesn't exist/wrong location? same as passing a bad token/different?

lib/kubeclient.rb
jimmidyson: Well OK - it's your client after all... I can just see this leading to lots of wasted time by devs using this as they hunt down typos in the token file name & actually finding the token file name... but I''ll remove it at your request.

abonas: > Well OK - it's your client after all... I can just see this leading to lots of wasted time by devs using this as they hunt down typos in the token file name & actually finding the token file name... but I''ll remove it at your request.

thank you :)
This token/path should be available and documented on server side docs.
plus, as we discussed in the past, using this from inside k8s is not the main use case, hence I currently do not expect a lot of issues around this.
And as in any issue and design, this can be always reevaluated and change in the future if required.

lib/kubeclient.rb
jimmidyson: So wrong token will be handled as before - returning 403s, but if the file doesn't exist or is unreadable do you think this should return an error in the constructor? I think it should so let me update & push that change too. If you think differently let me know & I will update accordingly.

jimmidyson:
@abonas Updated as per your review, added test in for non-existent token file.

oschreib:
@jimmidyson @abonas it's looks good to me, however, I'd rather have a "full loop" test (e.g give the client a token file, make sure that it reads it properly and sends the right token to destination) and not only argument error checks.

jimmidyson:
@oschreib Fair point. Let me add that in too.

abonas:
> @jimmidyson @abonas it's looks good to me, however, I'd rather have a "full loop" test (e.g give the client a token file, make sure that it reads it properly and sends the right token to destination) and not only argument error checks.

:+1:  agree

jimmidyson:
@oschreib @abonas Added - https://github.com/abonas/kubeclient/pull/98/files#diff-1efd65c010e2f18aadd1dd73e84c68fdR365

README.md
```diff
@@ -89,12 +89,31 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , 'v1beta3',
...
 mounted inside your pod by using a
 [Service Account](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/design/service_accounts.md). This
 will mount a bearer token [secret](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/design/secrets.md)
 a/ `/var/run/secrets/kubernetes.io/serviceaccount/token` (see [here](https://github.com/GoogleCloudPlatform/kubernetes/pull/7101)
-for more details).
+for more details). For example:
+
+```ruby
+auth_options = {
+  bearer_token_file: '/var/run/secrets/kubernetes.io/servicaccount/token'
```
abonas: I think you are missing an "e" in "servicaccount"

jimmidyson:
@abonas Good catch! Updated

test/test_kubeclient.rb
```diff
@@ -343,6 +343,43 @@ def test_init_user_and_bearer_token
     end
   end
 
+  def test_bearer_token_and_bearer_token_file
```
abonas: tests look good!
I think it's worth to add an additional validation on the specific error messages in all the added tests, since all the failures produce exactly the same ArgumentError, to validate which exact message happens in each case.
It can be done by getting the err returned from assert_raises, and doing an assert_match on the err.message object with the relevant expected message.

jimmidyson:
@abonas Updated tests as requested

abonas:
:+1:  from my side.
@simon3z ?

lib/kubeclient.rb
```diff
@@ -67,13 +69,26 @@ def all_entities
     private
 
     def validate_auth_options(opts)
+      bearer_token_keys = [:bearer_token, :bearer_token_file]
+      exclusive_keys = bearer_token_keys << :user
```
simon3z: Would this be equal to:

    exclusive_keys = [:bearer_token, :bearer_token_file, :user]

Any reason to define `bearer_token_keys`?

lib/kubeclient.rb
```diff
@@ -67,13 +69,26 @@ def all_entities
...
+
+      msg = 'Invalid auth options: specify only one of user/password,' \
+            ' bearer_token or bearer_token_file'
+      fail ArgumentError, msg unless exclusive_keys.one? { |s| opts.key?(s)  }
+
+      msg = 'Basic auth requires both user & password'
+      fail ArgumentError, msg if opts.key?(:user) && !opts.key?(:password)
+    end
+
+    def validate_bearer_token_file(bearer_token_file)
```
simon3z: Unless it's a display problem here on github it seems that this function is indented and lines are grouped in a misleading way. I think it should be something like:

    def validate_bearer_token_file(bearer_token_file)
        msg = "Token file #{bearer_token_file} does not exist"
        fail ArgumentError, msg unless File.file?(bearer_token_file)
    
        msg = "Cannot read token file #{bearer_token_file}"
        fail ArgumentError, msg unless File.readable?(bearer_token_file)
    end

lib/kubeclient.rb
jimmidyson: There was some refactoring done & I didn't tidy it back up again - will do now.

lib/kubeclient.rb
jimmidyson: This happened during refactoring due to line length - should be ok to put on same line again now. Will do it

jimmidyson:
@simon3z Addressed your comments.

simon3z:
:+1: ready for merge

PR #102, opened 2015-07-16
Add Secrets

iterion:
I added tests for secrets that are on v1, while the `get_all_entities` test still works with v1beta3. I'm happy to change them all to v1beta3 for consistency, but figured we'll probably be moving tests to v1 anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![Dependency Status](https://gemnasium.com/abonas/kubeclient.svg)](https://gemnasium.com/abonas/kubeclient)
 
 A Ruby client for Kubernetes REST api.
-The client supports GET, POST, PUT, DELETE on nodes, pods, services, replication controllers, namespaces and endpoints.
+The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces and endpoints.
 The client currently supports Kubernetes REST api version v1beta3.
 
 ## Installation
@@ -118,7 +118,7 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , 'v1beta3',
 ## Examples:
 
 #### Get all instances of a specific entity type
-Such as: `get_pods`, `get_services`, `get_nodes`, `get_replication_controllers`
+Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`
 
 ```ruby
 pods = client.get_pods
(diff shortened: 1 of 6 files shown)


abonas:
@iterion, this is great, thanks for the contribution.
for starters (before the code review):
1. please squash the rubocop fixes, it shouldn't be a separate commit.
2. please address the newly added code in the readme - it should reflect this change for the users.
thanks!

iterion:
Rubocop commit squashed and readme updated.

abonas:
@oschreib @jimmidyson fyi, please review.
@jimmidyson is this something you are using too?

abonas:
@iterion, what is the server behavior with delete? who can delete a secret?

jimmidyson:
@abonas We use secrets for most of our stuff but with this gem we're just a consumer of secrets that are mounted inside pods.

Behaviour is no different to any other resource from a management point of view. Kubernetes has no idea of authorisation afaik so anyone with a valid token or authentication can create/delete/etc. In openshift the appropriate user roles need to be set up to manage secrets in a namespace but that is outside of the scope of this gem obviously.

iterion:
@jimmidyson is correct. Secrets aren't given any special treatment.

abonas:
@iterion, what is the server behavior with delete? who can delete a secret?
I saw you also didn't add secrets to the delete section in the readme, hence I'd like to understand better the use cases and the behavior here.

iterion:
I'm happy to add secrets to the delete section if that's what you want?

abonas:
> I'm happy to add secrets to the delete section if that's what you want?

@iterion 
I'm just trying to understand why are they missing from the readme and the functionality around deleting secrets - is this missing from readme by mistake or because there's some issue with deleting secrets? can anyone delete a secret via REST api? (permissions wise)

iterion:
I see, for me there are no issues deleting, nor should there be for anyone else, I think I just missed it from that section. I can add it to that section.

abonas:
> I see, for me there are no issues deleting, nor should there be for anyone else, I think I just missed it from that section. I can add it to that section.

ok now it's clear. so yes, please add it to the readme :)

jonmoter:
Thanks for adding this, I was just looking for this functionality, and saw that it's already in a PR!

abonas:
> ok now it's clear. so yes, please add it to the readme :)

@iterion , just waiting for the readme addition before merge.

abonas:
> Thanks for adding this, I was just looking for this functionality, and saw that it's already in a PR!

@jonmoter glad to see one more user of the gem :)
can you perhaps briefly describe where (open source?)/how do you use the gem? 
just trying to gather some real users use cases...

jonmoter:
I work at Zendesk, and am currently researching Kubernetes.  We have a deploy tool called [Samson](https://github.com/zendesk/samson) that's written in ruby, and currently does deploys via shell scripts & capistrano.  I'm experimenting with modifying Samson to be able to deploy applications to Kubernetes, and this gem has been great for using the k8s API.

abonas:
thanks @jonmoter

iterion:
@abonas Sorry for the delay, added the line in the readme.

PR #104, opened 2015-08-09
increase max line length from 80 to 100

abonas:


abonas:
cc @simon3z

simon3z:
I'm a fan of 80 :smile:  but 100 is ok if it improves clarity

abonas:
> I'm a fan of 80  but 100 is ok if it improves clarity

yes it does improve readability.

PR #105, opened 2015-08-09
Add Kubernetes guestbook example tests

abonas:
Add support for get_entities with namespace
Add tests for create and delete services, replication controllers
Based on guestbook example
https://github.com/kubernetes/kubernetes/tree/master/examples/guestbook-go
Record VCR

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -128,7 +128,6 @@ module Kubeclient
 
     def watch_entities(entity_type, resource_version = nil)
       resource = resource_name(entity_type.to_s)
-
       uri = @api_endpoint
             .merge("#{@api_endpoint.path}/#{@api_version}/watch/#{resource}")
 
@@ -152,15 +151,15 @@ module Kubeclient
       Kubeclient::Common::WatchStream.new(uri, options)
     end
 
-    def get_entities(entity_type, klass, options)
+    def get_entities(entity_type, klass, options = {})
       params = {}
       if options[:label_selector]
         params['params'] = { labelSelector: options[:label_selector] }
       end
 
-      # TODO: namespace support?
+      ns_prefix = build_namespace_prefix(options[:namespace])
       response = handle_exception do
-        rest_client[resource_name(entity_type)]
+        rest_client[ns_prefix + resource_name(entity_type)]
         .get(params.merge(@headers))
       end
 
--- a/README.md
+++ b/README.md
@@ -124,10 +124,18 @@ Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replicatio
 pods = client.get_pods
 ```
 
+Get all entities of a specific type in a namespace:<br>
+
+```ruby
+services = client.get_services(namespace: 'development')
+```
+
 You can get entities which have specific labels by specifying a parameter named `label_selector` (named `labelSelector` in Kubernetes server):
+
 ```ruby
 pods = client.get_pods(label_selector: 'name=redis-master')
 ```
+
 You can specify multiple labels (that option will return entities which have both labels:
 ```ruby
 pods = client.get_pods(label_selector: 'name=redis-master,app=redis')
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake', '~> 10.0'
   spec.add_development_dependency 'minitest'
   spec.add_development_dependency 'webmock'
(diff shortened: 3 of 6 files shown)


abonas:
@simon3z thoughts on this change?

abonas:
@simon3z please review

iterion:
@abonas It'd be great if we could get this merged in. The namespace support is probably the most important piece for us, so if it's easier, we could break that piece out into a separate PR. I'm happy to do the work for that since this one seems a little stuck?

abonas:
hi @iterion, yeah, I was hoping for a long time this patch would get reviewed before merging it but it wasn't. I understand this feature is needed, so I'll try to rebase it next week and you could review it then perhaps? how does that sound?

iterion:
Sounds perfect, thanks!

abonas:
> Sounds perfect, thanks!

@iterion, I rebased and repushed it, please give it a round of review.
I'll go over it again to see if anything can be simplified or shortened too.

iterion:
Looks great, thanks for updating this.

Some thoughts:
* Do we have a unit test for get_entities and/or do we want one? It looks like the method was updated without any test changes.
* In the guestbook example, might it be easier/cleaner to store the various rcs/services in yaml or json files then just read them in for the test? It would save writing many methods to build such objects and might make it easier to write integration tests easier in the future.

Other than that, awesome! :+1:

lib/kubeclient/common.rb
```diff
@@ -152,15 +151,15 @@ def watch_entities(entity_type, resource_version = nil)
       Kubeclient::Common::WatchStream.new(uri, options)
     end
 
-    def get_entities(entity_type, klass, options)
+    def get_entities(entity_type, klass, namespace = nil, options = {})
```
iterion: Maybe we could just include namespace in the options hash. We could also do the same for the specific implementations of `get_#{entity_type}`. I'm torn on this though, namespace is specified fairly often, so it's nice that it's treated separately. But, it's also possible for it to be nil, in which case it might be nice to not have to pass a nil parameter around.

abonas:
@iterion thanks a lot for this, it's really helpful. I'll go over your comments and address them in the next cpl of days. vacation + some public holidays causes a bit of delay, sorry about that...

lib/kubeclient/common.rb
```diff
@@ -73,8 +73,8 @@ def self.define_entity_methods(entity_types)
         entity_name_plural = pluralize_entity(entity_name)
 
         # get all entities of a type e.g. get_nodes, get_pods, etc.
-        define_method("get_#{entity_name_plural}") do |options = {}|
-          get_entities(entity_type, klass, options)
+        define_method("get_#{entity_name_plural}") do |namespace:nil, options:{}|
+          get_entities(entity_type, klass, namespace, options)
```
jonmoter: Couldn't we just include `namespace` in the options hash? It feels weird to say:

`client.get_events(namespace: 'default', options: { label_selector: 'foo=bar' })`
rather than:
`client.get_events(namespace: 'default', label_selector: 'foo=bar')`

abonas: > Couldn't we just include namespace in the options hash? It feels weird to say:
> 
> client.get_events(namespace: 'default', options: { label_selector: 'foo=bar' })
> rather than:
> client.get_events(namespace: 'default', label_selector: 'foo=bar')

yep, based on your and @iterion feedback, I revised it to be like it.

abonas:
> Do we have a unit test for get_entities and/or do we want one? It looks like the method was updated without any test changes.



> Looks great, thanks for updating this.
> 
> Some thoughts:
> 
> Do we have a unit test for get_entities and/or do we want one? It looks like the method was updated without any test changes.

there are tests for get_entities in test guestbook go - or you meant something else?

> In the guestbook example, might it be easier/cleaner to store the various rcs/services in yaml or json files then just read them in for the test? It would save writing many methods to build such objects and might make it easier to write integration tests easier in the future.

generally speaking - yes, but one of my goals in this was to 
check that it is possible to construct the objects via the api.
> Other than that, awesome!

iterion:
> there are tests for get_entities in test guestbook go - or you meant something else?

I was thinking something that tested get_entities in isolation. But, the coverage in test guestbook is probably sufficient for now.

> generally speaking - yes, but one of my goals in this was to 
check that it is possible to construct the objects via the api.

Fair enough, it's also true that we should test that.

Overall, :+1: on this! I can't merge, but definitely would if I could at this point :)

abonas:
> I was thinking something that tested get_entities in isolation. But, the coverage in test guestbook is probably sufficient for now.

can be done separately in addition to this, adding tests is always welcome.

PR #92, opened 2015-06-04
Fixes #91: Move SSL options, basic auth & bearer token to constructor

jimmidyson:

Issue #81, opened 2015-05-11
Endpoints entity_type is not plural


stephenjlevy:
When creating an endpoint, Kubernetes requires the "kind" to be set to "Endpoints" instead of "Endpoint".  Changing "Endpoint" to "Endpoints" in the ENTITY_TYPES definition will not fix this problem because then a conflict arises when the pluralized version of the "get_endpoints" method is created.

stephenjlevy:
The easiest (albeit ugly) fix for this problem would be to have a special case for "endpoint" in the create_entity method that sets the kind to "Endpoints".

abonas:
@stephenjlevy, 
1. Have you opened an issue/discussion in k8s repo on this inconsistency? it's better to get this fixed on server side instead of each client's custom handling. cc @thockin 

2. What is your use case to manually create endpoints rather than use a selector and get them calculated and updated by the server? 
Is this something like the below example, or different?
https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/services.md#services-without-selectors

Thanks

cc @simon3z

stephenjlevy:
@abonas Yes, we need the ability to create external services (services without selectors).  Unfortunately, there's no way to declare the endpoints from within the service definition, so they must be declared manually.

I haven't opened an issue with the k8s developers about this...it's really a matter of semantics.  When creating an endpoint entity type, you really do have the ability to create multiple "endpoints", which is why I guess they made it plural, although technically, you're only creating one endpoint entity type.

thockin:
Yes, Endpoints objects are a set of individual IP endpoints.  It's perhaps
an unfortunate name in the hidden double-pluralness - technically the REST
path should be "endpointses" I guess, but I don't think we're going to do
that.

I'd MAYBE be open to a better name, but that is a post-1.0 change, at
earliest.

As for defining endpoints for external services, we toyed with embedding an
Endpoints into Service as an alternative to Selector, but we just ran out
of time to explore it further.

abonas:
@thockin we kind of worked in parallel :)
https://github.com/GoogleCloudPlatform/kubernetes/issues/8115

khaliullov:
@abonas so, how to be with Endpoints? I am not good in Ruby and I temporary created workaround like this:
https://github.com/khaliullov/kubeclient/commit/4d328b00ea057f269f75c60158e00c89b7b90502
but is so ugly.

abonas:
[...]

PR #106, opened 2015-08-09
Add resource quotas and limit ranges.

moolitayer:
Adding resource quotas & limit ranges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![Dependency Status](https://gemnasium.com/abonas/kubeclient.svg)](https://gemnasium.com/abonas/kubeclient)
 
 A Ruby client for Kubernetes REST api.
-The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces and endpoints.
+The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas and endpoints.
 The client currently supports Kubernetes REST api version v1beta3.
 
 ## Installation
@@ -118,7 +118,7 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , 'v1beta3',
 ## Examples:
 
 #### Get all instances of a specific entity type
-Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`
+Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`, `get_resource_quotas`
 
 ```ruby
 pods = client.get_pods
@@ -134,7 +134,7 @@ pods = client.get_pods(label_selector: 'name=redis-master,app=redis')
 ```
 
 #### Get a specific instance of an entity (by name)
-Such as: `get_service "service name"` , `get_pod "pod name"` , `get_replication_controller "rc name"`, `get_secret "secret name"`
+Such as: `get_service "service name"` , `get_pod "pod name"` , `get_replication_controller "rc name"`, `get_secret "secret name"`, `get_resource_quota "resource quota name"`
 
 The GET request should include the namespace name, except for nodes and namespaces entities.
 
@@ -159,7 +159,7 @@ client.delete_service "redis-service"
 ```
 
 #### Create an entity
-For example: `create_pod pod_object`, `create_replication_controller rc_obj`, `create_secret secret_object`
+For example: `create_pod pod_object`, `create_replication_controller rc_obj`, `create_secret secret_object`, `create_resource_quota resource_quota_object`
 
 Input parameter - object of type `Service`, `Pod`, `ReplicationController`.
 
@@ -177,9 +177,9 @@ client.create_service service`
 ```
 
 #### Update an entity
-For example: `update_pod`, `update_service`, `update_replication_controller`, `update_secret`
+For example: `update_pod`, `update_service`, `update_replication_controller`, `update_secret`, `update_resource_quota`
 
-Input parameter - object of type `Service`, `Pod`, `ReplicationController`
+Input parameter - object of type `Pod`, `Service`, `ReplicationController` etc.
 
 The below example is for v1beta3
 
@@ -188,7 +188,7 @@ client.update_service service1
 ```
 
 #### Get all entities of all types : all_entities
-Returns a hash with 8 keys (node, secret, service, pod, replication_controller, namespace, endpoint and event). Each key points to an EntityList of same type.
+Returns a hash with 9 keys (node, secret, service, pod, replication_controller, namespace, resource_quota, endpoint and event). Each key points to an EntityList of same type.
 
 This method is a convenience method instead of calling each entity's get method separately.
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![Dependency Status](https://gemnasium.com/abonas/kubeclient.svg)](https://gemnasium.com/abonas/kubeclient)
 
 A Ruby client for Kubernetes REST api.
-The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas and endpoints.
+The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas, limit ranges and endpoints.
 The client currently supports Kubernetes REST api version v1beta3.
 
 ## Installation
@@ -118,7 +118,7 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , 'v1beta3',
 ## Examples:
 
 #### Get all instances of a specific entity type
-Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`, `get_resource_quotas`
+Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`, `get_resource_quotas`, `get_limit_ranges`
 
 ```ruby
 pods = client.get_pods
@@ -134,7 +134,7 @@ pods = client.get_pods(label_selector: 'name=redis-master,app=redis')
 ```
 
 #### Get a specific instance of an entity (by name)
-Such as: `get_service "service name"` , `get_pod "pod name"` , `get_replication_controller "rc name"`, `get_secret "secret name"`, `get_resource_quota "resource quota name"`
+Such as: `get_service "service name"` , `get_pod "pod name"` , `get_replication_controller "rc name"`, `get_secret "secret name"`, `get_resource_quota "resource quota name"`, `get_limit_range "limit range name"`
 
 The GET request should include the namespace name, except for nodes and namespaces entities.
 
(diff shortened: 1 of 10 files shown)


moolitayer:
fyi @simon3z @abonas

README.md
```diff
@@ -118,7 +118,7 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , 'v1beta3',
 ## Examples:
 
 #### Get all instances of a specific entity type
-Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`
+Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`, `get_resource_quotas`, `get_limit_range`
```
abonas: why is this not plural?

abonas:
@moolitayer this PR's title is too generic. Please make it a clear one.

lib/kubeclient/inflections.rb
```diff
@@ -0,0 +1,4 @@
+ActiveSupport::Inflector.inflections do |inflect|
+  inflect.irregular 'resource_quota', 'resource_quotas'
+  inflect.irregular 'resourcequota', 'resourcequotas'
+end
```
abonas: @tenderlove @matthewd it seems that the inflector doesn't pluralize the word quota.
I saw it also here
http://codereply.com/answer/8hnvfd/method-pluralize-seems-working.html
I wanted to make sure that the above solution is the recommended one.
(Also based on what's written here https://github.com/rails/rails/blob/5373bf228d1273deae0ed03370ec4a63c580422b/activesupport/lib/active_support/inflector/methods.rb#L11)

test/test_limit_range.rb
```diff
@@ -0,0 +1,21 @@
...
+  def test_get_from_json_v3
+    stub_request(:get, %r{/limitranges})
+      .to_return(body: open_test_json_file('limit_range.json'),
+                 status: 200)
+
+    client = Kubeclient::Client.new 'http://localhost:8080/api/', 'v1'
+    limits = client.get_limit_range 'limits', 'quota-example'
+
+    assert_instance_of(Kubeclient::LimitRange, limits)
+    assert_equal('limits', limits.metadata.name)
```
abonas: nit, but the name is a bit confusing - you are pulling a single entity here

test/test_limit_range.rb
```diff
@@ -0,0 +1,21 @@
...
+    assert_equal('Container', limits.spec.limits[0].type)
+
```
abonas: no validation for the content of the limit?

lib/kubeclient/inflections.rb
matthewd: I *think* you can just do `inflect.irregular 'quota', 'quotas'`, and it'll handle it as a suffix.

But either way, yes, this looks right.

The only potential problem, is that the library is now registering custom pluralizations which could affect the surrounding application -- that's unlikely to be something a developer expects when adding your gem to their Gemfile. I don't know of any better solution for that, though.

abonas: > The only potential problem, is that the library is now registering custom pluralizations which could affect the surrounding application -- that's unlikely to be something a developer expects when adding your gem to their Gemfile. I don't know of any better solution for that, though.

so hardcoding it without messing with the inflector would be better?

matthewd: Yeah, if there's a way you can do that without uglifying your code too much, that's probably a more "pure" choice, for a library.

That said, in practice, I doubt you'd actually upset/inconvenience anyone by defining the inflection. ¯\\\_(ツ)_/¯

abonas: > Yeah, if there's a way you can do that without uglifying your code too much, that's probably a more "pure" choice, for a library.

oh it surely will uglify :)
because it must add an "if"

README.md
moolitayer: done.

test/test_limit_range.rb
moolitayer: fixed

test/test_limit_range.rb
moolitayer: fixed to limit_range

README.md
moolitayer: fixed.

test/test_limit_range.rb
moolitayer: added validations.

moolitayer:
@abonas 
>> @moolitayer this PR's title is too generic. Please make it a clear one.
what would you like to see?

lib/kubeclient/inflections.rb
moolitayer: @matthewd 
>> I think you can just do inflect.irregular 'quota', 'quotas', and it'll handle it as a suffix.
That works. thanks

@abonas it seems that this is the preferred solution (Inflector)?

abonas:
> what would you like to see?

for example, *which* entities are added in the patch

lib/kubeclient/inflections.rb
abonas: > @abonas it seems that this is the preferred solution (Inflector)?

@moolitayer yes
@matthewd thanks a lot for your inputs!!!

lib/kubeclient/inflections.rb
```diff
@@ -0,0 +1,3 @@
+ActiveSupport::Inflector.inflections do |inflect|
+  inflect.irregular 'quota', 'quotas'
```
simon3z: Is this global? Any way to scope this just for this gem? Maybe @matthewd or @tenderlove know that.

simon3z:
One question about the Inflector (inline), the rest looks OK :+1:

lib/kubeclient/inflections.rb
matthewd: [Nope](https://github.com/abonas/kubeclient/pull/106#discussion-diff-36607177)

simon3z:
The `active_support` inflection culprit is:

    inflect.plural(/([ti])a$/i, '\1a')

https://github.com/rails/rails/blob/3-2-stable/activesupport/lib/active_support/inflections.rb#L14

For the same reason mentioned here:

https://github.com/rails/rails/blob/5373bf228d1273deae0ed03370ec4a63c580422b/activesupport/lib/active_support/inflector/methods.rb#L11

>  # The Rails core team has stated patches for the inflections library will not
>  # be accepted in order to avoid breaking legacy applications which may be
>  # relying on errant inflections.

I'd prefer to have a local wrapper here just in our gem (to not affect interaction with other libraries), something like:

    def pluralize_entity(entity_name)
        return 'quotas' if entity_name == 'quota'
        entity_name.pluralize
    end

and then we'll just call our `pluralize_entity`.

moolitayer:
@abonas agreement on a local wrapper?

abonas:
> @abonas agreement on a local wrapper?

@simon3z @moolitayer 
I'm not a big fan of, plus @matthewd said it's not *that* bad to configure the inflector, but if you tend strongly towards this local solution, I won't block it.

moolitayer:
@simon3z @abonas 
In my opinion, as I've expressed to Alissa, the best solution would be to replace
```
%w(Pod Service ReplicationController Node Event Endpoint Namespace Secret ResourceQuota LimitRange)
```
with:
```
[['pod', 'pods'], ['Service', 'Services'] ...
```

Yes its more verbose but feels better for api end points then a text / localization solution with multiple pluralize/singularize calls.

Any way since we are 3 people with 3 different opinions any chance we agree to go with the current solution and patch it up to what ever solution in the future if need arises ?
(also I'm lazy)

jimmidyson:
Just noticed - isn't it `Endpoints`, not `Endpoint`? Type is `endpoints` - https://github.com/GoogleCloudPlatform/kubernetes/blob/master/pkg/api/v1/types.go#L1249

simon3z:
> Just noticed - isn't it Endpoints, not Endpoint? Type is endpoints -

Where exactly? we keep all the entities names singular and then we pluralize them when required.

     ENTITY_TYPES = %w(Pod Service ReplicationController Node Event Endpoint
                       Namespace Secret).map do |et|

jimmidyson:
@simon3z Which works for most types, but `Endpoints` is an exception: it's pluralized in the Kubernetes v1 API kind as linked above.

abonas:
> @simon3z Which works for most types, but Endpoints is an exception: it's pluralized in the Kubernetes v1 API kind as linked above.

@simon3z @jimmidyson known issue
https://github.com/abonas/kubeclient/issues/81

@moolitayer, I'm not in favor of copy-pasting all entities, so it's either the inflector configuration or the solution suggested by @simon3z . Both work for me.

simon3z:
> In my opinion, as I've expressed to Alissa, the best solution would be to replace with:
> 
>     [['pod', 'pods'], ['Service', 'Services'] ...
> 
> Yes its more verbose but feels better for api end points then a text / localization solution with multiple pluralize/singularize calls.

Fine by me if you prefer it and you think that code would be nicer than my proposal :+1:
(I have the feeling that you'll have to deal with uppercase/lowercase)

> I'm not a big fan of, plus @matthewd said it's not that bad to configure the inflector, but if you tend strongly towards this local solution, I won't block it.

Touching a singleton that is shared between multiple libraries is a big no-no for me. (Think how upset you would be if one of the other libraries would touch one of our pluralizations :smile:  )

simon3z:
I haven't invested much time in reviewing documentation and tests, we'd need another couples of eyes there. The core LGTM :+1:

moolitayer:
@simon3z @abonas 
1) fyi afaiu releasing a version of kubeclient might cause problems;
#92 removed the ssl_options method from common.rb which is used by openshift_client.
the dependency on kubeclient at openshift_client gemspec is stated as: 
spec.add_dependency 'kubeclient', '>= 0.1.17'
but we don't want old openshift_clients to get new kubeclient and break.
afaiu this means we should release kubeclient not before we have a new version of openshift_client.

2) after making the fix to openshift_clien I have not yet found a way to test it - I was unable to force openshift_client work with the new version of kubeclient using it's gemspec (like you can do :path => 'local/' in a gem file) so I am stuck. please advise

abonas:
> after making the fix to openshift_clien I have not yet found a way to test it - I was unable to force openshift_client work with the new version of kubeclient using it's gemspec (like you can do :path => 'local/' in a gem file) so I am stuck. please advise

have you tried bumping kubeclient dep in gemspec to 0.2.0 or 0.3.0 in openshift_gem?
Why do you need to use a local path to test it? 
https://github.com/abonas/kubeclient/pull/92 was released in one of the above releases, iirc 0.2.0

In any case, the way I know to test a local gem is as following (only for local tests, not to be commited): 
1. create a Gemfile in openshift client. 
2. Add there the dependency you want to kubeclient with the local path (gem kubeclient .....) . 
3. Add there also a reference to the gemspec (iirc just write there "gemspec"), 
4. remove the version requirement for kubeclient in gemspec.

simon3z:
@abonas LGTM :+1: ready for merge

README.md
```diff
@@ -177,9 +177,9 @@ client.create_service service`
 ```
 
 #### Update an entity
-For example: `update_pod`, `update_service`, `update_replication_controller`, `update_secret`
+For example: `update_pod`, `update_service`, `update_replication_controller`, `update_secret`, `update_resource_quota`, `update_limit_range`
 
-Input parameter - object of type `Service`, `Pod`, `ReplicationController`
+Input parameter - object of type `Pod`, `Service`, `ReplicationController` etc.
```
abonas: why is this change relevant to the patch and what is its purpose? (just swaping pod and service in places?)

moolitayer: to align with the text above: `update_pod`, `update_service`.
do you want this removed?

abonas: > to align with the text above: update_pod, update_service.
> do you want this removed?

in the future please separate changes like this to a separate patch since it is unrelated and it confuses when reviewing.
but nonetheless such alignment fixes are welcome.

PR #107, opened 2015-08-13
Add HTTP Proxy Wrapper

oschreib:
Introducing a new class which is used to access Kubernetes proxied
servies.
It's actually a wrapping around Net:HTTP with an additional prefix, so
all the requests made includes this prefix.


oschreib:
cc @simon3z @abonas

lib/kubeclient/common.rb
```diff
@@ -105,8 +96,20 @@ def watch_entities(entity_type, resource_version = nil)
           basic_auth_password: @basic_auth_password,
           headers: @headers
         }
+        options
```
abonas: why is this line needed?

abonas:
> servies.

s/servies/services
Also, please specify which proxied services is this about or add a link to the k8s relevant features

test/test_http_proxy.rb
```diff
@@ -0,0 +1,25 @@
+require 'test_helper'
+
+# Node entity tests
```
abonas: irrelevant comment

test/test_http_proxy.rb
```diff
@@ -0,0 +1,25 @@
...
+class TestHttpProxy < MiniTest::Test
+  def test_get_from_json_v3
+    path = 'http://localhost:8080/api/v1/proxy/namespaces/default/services/image-scanner:5001-tcp/image-scanner/api/scan'
+    stub_request(:post, path)
```
abonas: test name is test_get_* but the request type in the test is post - can you please explain?

lib/kubeclient/common.rb
```diff
@@ -202,6 +205,14 @@ def resource_name(entity_type)
         entity_type.pluralize.downcase
       end
 
+      def proxy_url_service(namespace, service, protocol)
```
jimmidyson: Do you mean `port` instead of `protocol`? Also this is optional for a single port service.

lib/kubeclient/http_proxy_wrapper.rb
```diff
@@ -0,0 +1,49 @@
+module Kubeclient
+  module Common
+    # Wrapper around Net:HTTP to support k8s services
+    class HttpProxyWrapper
```
abonas: Can you please specify what is the use case for post to those proxied services?
If both get and post are relevant, then some tests seem missing.

lib/kubeclient/http_proxy_wrapper.rb
```diff
@@ -0,0 +1,49 @@
...
+      def initialize(uri, options)
+        @uri = uri
+        @options = options
+      end
+
+      def get(path)
```
jimmidyson: What about all the other HTTP verbs, e.g. `PUT`? All need to be catered for.

lib/kubeclient/common.rb
jimmidyson: Also websocket support, which IIRC is supported by the kubernetes API server proxy (need to verify).

lib/kubeclient/common.rb
oschreib: It's not :)

test/test_http_proxy.rb
oschreib: Just copied that function name, nothing deep. Will change that.

lib/kubeclient/common.rb
oschreib: @jimmidyson nope. The usage of that function is proxy_url_service('default', 'image-scanner', '5001-tcp'), which builds the uri http://<host>:<port>/api/v1/proxy/namespaces/default/services/image-scanner:5001-tcp

I'm not sure how websockets are relevant here.

lib/kubeclient/http_proxy_wrapper.rb
oschreib: I thought about creating them when they'll be needed.
The general idea is to get an HTTP object from kubeclient which have already the proper path prefix (e.g api/v1/proxy/namespaces/default/services/image-scanner:5001-tcp), the problem is that the Net:HTTP object doesn't hold a path, so I created this one, which actually wraps the HTTP:Net object and a path.

lib/kubeclient/http_proxy_wrapper.rb
oschreib: see comment for @jimmidyson about the use case

lib/kubeclient/common.rb
jimmidyson: @oschreib So you're mixing port & protocol into one? Port isn't required btw by the  API server proxy so should be made optional. Nor is protocol...

lib/kubeclient/http_proxy_wrapper.rb
jimmidyson: How will you know when they're needed? Don't like the idea of adding them in one at a time when issues come in - should be simple enough to do in one go now I would have thought, delegating to a wrapped HTTP object?

lib/kubeclient/common.rb
oschreib: @jimmidyson tbh, I just looked at it as a string as it's part of the path.
Do you think it's better to get both parameters separably and optionally?

jimmidyson: @oschreib Yes I do. `def proxy_url_service(namespace, service, port, protocol)` makes more sense to me, only adding in to URL if port & protocol are specified.

lib/kubeclient/http_proxy_wrapper.rb
oschreib: @jimmidyson I'm not sure I understood what you meant: "should be simple enough to do in one go now I would have thought, delegating to....".

lib/kubeclient/common.rb
simon3z: This should be:

    proxy_url_service(namespace, service, port_name)

@jimmidyson the port/protocol names are irrelevant because you address them with a name, which in @oschreib example was in the format of `port-protocol`, but in general it could be anything "foobar", etc.

As far as I saw you're required to specify the `port_name` also when the port is only one.

jimmidyson: @simon3z No, port name (or number - both are acceptable) are definitely not required for single port services.

jimmidyson: Actually only port name is allowed, not number - sorry. But definitely not required for single port services (which are the norm remember).

jimmidyson: OK so you've just found helped me find a bug in OpenShift/Kubernetes :) Services with an unnamed single port (totally allowed) don't require a port name for proxying; if the port has a name, even if it's single port, then it's required for proxying.

test/json/http_proxy.json
```diff
@@ -0,0 +1,21 @@
+{
```
simon3z: @oschreib don't make this test so specific to image-scanner, also an "Hello World!" here would do.

test/test_http_proxy.rb
```diff
@@ -0,0 +1,41 @@
...
+                 status: 200)
+
+    @proxy.post('/image-scanner/api/scan',
+                '{"number" => 2, "scan" => ["docker.io/richxsl/rhel7"]}')
+    assert_requested(:post,
+                     @path,
+                     times: 1)
+  end
+
+  def test_get_from_json_v3
```
simon3z: The name of the method is not descriptive of what the test does (should be related to proxy).

lib/kubeclient/common.rb
```diff
@@ -202,6 +204,14 @@ def resource_name(entity_type)
         entity_type.pluralize.downcase
       end
 
+      def proxy_url_service(namespace, service, port_name)
```
simon3z: Let's try to support all the possible proxies (nodes, pods, services):

    /api/v1/proxy/nodes/{name}/{path:*}
    /api/v1/proxy/namespaces/{namespace}/pods/{name}/{path:*}
    /api/v1/proxy/namespaces/{namespace}/services/{name}/{path:*}

test/json/http_proxy.json
oschreib: Done

lib/kubeclient/common.rb
```diff
@@ -202,6 +204,14 @@ def resource_name(entity_type)
         entity_type.pluralize.downcase
       end
 
+      def proxy_url_service(type, service, port_name, namespace = '')
```
simon3z: I'd drop `_url` as you're not returning or passing any url.

abonas:
@simon3z @oschreib what is the status of this PR?

simon3z:
> @simon3z @oschreib what is the status of this PR?

@abonas at the moment we don't need it. I think we can close it in a week or so if things don't change.

PR #108, opened 2015-08-18
Add Component statuses

dkorn:
Adding the component status entity

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![Dependency Status](https://gemnasium.com/abonas/kubeclient.svg)](https://gemnasium.com/abonas/kubeclient)
 
 A Ruby client for Kubernetes REST api.
-The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas, limit ranges, endpoints, persistent volumes and persistent volume claims.
+The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas, limit ranges, endpoints, persistent volumes, persistent volume claims and component statuses.
 The client currently supports Kubernetes REST api version v1.
 
 ## Installation
@@ -118,7 +118,7 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , 'v1',
 ## Examples:
 
 #### Get all instances of a specific entity type
-Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`, `get_resource_quotas`, `get_limit_ranges`, `get_persistent_volumes`, `get_persistent_volume_claims`
+Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`, `get_resource_quotas`, `get_limit_ranges`, `get_persistent_volumes`, `get_persistent_volume_claims`, `get_component_statuses`
 
 ```ruby
 pods = client.get_pods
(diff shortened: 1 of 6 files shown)


dkorn:
@abonas @simon3z please review

README.md
```diff
@@ -134,7 +134,8 @@ pods = client.get_pods(label_selector: 'name=redis-master,app=redis')
 ```
 
 #### Get a specific instance of an entity (by name)
-Such as: `get_service "service name"` , `get_pod "pod name"` , `get_replication_controller "rc name"`, `get_secret "secret name"`
+Such as: `get_service "service name"` , `get_pod "pod name"` , `get_replication_controller "rc name"`, `get_secret "secret name"`,
+`get_component_status "component name"`
```
abonas: is component status retrieval also working per namespace?

abonas:
since current code as is adds deletion method for component status too (although not mentioned in the readme) - what would be k8s behavior in this case shall it receive a delete command on a component status?

jimmidyson:
FYI component status is currently broken in OpenShift - see openshift/origin#3667.

abonas:
> FYI component status is currently broken in OpenShift - see openshift/origin#3667.

@jimmidyson thanks for pointing this out :) . this looks like an ose deployment/arch specific issue only, no?
the feature is a core k8s feature -  is it operational on a plain k8s env?

jimmidyson:
Yes it should work fine on Kubernetes, but broken on OpenShift Origin & Enterprise.

abonas:
@dkorn just making sure you saw my review comments from last week

README.md
dkorn: > is component status retrieval also working per namespace?

it works *only* per namespace, but as I noticed other entities do as well. so, should I add ', "namespace"' or leave it as is (to maintain consistency)?

dkorn:
> since current code as is adds deletion method for component status too (although not mentioned in the readme) - what would be k8s behavior in this case shall it receive a delete command on a component status?

In the case one tries to delete a component status the operation fails and the following error appear: 
KubeException: HTTP status code 405, the server does not allow this method on the requested resource

dkorn:
@simon3z please review

simon3z:
LGTM :+1: cc @abonas

test/json/component_status_list.json
```diff
@@ -0,0 +1,52 @@
...
+          "type": "Healthy",
+          "status": "Unknown",
+          "error": "Get http://127.0.0.1:10251/healthz: dial tcp 127.0.0.1:10251: connection refused"
+        }
+      ]
+    },
+    {
+      "metadata": {
+        "name": "etcd-0",
+        "selfLink": "/api/v1/namespaces/componentstatuses/etcd-0",
```
abonas: this url suggests that component status is actually a name of a special namespace, rather than a resource of type component status on its own. 
is that the case?

abonas: it also doesn't make sense to me that k8s components running on a server are namespace scoped. how come etcd is namespace scoped? can you please explain it?

jimmidyson: @abonas - you're correct. See https://docs.openshift.org/latest/rest_api/kubernetes_v1.html#list-objects-of-kind-componentstatus for the real URL.

simon3z: Generation of `selfLink` is broken in several places sadly. That has to be addressed in kubernetes. @dkorn can you open an issue? (it shouldn't block us anyway)

abonas: so is this entity namespace scoped or not?

dkorn: As far as I understand, component statuses are resources/entities shared across the kubernetes cluster, not related to a specific namespace. 

one can list the component statuses in two ways (also according to swagger):
a. /api/v1/componentstatuses
b. /api/v1/namespaces/{namespace}/componentstatuses

or can get a specific component status: /api/v1/namespaces/{namespace}/componentstatuses/{name}

The selflink for a specific component status kubernetes currently provides in the response is broken. I will open an issue regarding it.

abonas: @dkorn please link the k8s issue here

simon3z: @abonas is the PR (not considering the `selfLink` kube bug) looking good to you?

abonas: > b. /api/v1/namespaces/{namespace}/componentstatuses

this doesn't make sense, it should return everything for every namespace? can you also open a question on k8s side if this is unclear?

dkorn: linking to the kubernetes issue:
https://github.com/kubernetes/kubernetes/issues/15618

simon3z: > b. /api/v1/namespaces/{namespace}/componentstatuses
> this doesn't make sense, it should return everything for every namespace? can you also open a question on k8s side if this is unclear?

This has already been shipped in v1 so they won't remove it (and anyway it's already out there in existing deployments). Kubeclient only "blindly" exposes entry points for entities and what they return, that said @abonas do you see any reason to block this PR?

abonas: no, I don't think it should be blocked.

PR #109, opened 2015-08-18
move validate_auth_options to common so it can be shared.

moolitayer:


abonas:
LGTM. @simon3z, if LGTY feel free to merge.

simon3z:
@moolitayer please check that this is enough to satisfy the `openshift_client` needs and we'll merge. (ping me when verified).

moolitayer:
verified https://github.com/abonas/openshift_client/pull/10/ with this change

PR #110, opened 2015-08-20
client: add compatibility with more_core_extensions

simon3z:
This patch adds the support to use kubeclient in applications that require more_core_extensions.
Since the attribute namespace is monkey patched by ObjectNamespace we're now forced to use the RecursiveOpenStruct table to retrieve the namespace value.


abonas:
:+1:

PR #111, opened 2015-08-23
add a changelog

abonas:


abonas:
@simon3z fyi

simon3z:
@abonas how should we maintain this? Can you document the procedure to get it generated? Thanks.

abonas:
@simon3z the manual way is rather simple:
gem install github_changelog_generator
and then:
https://github.com/skywinder/Github-Changelog-Generator#usage
(check there the note regarding the token needed to increase num of requests to github)
I do think it should be automatically generated, still haven't reached a conclusion where it's best to be added.
Opened https://github.com/abonas/kubeclient/issues/112 to track it.

abonas:
not relevant anymore as newer gems were released

PR #113, opened 2015-08-25
Add persistent volumes and persistent volume claims

zakiva:


zakiva:
@simon3z Please review

simon3z:
LGTM :+1: @abonas

abonas:
couple of functional questions to make sure I understand the features:
1. are volumes and/or claims retrieved per namespace? it looks from the tests that claims are but volumes are not?
2. since the patch adds support for update and delete, what is the behavior in those cases? can a volume or a claim be deleted?

simon3z:
> 1. are volumes and/or claims retrieved per namespace? it looks from the tests that claims are but volumes are not?

@abonas correct, looking at kubernetes api these are the relevant paths:

    GET|POST         /api/v1/persistentvolumeclaims
    GET|POST         /api/v1/namespaces/{namespace}/persistentvolumeclaims
    GET|PATCH|DELETE /api/v1/namespaces/{namespace}/persistentvolumeclaims/{name}
    GET|POST         /api/v1/persistentvolumes
    GET|PATCH|DELETE /api/v1/persistentvolumes/{name}

> 2. since the patch adds support for update and delete, what is the behavior in those cases? can a volume or a claim be deleted?

Yes update/delete are expected, but anyway what is allowed is up to the server to decide (for example as anything else it could be allowed in principles but denied to your user, even gets, etc.)
We should always support create, get, update and delete on entities and it's up to the server to reply with the relevant errors.

PR #114, opened 2015-08-26
move code that can be shared between clients to common

moolitayer:
This pr's main change is to move as much code as possible to common.rb so code can be shared between client as originally designed. A mixin is introduces to share code between common and clients.

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -10,8 +10,8 @@ require 'kubeclient/common'
 
 module Kubeclient
   # Kubernetes Client
-  class Client < Common::Client
-    attr_reader :api_endpoint
+  class Client
+    include ClientMixin
     # Dynamically creating classes definitions (class Pod, class Service, etc.),
     # The classes are extending RecursiveOpenStruct.
     # This cancels the need to define the classes
@@ -29,6 +29,8 @@ module Kubeclient
       [Kubeclient.const_set(et, clazz), et]
     end
 
+    ClientMixin.define_entity_methods(ENTITY_TYPES)
+
     def initialize(uri,
                    version = 'v1beta3',
                    ssl_options: {
@@ -44,39 +46,11 @@ module Kubeclient
                      bearer_token_file: nil
                    }
                   )
-
-      fail ArgumentError, 'Missing uri' if uri.nil?
-
-      validate_auth_options(auth_options)
-
-      handle_uri(uri, '/api')
-      @api_version = version
-      @headers = {}
-      @ssl_options = ssl_options
-      @auth_options = auth_options
-
-      if auth_options[:bearer_token]
-        bearer_token(auth_options[:bearer_token])
-      elsif auth_options[:bearer_token_file]
-        validate_bearer_token_file(auth_options[:bearer_token_file])
-        bearer_token(File.read(auth_options[:bearer_token_file]))
-      end
+      initialize_client(uri, '/api', version, ssl_options: ssl_options, auth_options: auth_options)
     end
 
     def all_entities
       retrieve_all_entities(ENTITY_TYPES)
-    end
-
-    define_entity_methods(ENTITY_TYPES)
-
-    private
-
-    def validate_bearer_token_file(bearer_token_file)
-      msg = "Token file #{bearer_token_file} does not exist"
-      fail ArgumentError, msg unless File.file?(bearer_token_file)
-
-      msg = "Cannot read token file #{bearer_token_file}"
-      fail ArgumentError, msg unless File.readable?(bearer_token_file)
     end
   end
 end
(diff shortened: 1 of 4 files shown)


abonas:
it's unclear to me what you are trying to do here.
I suggest as first step:
a. make a clear PR title that explains the purpose of the PR
b. squash the 5 commits to fewer, plus one of them is named wip
c. check why tests fail
cc @simon3z

moolitayer:
@abonas I've applied to your comments.

abonas:
@simon3z please review. if you think it's good to go you can merge and release.
I still don't understand why all those changes are required including the several renames to modules and parameters, and why most of the client need to be refactored, but perhaps you discussed this with @moolitayer and you have a better insight than me.

jimmidyson:
Does the change usage for existing users?

lib/kubeclient.rb
```diff
@@ -36,43 +38,18 @@ def initialize(uri,
...
-      if auth_options[:user]
-        @basic_auth_user = auth_options[:user]
-        @basic_auth_password = auth_options[:password]
-      elsif auth_options[:bearer_token]
-        bearer_token(auth_options[:bearer_token])
-      elsif auth_options[:bearer_token_file]
-        validate_bearer_token_file(auth_options[:bearer_token_file])
-        bearer_token(File.read(auth_options[:bearer_token_file]))
-      end
+      init_mixin(uri, '/api', version, ssl_options: ssl_options, auth_options: auth_options)
```
simon3z: @moolitayer rename to `initialize_client`

abonas:
> Does the change usage for existing users?

It looks like it does. for example auth option parameter 'user' was renamed
https://github.com/abonas/kubeclient/pull/114/files#diff-04c6e90faac2675aa89e2176d2eec7d8R75

moolitayer:
@jimmidyson yes: auth_options.name => auth_options.username

jimmidyson:
@moolitayer Not that it affects me but why the change? Seems redundant & would break stuff for some users (although I bet there's not many actually using basic auth).

simon3z:
All patches are easy to review singularly (except one, because of the new indentation)

- `rename auth_option s/user/username/` :+1:
- `list 'auth_options' keys in ctor` :+1:
- `store auth_options as hash instead of attributes` :+1:
- `move all common code to common.rb` :-1: a couple of comments (it actually touches few things, it's just diff artifacts)
- `changes in validate_auth_options` :+1:
- `add attr accessors` :+1:

cc @abonas @jimmidyson

simon3z:
> @moolitayer Not that it affects me but why the change? Seems redundant & would break stuff for some users (although I bet there's not many actually using basic auth).

@jimmidyson at the moment we kept breaking compatibility (e.g. when we moved the `ssl_options` to the constructor)
I acutally like the rename because it makes the options more consistent `username/password` vs. `user/password`.
For us it's better to change it now if possible.

Maybe @moolitayer can introduce a one-liner compatibility layer:

    @auth_options['username'] = @auth_options['user'] if @auth_options['user']

abonas:
> @jimmidyson at the moment we kept breaking compatibility (e.g. when we moved the ssl_options to the constructor)
> I acutally like the rename because it makes the options more consistent username/password vs. user/password.
> For us it's better to change it now if possible.
> 
> Maybe @moolitayer can introduce a one-liner compatibility layer:
> 
> @auth_options['username'] = @auth_options['user'] if @auth_options['user']

:+1:

moolitayer:
maintaining backward compatibility and test it. comment about it in commit.
expand commit message of add attr accessors

simon3z:
LGTM :+1: (despite the apparent large change in `move all common code to common.rb` which is just indentation diff artifacts). Let's queue this behind #115

lib/kubeclient/common.rb
```diff
@@ -1,255 +1,299 @@
...
+        result_hash[key_name] = send(method_name)
       end
+    end
 
-      def api
-        response = handle_exception do
-          create_rest_client.get(@headers)
-        end
-        JSON.parse(response)
+    def resource_name(entity_type)
```
simon3z: @moolitayer you mistakenly inverted the order of the two methods `resource_name` and `proxy_url`

`resource_name` was after `proxy_url`

moolitayer:
fix inverted methods

lib/kubeclient/common.rb
```diff
@@ -1,255 +1,299 @@
...
+
+    def initialize_client(
+      uri,
+      path,
+      version = nil,
+      ssl_options: {
+        client_cert: nil,
+        client_key: nil,
+        ca_file: nil,
+        verify_ssl: nil
```
simon3z: @moolitayer I think it's safer to set `verify_ssl` to `OpenSSL::SSL::VERIFY_PEER` here as well.

simon3z:
Other than the missing `OpenSSL::SSL::VERIFY_PEER` in `common.rb`, the rest LGTM :+1: ready for merge.

moolitayer:
@abonas @simon3z

simon3z:
@abonas LGTM :+1: ready for merge

PR #115, opened 2015-08-26
Add proxy URL

oschreib:
Introduce the ability to receive a full proxy URL from kubeclient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,14 @@ It is possible to interrupt the watcher from another thread with:
 watcher.finish
 ```
 
+#### Get a proxy URL
+You can get a complete URL for connecting a kubernetes entity via the proxy.
+
+```ruby
+client.proxy_url('services', 'srvname', '5001-tcp', 'ns')
+ => "https://localhost.localdomain:8443/api/v1/proxy/namespaces/ns/services/srvname:5001-tcp"
+```
+
 ## Contributing
 
 1. Fork it ( https://github.com/[my-github-username]/kubeclient/fork )
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -204,6 +204,16 @@ module Kubeclient
         end
       end
 
+      def proxy_url(kind, name, port, namespace = '')
+        ns_prefix = build_namespace_prefix(namespace)
+        # TODO: Change this once services supports the new scheme
+        if kind.to_s == 'pods'
+          rest_client["#{ns_prefix}#{kind}/#{name}:#{port}/proxy"].url
+        else
+          rest_client["proxy/#{ns_prefix}#{kind}/#{name}:#{port}"].url
+        end
+      end
+
       def resource_name(entity_type)
         self.class.pluralize_entity entity_type.downcase
       end
--- a/test/test_kubeclient.rb
+++ b/test/test_kubeclient.rb
@@ -406,6 +406,19 @@ class KubeClientTest < MiniTest::Test
     assert_equal(1, pods.size)
   end
 
+  def test_proxy_url
+    client = Kubeclient::Client.new 'http://host:8080', 'v1'
+    assert_equal('http://host:8080/api/v1/proxy/namespaces/ns/services/srvname:5001-tcp',
+                 client.proxy_url('services', 'srvname', '5001-tcp', 'ns'))
+
+    assert_equal('http://host:8080/api/v1/namespaces/ns/pods/srvname:5001-tcp/proxy',
+                 client.proxy_url('pods', 'srvname', '5001-tcp', 'ns'))
+
+    # Check no namespace provided
+    assert_equal('http://host:8080/api/v1/proxy/nodes/srvname:5001',
+                 client.proxy_url('nodes', 'srvname', 5001))
+  end
+
   private
 
   # dup method creates a shallow copy which is not good in this case


oschreib:
@simon3z  is that what you had in mind?

test/test_kubeclient.rb
```diff
@@ -406,6 +406,15 @@ def test_api_bearer_token_file_success
     assert_equal(1, pods.size)
   end
 
+  def test_proxy_url
+    client = Kubeclient::Client.new 'http://host:8080', 'v1'
+    assert_equal('http://host:8080/api/v1/proxy/namespaces/ns/services/srvname:5001-tcp',
+                 client.proxy_url('services', 'srvname', '5001-tcp', 'ns'))
+
+    assert_equal('http://host:8080/api/v1/proxy/pods/srvname:5001-tcp',
+                 client.proxy_url('pods', 'srvname', '5001-tcp'))
```
simon3z: Pods ports have no name iirc, it should just be an integer.

simon3z:
@oschreib :+1: in general but documentation is missing. @abonas

test/test_kubeclient.rb
oschreib: @simon3z what about service ports? I thought they does have a name.

Just to be clear - you just want  to remove the "-tcp" from the test in this case? or chaning the interface to receive integer as well?

test/test_kubeclient.rb
```diff
@@ -406,6 +406,15 @@ def test_api_bearer_token_file_success
     assert_equal(1, pods.size)
   end
 
+  def test_proxy_url
+    client = Kubeclient::Client.new 'http://host:8080', 'v1'
+    assert_equal('http://host:8080/api/v1/proxy/namespaces/ns/services/srvname:5001-tcp',
+                 client.proxy_url('services', 'srvname', '5001-tcp', 'ns'))
+
+    assert_equal('http://host:8080/api/v1/proxy/pods/srvname:5001-tcp',
```
jimmidyson: Is that actually a valid proxy URL? I thought it had to include the namespace like `/api/v1/proxy/namespaces/{namespace}/pods/{name}/{path:*}`? FYI on OpenShift this requires cluster-admin role to use. This is also deprecated.

Also, pod proxy has moved under the namespace now so is at `/api/v1/namespaces/{namespace}/pods/{name}/proxy/{path:*}`. FYI on OpenShift this can go through proper namespaced authzn so is preferred.

Sadly not the same for service proxying though, although there is an open issue on that.

test/test_kubeclient.rb
simon3z: Only this test.

simon3z: And BTW you can pass an integer here no problem.

test/test_kubeclient.rb
simon3z: @jimmidyson how do you specify what port to use with `/api/v1/namespaces/{namespace}/pods/{name}/proxy/{path:*}` ?

simon3z: Ah it seems that `/api/v1/namespaces/{namespace}/pods/{name}:{port}/proxy/{path:*}` is supported.

lib/kubeclient/common.rb
```diff
@@ -204,6 +204,11 @@ def retrieve_all_entities(entity_types)
         end
       end
 
+      def proxy_url(type, service, port_name, namespace = '')
+        ns_prefix = build_namespace_prefix(namespace)
+        rest_client["proxy/#{ns_prefix}#{type}/#{service}:#{port_name}"].url
```
simon3z: @oschreib let's move the proxy to `/api/v1/namespaces/{namespace}/pods/{name}:{port}/proxy/{path:*` as suggested by @jimmidyson

Leave services as it is (current url) and add a comment to switch services to the uri when ready.

test/test_kubeclient.rb
jimmidyson: Yeah that's right.

lib/kubeclient/common.rb
```diff
@@ -204,6 +204,16 @@ def retrieve_all_entities(entity_types)
         end
       end
 
+      def proxy_url(type, service, port_name, namespace = '')
+        ns_prefix = build_namespace_prefix(namespace)
+        # TODO: Change this once services supports the new scheme
+        if type == 'pods'
```
simon3z: @oschreib use `type.to_s` so we can also use symbols.

simon3z:
LGTM, @abonas @jimmidyson please review

lib/kubeclient/common.rb
```diff
@@ -204,6 +204,16 @@ def retrieve_all_entities(entity_types)
         end
       end
 
+      def proxy_url(type, service, port_name, namespace = '')
```
jimmidyson: Perhaps change `service` to `name`?

jimmidyson:
Other than that arg name, :+1:

lib/kubeclient/common.rb
simon3z: @jimmidyson @oschreib yes the names should probably be (to be consistent with kubernetes):

    proxy_url(kind, name, port, namespace = '')

abonas:
Can you put here a link to k8s documentation explaining the purpose of this feature?

lib/kubeclient/common.rb
```diff
@@ -204,6 +204,16 @@ def retrieve_all_entities(entity_types)
         end
       end
 
+      def proxy_url(kind, name, port, namespace = '')
+        ns_prefix = build_namespace_prefix(namespace)
+        # TODO: Change this once services supports the new scheme
+        if kind.to_s == 'pods'
```
simon3z: This should probably be: `kind.to_s.pluralize` it can be done in another patch though.

simon3z: Actually `pluralize_entity(kind.to_s)`

simon3z:
> Can you put here a link to k8s documentation explaining the purpose of this feature?

https://github.com/kubernetes/kubernetes/blob/release-1.0/docs/user-guide/accessing-the-cluster.md#manually-constructing-apiserver-proxy-urls
https://github.com/kubernetes/kubernetes/blob/release-1.0/docs/user-guide/connecting-to-applications-proxy.md

lib/kubeclient/common.rb
```diff
@@ -204,6 +204,16 @@ def retrieve_all_entities(entity_types)
         end
       end
 
+      def proxy_url(kind, name, port, namespace = '')
```
abonas: I think that it's confusing to call it "kind" and expect something pluralized from the user

README.md
```diff
@@ -212,6 +212,14 @@ It is possible to interrupt the watcher from another thread with:
 watcher.finish
 ```
 
+#### Get a proxy URL
+You can get a complete URL for connecting a kubernetes entity via the proxy.
+
+```ruby
+client.proxy_url('services', 'srvname', '5001-tcp', 'ns')
```
abonas: how can a user know that port is actually port AND protocol?

lib/kubeclient/common.rb
simon3z: @abonas yes in fact there was already this comment: https://github.com/abonas/kubeclient/pull/115#discussion_r38085297

README.md
simon3z: @abonas it's not port and protocol, it's either the port or the port name (ref: https://github.com/kubernetes/kubernetes/blob/release-1.0/pkg/api/v1/types.go#L1112), which by accident in this example was called with the convention `port-protocol`.

jimmidyson: I think port has to be only an int - don't think even port works & all proxies are tcp iirc.

abonas: > @abonas it's not port and protocol, it's either the port or the port name (ref: https://github.com/kubernetes/kubernetes/blob/release-1.0/pkg/api/v1/types.go#L1112), which by accident in this example was called with the convention port-protocol.

it should be clarified in readme and I suggest to change the example to something less confusing.

simon3z: > I think port has to be only an int - don't think even port works & all proxies are tcp iirc.

@jimmidyson for services you have to use the port name

simon3z: @jimmidyson e.g.: `/api/v1/proxy/namespaces/metrics/services/hawkular-metrics:http-endpoint/` to reach the hawkular metrics service.

jimmidyson: Yeah you're right just tested that here. Has to be name for service proxying (or no port at all if single port service).

Name doesn't work for pod proxying though, at least under the "new" location... Has to be no port or int. Name doesn't work.

Inconsistency ftw.

simon3z: @abonas @jimmidyson best compromise now is `port` for the argument name.

simon3z:
@abonas :+1: LGTM ready for merge IMO, @moolitayer will handle the pluralize issue in a next PR.

simon3z:
> it should be clarified in readme and I suggest to change the example to something less confusing.

cc @oschreib ref: https://github.com/abonas/kubeclient/pull/115#discussion_r38094296

simon3z:
> > it should be clarified in readme and I suggest to change the example to something less confusing.
> cc @oschreib ref: #115 (comment)

Let's address this in another patch @moolitayer

PR #116, opened 2015-08-27
proxy url: expect singular entity names & port changes

moolitayer:


simon3z:
cc @jimmidyson @abonas @oschreib

lib/kubeclient/common.rb
```diff
@@ -205,12 +205,13 @@ def retrieve_all_entities(entity_types)
       end
 
       def proxy_url(kind, name, port, namespace = '')
+        entity_name_plural = self.class.pluralize_entity(kind.to_s)
```
simon3z: @moolitayer  this may have collision with your #114 (`self.class`)

lib/kubeclient/common.rb
```diff
@@ -205,12 +205,13 @@ def retrieve_all_entities(entity_types)
...
         ns_prefix = build_namespace_prefix(namespace)
         # TODO: Change this once services supports the new scheme
-        if kind.to_s == 'pods'
-          rest_client["#{ns_prefix}#{kind}/#{name}:#{port}/proxy"].url
+        if entity_name_plural.to_s == 'pods'
```
simon3z: @moolitayer you don't need `to_s` here anymore

README.md
```diff
@@ -216,10 +216,19 @@ watcher.finish
...
- => "https://localhost.localdomain:8443/api/v1/proxy/namespaces/ns/services/srvname:5001-tcp"
+client.proxy_url('service', 'srvname', 'srvportname', 'ns')
+ => "https://localhost.localdomain:8443/api/v1/proxy/namespaces/ns/services/srvname:srvportname"
 ```
 
+Note the third parameter 'port' is a port name for services and an integer for pods:
+
+```ruby
+client.proxy_url('pod', 'podname', 5001, 'ns')
+ => "https://localhost.localdomain:8443/api/v1/proxy/namespaces/ns/pods/podname:5001"
```
simon3z: @moolitayer this should be `/api/v1/namespaces/ns/pods/podname:5001/proxy`

simon3z:
LGTM :+1: @abonas @jimmidyson ready for merge

jimmidyson:
:+1:

lib/kubeclient/common.rb
abonas: what if the user will pass  a capitalized version of kind? Pod/Service instead of pod/service?

abonas:
@simon3z  what if the user will pass  a capitalized version of kind? Pod/Service instead of pod/service?

abonas:
plus, if there's an interest in reviewing patches so more people should be able to have a look, they should not be merged within less than an hour from their posting.

simon3z:
> @simon3z what if the user will pass a capitalized version of kind? Pod/Service instead of pod/service?

@abonas as you know capitalized kinds are deprecated and it's not up to kubeclient to try and correct users requests (which in some cases my be legitimate: e.g. when used against an old kubernetes server where capitalized were allowed).

Anyway this was not brought up in the original PR #115 and since this PR was created only to address comments from there it's not relevant here.

> plus, if there's an interest in reviewing patches so more people should be able to have a look, they should not be merged within less than an hour from their posting.

Right, but again, what is introduced here is what was reviewed and agreed on #115 it's nothing new.

Do you have any comment on what to add or change?

moolitayer:
we also need support for services with only one service port in which case the name is optional and is from the url.

simon3z:
> we also need support for services with only one service port in which case the name is optional and is from the url.

Correct, but since it was not discussed it wasn't part of this PR. @moolitayer please open an issue.

PR #121, opened 2015-09-18
Implement Client#get_pod_log and #watch_pod_log

### LINKED ISSUE #119: Allow reading the log of a pod
There's an endpoint for it: http://kubernetes.io/v1.0/docs/api-reference/operations.html#_read_log_of_the_specified_pod

I can't figure out how to use kubeclient to do it, though.

jonmoter:
This allows a user to retrieve (or stream) the logs for a particular pod, via the Kubernetes API.
Resolves #119.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,30 @@ client.proxy_url('pod', 'podname', 5001, 'ns')
 ```
 
 
+#### Get the logs of a pod
+You can get the logs of a running pod:
+
+```ruby
+client.get_pod_log('pod-name', 'default')
+ => "Running...\nRunning...\nRunning...\n"
+```
+
+If that pod has more than one container, you must specify the container:
+
+```ruby
+client.get_pod_log('pod-name', 'default', container: 'ruby')
+ => "..."
+```
+
+You can also watch the logs of a pod to get a stream of data:
+
+```ruby
+watcher = client.watch_pod_log('pod-name', 'default', container: 'ruby')
+watcher.each do |line|
+  puts line
+end
+```
+
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -251,6 +251,14 @@ client.get_pod_log('pod-name', 'default', container: 'ruby')
  => "..."
 ```
 
+If a container in a pod terminates, a new container is started, and you want to
+retrieve the logs of the dead container, you can pass in the `previous` option:
+
+```ruby
+client.get_pod_log('pod-name', 'default', previous: true)
+ => "..."
+```
+
 You can also watch the logs of a pod to get a stream of data:
 
 ```ruby
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -50,7 +50,7 @@ module Kubeclient
         json_error_msg = {}
       end
       err_message = json_error_msg['message'] || e.message
-      raise KubeException.new(e.http_code, err_message)
+      raise KubeException.new(e.http_code, err_message, e.response)
     end
 
     def handle_uri(uri, path)
@@ -135,20 +135,7 @@ module Kubeclient
         uri.query = URI.encode_www_form('resourceVersion' => resource_version)
       end
 
-      options = {
-        use_ssl: uri.scheme == 'https',
-        ca_file: @ssl_options[:ca_file],
-        # ruby Net::HTTP uses verify_mode instead of verify_ssl
-        # http://ruby-doc.org/stdlib-1.9.3/libdoc/net/http/rdoc/Net/HTTP.html
-        verify_mode: @ssl_options[:verify_ssl],
-        cert: @ssl_options[:client_cert],
-        key: @ssl_options[:client_key],
-        basic_auth_user: @auth_options[:username],
-        basic_auth_password: @auth_options[:password],
-        headers: @headers
-      }
-
-      Kubeclient::Common::WatchStream.new(uri, options)
+      Kubeclient::Common::WatchStream.new(uri, net_http_options(uri))
     end
 
     def get_entities(entity_type, klass, options = {})
@@ -253,6 +240,29 @@ module Kubeclient
       end
     end
 
+    def get_pod_log(pod_name, namespace = nil, container: nil, previous: false)
+      params = { params: {} }
+      params[:params][:previous] = true if previous
+      params[:params][:container] = container if container
+
+      ns = build_namespace_prefix(namespace || 'default')
+      handle_exception do
+        rest_client[ns + "pods/#{pod_name}/log"]
+          .get(params.merge(@headers))
+      end
+    end
+
+    def watch_pod_log(pod_name, namespace = nil, container: nil)
+      params = 'follow'
+      params += "&container=#{container}" if container
+
+      ns = build_namespace_prefix(namespace || 'default')
+      uri = @api_endpoint
+            .merge("#{@api_endpoint.path}/#{@api_version}/#{ns}/pods/#{pod_name}/log?#{params}")
+
+      Kubeclient::Common::WatchStream.new(uri, net_http_options(uri))
+    end
+
     def resource_name(entity_type)
       ClientMixin.pluralize_entity entity_type.downcase
     end
@@ -294,6 +304,28 @@ module Kubeclient
 
       msg = "Cannot read token file #{@auth_options[:bearer_token_file]}"
       fail ArgumentError, msg unless File.readable?(@auth_options[:bearer_token_file])
+    end
+
+    def net_http_options(uri)
+      options = {
+        basic_auth_user: @auth_options[:username],
+        basic_auth_password: @auth_options[:password],
+        headers: @headers
+      }
+
+      if uri.scheme == 'https'
+        options.merge!(
+          use_ssl: true,
+          ca_file: @ssl_options[:ca_file],
+          cert: @ssl_options[:client_cert],
+          key: @ssl_options[:client_key],
+          # ruby Net::HTTP uses verify_mode instead of verify_ssl
+          # http://ruby-doc.org/stdlib-1.9.3/libdoc/net/http/rdoc/Net/HTTP.html
+          verify_mode: @ssl_options[:verify_ssl]
+        )
+      end
+
+      options
     end
   end
 end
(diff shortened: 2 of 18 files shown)


README.md
```diff
@@ -228,6 +228,30 @@ client.proxy_url('pod', 'podname', 5001, 'ns')
...
+
+```ruby
+client.get_pod_log('pod-name', 'default')
+ => "Running...\nRunning...\nRunning...\n"
+```
+
+If that pod has more than one container, you must specify the container:
+
+```ruby
+client.get_pod_log('pod-name', 'default', container: 'ruby')
```
abonas: Is this documented somewhere on k8s side that a container must be specified if there's more than one?

lib/kubeclient/common.rb
```diff
@@ -136,20 +136,7 @@ def watch_entities(entity_type, resource_version = nil)
         uri.query = URI.encode_www_form('resourceVersion' => resource_version)
       end
 
-      options = {
```
abonas: Is this refactoring related to the functionality added in the PR?

lib/kubeclient/common.rb
```diff
@@ -254,6 +241,29 @@ def proxy_url(kind, name, port, namespace = '')
       end
     end
 
+    def get_pod_log(pod_name, namespace = nil, container: nil, previous: false)
+      params = { params: {} }
+      params[:params][:previous] = true if previous
+      params[:params][:container] = container if container
+
+      ns = build_namespace_prefix(namespace || 'default')
```
abonas: the namespace is not defaulted to 'default' in other methods in the client, and I am thinking why should we start doing it here. If the user is not specifying correctly the namespace, the server will send response accordingly.

test/test_kubeclient.rb
```diff
@@ -4,6 +4,10 @@ def open_test_json_file(name)
   File.new(File.join(File.dirname(__FILE__), 'json', name))
 end
 
+def open_test_text_file(name)
```
abonas: can it be unified with the open_test_json_file? a method can receive both the name and the needed extension. it will reduce code duplication.

README.md
jonmoter: I haven't found documentation in the API that states it is necessary. But if I try running `kubectl logs` on a pod that has more than one container, I get this error message:

```
$ kubectl logs account-service-rc-ci2bc
error: POD account-service-rc-ci2bc has more than one container; please specify the container to print logs for
```

lib/kubeclient/common.rb
jonmoter: Yes.  The new `#watch_pod_log` method needs these options as well, so I extracted them out into the `#net_http_options` method to DRY up the code.

lib/kubeclient/common.rb
jonmoter: Okay, good point.  I can remove that.

Also, now that PR #105 is merged, I'll change `namespace` to be part of the options hash, rather than a regular parameter.

test/test_kubeclient.rb
jonmoter: Okay, can do.

lib/kubeclient/common.rb
jonmoter: Actually, given how the logs endpoint is designed, it's impossible for it to succeed if you don't specify a namespace.  The URL for the endpoint is `/api/v1/namespaces/{namespace}/pods/{name}/log`.

So what do you think of me changing the method signature to this?

```ruby
def get_pod_log(pod_name, namespace:, container: nil, previous: false)
  ...
end
```

That acts as documentation that it's required, and if you don't specify it, you'll immediate get:

`ArgumentError: missing keyword: namespace`

jonmoter: Nevermind.  That syntax doesn't work with ruby 2.0.x.

lib/kubeclient/watch_stream.rb
```diff
@@ -19,12 +19,13 @@ def each
 
         @http.request(request) do |response|
           unless response.is_a? Net::HTTPSuccess
-            fail KubeException.new(response.code, response.message)
+            fail KubeException.new(response.code, response.message, response)
           end
+          is_json = response['Content-Type'] == 'application/json'
```
jonmoter: Arg, I just discovered this doesn't necessarily work with all watch endpoints.  Coming up with a new solution...

jonmoter:
Okay, I fixed the bug with watch_entitites.  Should be ready for review again.

lib/kubeclient/common.rb
```diff
@@ -253,6 +240,29 @@ def proxy_url(kind, name, port, namespace = '')
       end
     end
 
+    def get_pod_log(pod_name, namespace, container: nil, previous: false)
```
abonas: Hi, 
What is the previous param standing for? didn't notice it above in the readme.

abonas:
left a question regarding one param, other than that it looks good.
@simon3z please review when you're back, it touches the watch part which you are much more familiar with.

lib/kubeclient/common.rb
jonmoter: According to the Kubernetes API docs, "Return previous terminated container logs" if true.

I believe it's for a case where a container terminates in a pod, and is automatically restarted. This is a mechanism to get the logs of the terminated container, rather than the currently running container.

abonas: > According to the Kubernetes API docs, "Return previous terminated container logs" if true.
> 
> I believe it's for a case where a container terminates in a pod, and is automatically restarted. This is a mechanism to get the logs of the terminated container, rather than the currently running container.

alright, please either link in the readme to the k8s doc, or just add it inline - whatever you prefer.
I would like the readme to be as clear as possible to the gem users.
the PR LGTM,
@simon3z please review, especially the watch part. (and you had a comment in the linked issue on k8s functionality regarding logs, please explain it, in case your comment applies to this patch)

jonmoter: Okay, I added an explanation to the README

abonas:
@simon3z this is waiting for your review

simon3z:
@jonmoter this PR is trying to do too many things at once, which could be ok if the changes were properly split in different patches (so that they could be reviewed separately).

What you do (all at once) is at least:

* add `response` to the KubeException initializer
* cleanup the `options` creating `net_http_options`
* add `get_pod_log` (with tests and documentation)
* add `format` to `WatchStream`

Keeping in mind all these changes in just one large diff is error-prone and also prevents us from committing parts that are already good enough to be merged.

Overall the PR seems ok but I am not confident enough that watches won't be broken (the json part), could you add a test for `watch_pod` as well?

Meanwhile if you can send separate pull requests for:

* add `response` to the KubeException initializer
* cleanup the `options` creating `net_http_options`
* add `format` to `WatchStream`

those can be reviewed and merged at a faster pace.

Good work, thanks!

lib/kubeclient/watch_stream.rb
```diff
@@ -4,10 +4,11 @@ module Kubeclient
   module Common
     # HTTP Stream used to watch changes on entities
     class WatchStream
-      def initialize(uri, options)
+      def initialize(uri, http_options, format: :json)
         @uri = uri
         @http = nil
-        @options = options.merge(read_timeout: nil)
+        @options = http_options.merge(read_timeout: nil)
```
simon3z: @jonmoter at this point it's probably worth renaming also `@options` to `@http_options`

simon3z:
@jonmoter in reference to:

    +          is_json = response['Content-Type'] == 'application/json'
> Arg, I just discovered this doesn't necessarily work with all watch endpoints. Coming up with a new solution...

Do you have more details? What was the problem you found?

jonmoter:
Adding `format` to `WatchStream` and refactoring into `net_http_options` isn't necessary and doesn't make much sense, except that it's required for the `{get|watch}_pod_log` functionality. I think if I made PRs for those changes by themselves, they'd be more atomic, but they wouldn't be useful changes without the context of the new log functionality.

Adding `response` to the `KubeException` initializer does stand on its own as useful added functionality, so I agree that would be good to split out.

If you insist I break this change up into multiple PRs, I'm willing to do that.  But for the record, I'm kindof frustrated that 20 days after opening this PR, and getting okay'ed from one maintainer, I'm now being asked to break up and reorganize my contributions.

jonmoter:
>> Arg, I just discovered this doesn't necessarily work with all watch endpoints. Coming up with a new solution...

> Do you have more details? What was the problem you found?

If you make a request to get a single object, like `/api/v1/pods/{pod-name}`, you get a response with Content-Type "application/json", which contains a single JSON hash.  That's great.

But if you request a watch endpoint, like `/api/v1/watch/pods`, you get a response with Content-Type "text/plain; charset=utf-8", containing a stream of JSON hashes separated by carriage returns.  So I couldn't depend on the response content-type to know whether to JSON parse each line or not.

simon3z:
> Adding format to WatchStream and refactoring into net_http_options isn't necessary and doesn't make much sense, except that it's required for the {get|watch}_pod_log functionality. I think if I made PRs for those changes by themselves, they'd be more atomic, but they wouldn't be useful changes without the context of the new log functionality.

Refactoring happens all the time and it makes sense on its own because improves code quality.
`net_http_options` is a good cleanup and improves code cleanliness so it makes sense. 

> If you insist I break this change up into multiple PRs, I'm willing to do that. But for the record, I'm kindof frustrated that 20 days after opening this PR, and getting okay'ed from one maintainer, I'm now being asked to break up and reorganize my contributions.

I understand that is frustrating but it happens all the time in open source (to put it in prospective try to see other projects), large patches tend to have a longer lifecycle, that's why it's in your best interest to split them.

I am sure that if you had separate PRs for each small step they would have been already merged and now you would be halfway through with only the real meat of `get_pod_log` pending to be reviewed and merged.

Also, if you read my message carefully, I said that the only thing I am not confident of is that the watches are not broken and I asked you to add a test (which should be very similar to `test_watch_pod_log`, but using json). To split the patches was a suggestion for you to carry less changes on your side (as they would be merged quickly).

For example also the `open_test_file` change would be something that can be merged separately and it would be less burden for you to keep rebasing it.

jonmoter:
Okay, I mistook what you were saying in that comment.  I'll add a test and break this PR up into smaller ones.

jonmoter:
Closing in favor of a few smaller PRs.

jonmoter:
Opened smaller PRs to prep for pod logging: #129, #130, #131, and #132.

Once those are merged, I'll create a PR containing the logic for getting and watching pod logs.

PR #124, opened 2015-10-06
Add service account entity

simon3z:
This pull request adds the support for the `service accounts` and the `security context constraints` entities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![Dependency Status](https://gemnasium.com/abonas/kubeclient.svg)](https://gemnasium.com/abonas/kubeclient)
 
 A Ruby client for Kubernetes REST api.
-The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas, limit ranges, endpoints, persistent volumes, persistent volume claims and component statuses.
+The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas, limit ranges, endpoints, persistent volumes, persistent volume claims, component statuses and service accounts.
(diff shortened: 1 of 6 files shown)


simon3z:
cc @abonas

abonas:
Is SecurityContextConstraint a k8s feature or OpenShift feature? I don't see it in k8s docs/api reference

jimmidyson:
OpenShift only atm, although at some point I believe they are hoping to merge it upstream.

abonas:
> OpenShift only atm, although at some point I believe they are hoping to merge it upstream.

thanks @jimmidyson . so if that's the case, this entity should be added at https://github.com/abonas/openshift_client and not here

simon3z:
@deads2k how come `/api/v1/securitycontextconstraints` in OpenShift is exposed as a Kubernetes api? (Instead of `/oapi`) Is it an exception? are there other exceptions as this one?

deads2k:
> @deads2k how come /api/v1/securitycontextconstraints in OpenShift is exposed as a Kubernetes api? (Instead of /oapi) Is it an exception? are there other exceptions as this one?

We were trying to get it merged upstream before 3.0.  We thought it was going to make it in, but it never did.  In order to support backwards compatibility, we've been carrying it ever since.

There aren't any other entire types that are under `/api`, but only exist in origin.  However, we do run ahead on some security related features, so some we have a few extra fields on certain types.  The swagger for whatever server you're talking to should be up to date.

Kube has also added some resource discovery APIs.  I'll try to look up examples today.  If I forget, remind me tomorrow.

simon3z:
Thanks @deads2k I remember you warned us about it, but I thought backward compatibility was only about the body of the entities.

PR #125, opened 2015-10-06
common: add support for the ca cert_store

simon3z:
This patch adds the support for the cert_store as an alternative to the ca_file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,19 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , "v1",
                                 ssl_options: ssl_options
 ```
 
+As an alternative to the `ca_file` it's possible to use the `cert_store`:
+
+```ruby
+cert_store = OpenSSL::X509::Store.new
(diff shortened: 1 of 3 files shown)


abonas:
@simon3z please rebase

abonas:
How does this behave when passing both the ca_file and the store?

simon3z:
> How does this behave when passing both the ca_file and the store?

It uses the same semantic of `net/http` and `OpenSSL`, we're not doing anything with it.

https://github.com/ruby/ruby/blob/v2_2_0/ext/openssl/ossl_ssl.c#L693-L704
https://github.com/ruby/ruby/blob/v2_2_0/ext/openssl/ossl_ssl.c#L750-L757
https://www.openssl.org/docs/manmaster/ssl/SSL_CTX_load_verify_locations.html

(the semantic is to consider them all, `cert_store`, `ca_file` and `ca_path`)

abonas:
LGTM

PR #126, opened 2015-10-06
get or watch entities using fieldSelector or labelSelector

### LINKED ISSUE #123: Allow watching events with a labelSelector
I want to be able to subscribe to watch events for pods that match a label selector, which is supported by the Kubernetes API, but not the client gem.

e.g. you can make a call to `/api/v1/watch/pods?labelSelector=foo%3Dbar` to get events for all pods that match the selector "foo=bar".

I'd like to add an `options` hash to `#watch_entities`, allowing you to pass in optional values for the labelSelector and fieldSelector parameters that the API accepts.

jonmoter:
This PR adds options to the `get_*` and `watch_*` methods of Client, so you can specify `:label_selector` and `:field_selector` when calling those.

This PR is built on top of PR #121, so we should merge the other PR first.

Resolves #123.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -79,8 +79,12 @@ module Kubeclient
 
         # watch all entities of a type e.g. watch_nodes, watch_pods, etc.
         define_method("watch_#{entity_name_plural}") \
-        do |resource_version = nil|
-          watch_entities(entity_type, resource_version)
+        do |options = {}|
+          # This method used to take resource_version as a param, so
+          # this conversion is to keep backwards compatibility
+          options = { resource_version: options } unless options.is_a?(Hash)
+
+          watch_entities(entity_type, options)
         end
 
         # get a single entity of a specific type by name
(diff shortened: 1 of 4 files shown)


abonas:
@simon3z please review.
@jonmoter looks like this needs a rebase

jonmoter:
@abonas, rebase complete

abonas:
@jonmoter thanks, @simon3z it's all yours for review :)

lib/kubeclient/common.rb
```diff
@@ -126,28 +130,41 @@ def rest_client
...
+    #   :name - the name of the entity to watch.
+    #   :label_selector - a selector to restrict the list of returned objects by their labels.
+    #   :field_selector - a selector to restrict the list of returned objects by their fields.
+    #   :resource_version - shows changes that occur after that particular version of a resource.
+    def watch_entities(entity_type, options = {})
+      ns = build_namespace_prefix(options[:namespace])
+
+      uri = @api_endpoint.dup
+      uri.path += "/#{@api_version}/watch/#{ns}#{resource_name(entity_type.to_s)}"
+      uri.path += "/#{options[:name]}" if options[:name]
```
simon3z: Minor (no need to change unless you fully agree) to be consistent with the rest of the code it should have been:

    path =  "/#{@api_version}/watch/#{ns}#{resource_name(entity_type.to_s)}"
    path += "/#{options[:name]}" if options[:name]
    uri = @api_endpoint.merge(path)

simon3z:
@jonmoter I would have split the patches in:

- add `options` to `watch_entities`
- add `field_selector` (both get and watch)

It would have been easier to review (no need to do it now anyway).

Can you check what we should do with the `test_api_bearer_token_with_params_success` test? Is it now covered by your `test_entities_with_selector`?

Can you add also a `field_selector` (and maybe a `resource_version`) if relevant to `get` as well? Thanks!

Other than that LGTM :+1:

lib/kubeclient/common.rb
jonmoter: Fair point. I'll update for consistency.

jonmoter:
@simon3z okay, I updated the code for consistency, per your comment.

I think we should keep `test_api_bearer_token_with_params_success`, since that's checking that passing in both a header and a parameter both work. The fact that's a `label_selector` is incidental.

> Can you add also a `field_selector` (and maybe a `resource_version`) if relevant to `get` as well? 

According to the API docs, `field_selector` and `label_selector` are only accepted params for endpoints that return a list of entities, rather than a single one specified by name. I assume that's what you mean when you say `get`?

simon3z:
> According to the API docs, field_selector and label_selector are only accepted params for endpoints that return a list of entities, rather than a single one specified by name. I assume that's what you mean when you say get?

Right. LGTM :+1: cc @abonas

Issue #122, opened 2015-09-18
Parse Kubernetes config file to create clients


jonmoter:
When you run the Kubernetes installer on Vagrant, it creates a YAML file for you in your home directory called `~/.kube/config`, that contains the same information if you ran `kube config view`.  For example:

```
$ kubectl config view
apiVersion: v1
clusters:
- cluster:
    server: http://127.0.0.1:9999
  name: kubepoc
- cluster:
    certificate-authority-data: REDACTED
    server: https://10.245.1.2
  name: vagrant
contexts:
- context:
    cluster: kubepoc
    user: ""
  name: kubepoc
- context:
    cluster: vagrant
    user: vagrant
  name: vagrant
current-context: vagrant
kind: Config
preferences: {}
users:
- name: vagrant
  user:
    client-certificate-data: REDACTED
    client-key-data: REDACTED
    password: vagrant
    username: vagrant
```

For a project I'm working on, I wrote up a class that parses that config file and can construct a `Kubeclient::Client` based on that information.  For example:

```ruby
config = ClientConfigFile.new(ENV['HOME'] + '/.kube/config')
client = config.client_for('vagrant')
client.get_pods
```

Would you like me to open up a PR that contains that code?  I figure it would be something generally useful for people using this gem.

abonas:
@jonmoter, my preference at this point is not to contain such configuration and parsing as part of the client.
There are several different ways to configure kubernetes, and by adding them here,  they will need maintenance and support as part of the client, and I would like at this point to avoid it.

jonmoter:
Okay, fair enough.

PR #127, opened 2015-10-07
config: add support to read kube config files

simon3z:
This pull requests add the support for parsing kube config files in order to create Kubeclient connections with the correct context (certificates, etc.).


simon3z:
I will add more information in the README file asap.

jonmoter:
FYI, I offered to implement this in issue #122, and @abonas said she'd rather not have that logic in the client.

Personally, I'm all in favor of it.  But you two might want to sync up.  :smile:

lib/kubeclient/config.rb
```diff
@@ -0,0 +1,111 @@
...
+  class Config
+    class Context
+      attr_reader :api_endpoint, :api_version, :ssl_options
+
+      def initialize(api_endpoint, api_version, ssl_options)
+        @api_endpoint = api_endpoint
+        @api_version = api_version
+        @ssl_options = ssl_options
+      end
+    end
```
jonmoter: For convenience, it would be nice to have a `#client` method of context that would return a `Kubeclient::Client`, so you could say:

```ruby
config = Kubeclient::Config.read(ENV['HOME'] + '/.kube/config')
client = config.context('vagrant').client
```

simon3z: @jonmoter yes that would have been nice but since this part should work also with `openshift_client` we should come up with something smarter (what client to return, etc.). At the moment I went with the simple implementation where we don't prepare any of the two clients.

jonmoter:
Any word on whether we're going to move forward with this PR?  I'd love to remove the code I have in my project that handles all the config file parsing in favor of using this one.  :-)

simon3z:
> Any word on whether we're going to move forward with this PR? I'd love to remove the code I have in my project that handles all the config file parsing in favor of using this one. :-)

@abonas let's discuss this when you're available. I assume it could happen toward end of next week.
@jonmoter would it work for you or do you need a quicker answer?

jonmoter:
That's fine, no huge rush. Thanks!

garethr:
@simon3z thanks for writing, I was just thinking the same and hoping someone had got there first. So definitely useful. This could also be released as a separate gem if it doesn't belong in the client.

jonmoter:
Hey, following up on this, since the PR has been open since October. @simon3z, any idea whether this functionality will get merged into kubeclient?

simon3z:
> Hey, following up on this, since the PR has been open since October. @simon3z, any idea whether this functionality will get merged into kubeclient?

@garethr @jonmoter we're lacking of patch reviews. If you have no comments on the code, then drop a "LGTM". I'll fix the travis complaints asap.

jonmoter:
LGTM!  👍

garethr:
I've been running with this vendored in the garethr-kubernetes module, and have had a few changes to support absolute paths and also to support bearer tokens for auth. Relevant commits: 

https://github.com/garethr/garethr-kubernetes/commit/55ebfcc1ef6df8e1b9278a3c7c3c4422de0b97f8
https://github.com/garethr/garethr-kubernetes/commit/298cf48e7e8f235d4cedf0a0ea0be67dee38400e

I'm happy to submit a PR after this is merged to address those, or feel free to apply them beforehand.

simon3z:
@abonas it seems that there are (unrelated?) VCR failures, do you have any insight? Thanks.

simon3z:
@garethr @jonmoter feel free to chime in if you have an idea on why the tests are failing :smile: (I can't debug it now). Thanks.

abonas:
no insights now, will have a look on Sunday.

lib/kubeclient/config.rb
```diff
@@ -0,0 +1,113 @@
...
+
+    def initialize(kcfg, kcfg_path)
+      @kcfg = kcfg
+      @kcfg_path = kcfg_path
```
abonas: why both kcfg and kcfg path are needed separately?

abonas:
@simon3z I don't understand why it fails based on PR content -I see those tests expect username:password as part of url but they are blank. Are you rebased?

simon3z:
> @simon3z I don't understand why it fails based on PR content -I see those tests expect username:password as part of url but they are blank. Are you rebased?

@abonas so it seems by looking at the branch:

https://github.com/simon3z/kubeclient/commits/kubeconfig-files


```
config: add support to read kube config files ￼…
simon3z committed on Oct 6, 2015
￼ 77e4c6a  
Commits on Apr 4, 2016
￼
Merge pull request #162 from simon3z/bump-1.1.3 ￼…
simon3z committed 28 days ago
￼ 82c5fab  
```

abonas:
@simon3z I think it should be solved by locking the webmock as done in #165 
saw this issue that might be related: https://github.com/bblimke/webmock/issues/591

jonmoter:
Tests appear to be passing now.  Good to merge?

lib/kubeclient/config.rb
```diff
@@ -0,0 +1,113 @@
...
+        break x['cluster'] if x['name'] == context['cluster']
+      end
+
+      fail "Unknown cluster #{context['cluster']}" unless cluster
+
+      user = @kcfg['users'].detect do |x|
+        break x['user'] if x['name'] == context['user']
+      end
+
+      fail "Unknown user #{context['user']}" unless cluster
```
jonmoter: I think you mean to test `user` instead of `cluster` here.

That being said, I believe it's valid in a kubeconfig file for a context to have a user of "" or nil. So this shouldn't throw an error.

jonmoter:
Okay, here's a branch we can use: https://github.com/zendesk/kubeclient/tree/samson-gem-branch

lib/kubeclient/config.rb
jonmoter: Here's a commit you can cherry-pick that handles this scenario:

https://github.com/zendesk/kubeclient/commit/4813f9b998f4b64372c389d3f930c40fad577c6c

abonas: @simon3z please see @jonmoter comments above

lib/kubeclient/config.rb
simon3z: @abonas because the configuration file and where the external files (e.g. CA) are stored, could be different paths.

simon3z:
@jonmoter please review (I added your changes about the user).

jonmoter:
Looks good!  👍

simon3z:
@abonas anything more to address here? :+1: :-1: ?

grosser:
ready for release ?

simon3z:
> ready for release ?

@grosser released in [1.1.4](https://rubygems.org/gems/kubeclient/versions/1.1.4)

PR #129, opened 2015-10-08
add response to KubeException

jonmoter:
Saves the response as an attribute of `KubeException` when it is raised.

### DIFF
--- a/lib/kubeclient/kube_exception.rb
+++ b/lib/kubeclient/kube_exception.rb
@@ -1,10 +1,11 @@
 # Kubernetes HTTP Exceptions
 class KubeException < StandardError
-  attr_reader :error_code, :message
+  attr_reader :error_code, :message, :response
 
-  def initialize(error_code, message)
+  def initialize(error_code, message, response)
     @error_code = error_code
     @message = message
+    @response = response
   end
 
   def to_s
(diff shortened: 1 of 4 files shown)


lib/kubeclient/kube_exception.rb
```diff
@@ -1,10 +1,11 @@
 # Kubernetes HTTP Exceptions
 class KubeException < StandardError
-  attr_reader :error_code, :message
+  attr_reader :error_code, :message, :response
 
-  def initialize(error_code, message)
+  def initialize(error_code, message, response = nil)
```
simon3z: Minor: I wouldn't set a default `nil` for `response`.
That means that you'd have to explicitly set it to `nil` in the tests:

    stub_request(:get, 'http://username:password@localhost:8080/api/v1/pods')
      .to_raise(KubeException.new(401, error_message, nil))

(and I wonder how come the test are still working, we should check what's the logic in `to_raise` and what's comparing)

simon3z:
Only a minor comment. Overall LGTM :+1:

simon3z:
Ah sorry I forgot to mention that technically this would also require a test (explicitly check that the `response` holds a value in the exception)

jonmoter:
Okay, I've removed the `nil` default of response, and added test logic to assert the response is there.

simon3z:
LGTM ￼ cc @abonas ready for merge. Thanks @jonmoter

PR #130, opened 2015-10-08
refactor the Net::HTTP options logic into its own helper method

jonmoter:
No logic change, merely extracting logic into a method.

This method will be used for logic to watch the logs of a pod.


simon3z:
LGTM :+1: cc @abonas ready for merge.

Thanks @jonmoter !

abonas:
actually this PR does introduce a logic change because some properties are now set only conditionally because an "if" was added in regards to ssl options.
however it looks like a good direction, so LGTM.

PR #131, opened 2015-10-08
rename open_test_json_file to open_test_file

jonmoter:
allows loading of files with different suffixes, such as .txt files


simon3z:
LGTM :+1: cc @abonas ready for merge.

PR #134, opened 2015-10-10
Add Ruby 2.2 to travis matrix

abonas:


abonas:
cc @simon3z following manageiq move to Ruby 2.2

PR #135, opened 2015-10-10
Flex 404 exception msg check to work with more rest-client versions

abonas:
Following rest-client change in https://github.com/rest-client/rest-client/pull/382/files


abonas:
@Fryguy @jimmidyson please review. thanks

PR #136, opened 2015-10-13
add functionality for getting and watching logs of pods

### LINKED ISSUE #119: Allow reading the log of a pod

jonmoter:
Resolves #119.

This is a much simpler PR than PR #121, now that a lot of the other functionality has been broken into other PRs.

/cc @simon3z

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -229,6 +229,33 @@ module Kubeclient
       end
     end
 
+    def get_pod_log(pod_name, namespace, container: nil, previous: false)
+      params = {}
+      params[:previous] = true if previous
+      params[:container] = container if container
+
+      ns = build_namespace_prefix(namespace)
+      handle_exception do
+        rest_client[ns + "pods/#{pod_name}/log"]
+          .get({ 'params' => params }.merge(@headers))
+      end
+    end
+
+    def watch_pod_log(pod_name, namespace, container: nil)
+      # Adding the "follow=true" query param tells the Kubernetes API to keep
+      # the connection open and stream updates to the log.
+      params = { follow: true }
+      params[:container] = container if container
+
+      ns = build_namespace_prefix(namespace)
+
+      uri = @api_endpoint.dup
+      uri.path += "/#{@api_version}/#{ns}pods/#{pod_name}/log"
+      uri.query = URI.encode_www_form(params)
+
+      Kubeclient::Common::WatchStream.new(uri, net_http_options(uri), format: :text)
+    end
+
     def proxy_url(kind, name, port, namespace = '')
       entity_name_plural = ClientMixin.pluralize_entity(kind.to_s)
       ns_prefix = build_namespace_prefix(namespace)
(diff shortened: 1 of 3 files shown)


jonmoter:
The Travis build was failing for some reason. I re-ran it, and it's happy now.

abonas:
@jonmoter looks great. the only really small nit I would add is to document in readme that the second param is namespace (where all examples pass "default")
ping @simon3z for a thorough review as it touches watches

jonmoter:
@abonas okay, I put a line in the Readme mentioning that the second parameter is the namespace.

Actually, I notice we're currently inconsistent with how we handle the namespace parameter. For `get_entities` it's part of the `options` hash, but for `get_entity` and `delete_entity` it's a regular parameter.

I would like to make `namespace` a required keyword parameter for the log methods, but support for that was added in 2.1, so it doesn't work on 2.0.x.  :disappointed:

lib/kubeclient/common.rb
```diff
@@ -229,6 +229,31 @@ def retrieve_all_entities(entity_types)
...
+
+      ns = build_namespace_prefix(namespace)
+      handle_exception do
+        rest_client[ns + "pods/#{pod_name}/log"]
+          .get({ 'params' => params }.merge(@headers))
+      end
+    end
+
+    def watch_pod_log(pod_name, namespace, container: nil)
+      params = { follow: true }
```
simon3z: @jonmoter is this a kubernetes api issue? Does `watch` always require `follow`? What if you're not passing it? Does it even make sense?

jonmoter: Yes, that's a Kubernetes API issue.  Notice that the endpoint for `get_pod_logs` and `watch_pod_logs` is the same. The only difference is for watch, we're adding the `follow=true` param.  If you check out the [API Docs](https://htmlpreview.github.io/?https://github.com/kubernetes/kubernetes/HEAD/docs/api-reference/v1/operations.html) and search for "/log", you'll see the parameters it accepts.

The Kubernetes team is planning to remove the /watch/* endpoints entirely, and use a query param instead.  (See [PR 8337](https://github.com/kubernetes/kubernetes/issues/8337).)  But I'll open up a separate PR to deal with that issue.

simon3z: @jonmoter it seems something that it's worth documenting with a comment in the code.

lib/kubeclient/common.rb
```diff
@@ -229,6 +229,33 @@ def retrieve_all_entities(entity_types)
...
+      handle_exception do
+        rest_client[ns + "pods/#{pod_name}/log"]
+          .get({ 'params' => params }.merge(@headers))
+      end
+    end
+
+    def watch_pod_log(pod_name, namespace, container: nil)
+      # Adding the "follow=true" query param tells the Kubernetes API to keep
+      # the connection open and stream updates to the log.
+      params = { follow: true }
```
jonmoter: Comment added

simon3z:
LGTM :+1:

PR #138, opened 2015-10-15
fix revert add component statuses

dkorn:
This is a fix for the revert "Add Component statuses" (https://github.com/abonas/kubeclient/pull/137)
re-adding the component status entity


simon3z:
@dkorn @abonas have you figured out what was failing (at first sight it seems the same patch)

dkorn:
@simon3z I missed a rebase on a PR that changed a method name

abonas:
@simon3z I still don't understand how this could be merged (and was all green) and not blocked/indicated by github that it can't be merged, especially given that the method change was merged 2-3 days before this one so there were no race conditions

Issue #140, opened 2015-11-18
Use a non-blocking HTTP library


msufa:
We have found that using `WatchStream`s inside `Celluloid` actors is a perfect fit for a background event watcher. The only problem is that `net/http` is incompatible with `Celluloid::IO` and it blocks the actor on reading the response, making it impossible to manage. Have you considered using a non-blocking HTTP library, like https://github.com/httprb/http ? We have successfully used it in a monkey patch of the `WatchStream` inside our own project which is using `kubeclient`: https://github.com/zendesk/samson/blob/msufa/cluster_pod_watcher/plugins/kubernetes/config/initializers/watchers.rb

I would appreciate your comments.

simon3z:
> I would appreciate your comments.

@msufa in the past I've considered several libraries, as `typhoeus`, `em-http-request`, `curl-multi`, etc. but they all had at least one problem that we couldn't address.

Off the top of my head requirements are:

- support basic authentication
- support specific headers
- support ssl (verify mode, client certificate, cert store etc.)
- long-standing connection (no timeout, events may come in rarely)
- read the body as a stream (don't wait for the end of the body)
- ability to cleanly close the connection (e.g. from another thread)

If you verified that all these are supported then I can start testing your changes and we'll take it from there. Thanks!

msufa:
@simon3z I did not perform any testing, but I had a closer look at the gem and found that:
 - support basic authentication
  - it supports user/password & bearer authentication
 - support specific headers
  - check
 - support ssl (verify mode, client certificate, cert store etc.)
  - it consumes `OpenSSL::SSL::SSLContext` so I assume it supports everything you can stuff in there
 - long-standing connection (no timeout, events may come in rarely)
  - it does not time out by default
 - read the body as a stream (don't wait for the end of the body)
  - check, I'm actually using this in my monkey patch
 - ability to cleanly close the connection (e.g. from another thread)
  - I'm using it through `Celluloid` actors and haven't bumped into problems closing connections yet

simon3z:
> @simon3z I did not perform any testing, but I had a closer look at the gem and found that:

Send a PR and I'll test it to check if there was anything else that I forgot. Thanks.

msufa:
It took me a while but the PR to resolve this is finally published. Please review #142 @simon3z

henders:
Nice! +1

msufa:
The relevant PR has been merged so I consider the issue closed. Thanks @simon3z & @abonas for handling that!

PR #142, opened 2015-12-02
Use the non-blocking IO enabled http.rb instead of net/http in watchers

msufa:
When using kubeclient in an actor based concurrent environment, such as Celuloid, we need a non blocking HTTP library to handle watching Kubernetes events. http.rb can be used with Celuloid::IO sockets, so that Celluloid actors remain responsive when awaiting cluster events.

I've tested with basic and ssl authentication options, both with and without the usage of Celluloid::IO. I've not tested thoroughly using bearer token, I only verified that the headers are set when making a request.

This resolves issue #140

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -24,6 +24,10 @@ module Kubeclient
         password: nil,
         bearer_token: nil,
         bearer_token_file: nil
+      },
+      socket_options: {
+        socket_class: nil,
+        ssl_socket_class: nil
       }
     )
       validate_auth_options(auth_options)
@@ -33,12 +37,13 @@ module Kubeclient
       @headers = {}
       @ssl_options = ssl_options
       @auth_options = auth_options
+      @socket_options = socket_options
 
       if auth_options[:bearer_token]
-        @headers[:Authorization] = "Bearer #{@auth_options[:bearer_token]}"
+        bearer_token(@auth_options[:bearer_token])
       elsif auth_options[:bearer_token_file]
         validate_bearer_token_file
-        @headers[:Authorization] = "Bearer #{File.read(@auth_options[:bearer_token_file])}"
+        bearer_token(File.read(@auth_options[:bearer_token_file]))
       end
     end
 
@@ -150,7 +155,7 @@ module Kubeclient
         uri.query = URI.encode_www_form(params.map { |k, v| [k.to_s.camelize(:lower), v] })
       end
 
-      Kubeclient::Common::WatchStream.new(uri, net_http_options(uri))
+      Kubeclient::Common::WatchStream.new(uri, http_options(uri))
     end
 
     # Accepts the following string options:
@@ -272,7 +277,7 @@ module Kubeclient
       uri.path += "/#{@api_version}/#{ns}pods/#{pod_name}/log"
       uri.query = URI.encode_www_form(params)
 
-      Kubeclient::Common::WatchStream.new(uri, net_http_options(uri), format: :text)
+      Kubeclient::Common::WatchStream.new(uri, http_options(uri), format: :text)
     end
 
     def proxy_url(kind, name, port, namespace = '')
@@ -329,7 +334,7 @@ module Kubeclient
       fail ArgumentError, msg unless File.readable?(@auth_options[:bearer_token_file])
     end
 
-    def net_http_options(uri)
+    def http_options(uri)
       options = {
         basic_auth_user: @auth_options[:username],
         basic_auth_password: @auth_options[:password],
@@ -337,19 +342,18 @@ module Kubeclient
       }
 
       if uri.scheme == 'https'
-        options.merge!(
-          use_ssl: true,
+        options[:ssl] = {
           ca_file: @ssl_options[:ca_file],
           cert: @ssl_options[:client_cert],
           cert_store: @ssl_options[:cert_store],
           key: @ssl_options[:client_key],
-          # ruby Net::HTTP uses verify_mode instead of verify_ssl
-          # http://ruby-doc.org/stdlib-1.9.3/libdoc/net/http/rdoc/Net/HTTP.html
+          # ruby HTTP uses verify_mode instead of verify_ssl
+          # http://ruby-doc.org/stdlib-1.9.3/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html
           verify_mode: @ssl_options[:verify_ssl]
-        )
+        }
       end
 
-      options
+      options.merge(@socket_options)
     end
   end
 end
--- a/lib/kubeclient/watch_stream.rb
+++ b/lib/kubeclient/watch_stream.rb
@@ -1,54 +1,65 @@
 require 'json'
-require 'net/http'
+require 'http'
 module Kubeclient
   module Common
     # HTTP Stream used to watch changes on entities
     class WatchStream
       def initialize(uri, http_options, format: :json)
         @uri = uri
-        @http = nil
-        @http_options = http_options.merge(read_timeout: nil)
+        @http_client = nil
+        @http_options = http_options
         @format = format
       end
 
-      def each
+      def each(&block)
         @finished = false
-        @http = Net::HTTP.start(@uri.host, @uri.port, @http_options)
-
-        buffer = ''
-        request = generate_request
-
-        @http.request(request) do |response|
-          unless response.is_a? Net::HTTPSuccess
-            fail KubeException.new(response.code, response.message, response)
-          end
-          response.read_body do |chunk|
-            buffer << chunk
-            while (line = buffer.slice!(/.+\n/))
-              yield @format == :json ? WatchNotice.new(JSON.parse(line)) : line.chomp
-            end
-          end
+        @http_client = build_client
+        response = @http_client.request(:get, @uri, build_client_options)
+        unless response.code < 300
+          fail KubeException.new(response.code, response.reason, response)
         end
-      rescue Errno::EBADF
+        read_stream(response.body, &block)
+      rescue IOError
         raise unless @finished
-      end
-
-      def generate_request
-        request = Net::HTTP::Get.new(@uri)
-        if @http_options[:basic_auth_user] && @http_options[:basic_auth_password]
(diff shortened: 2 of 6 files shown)


msufa:
Not sure how to handle the `Avoid parameter lists longer than 5 parameters.` rubocop restriction. Please let me know if you can live with it or how would you rather expose the socket options to the user.

simon3z:
Thanks @msufa! Don't mind the travis error for now. I expect this to be a PR with a long life-cycle as I want to be sure that everything works as expected. I'll start to test it now but I want to run with it for a week or more to check it thoroughly.

simon3z:
@msufa I see that you've done some generic refactoring (thanks!) to facilitate the introduction of non-blocking IO but that is good regardless, for example creating `read_stream`, etc.

If you split that out into its own patch (you can use this PR anyway) it would facilitate my work of reviewing (keeping refactoring and features separated).
Thanks!

msufa:
@simon3z, I only extracted that part because I got a warning from Code Climate that the host method is too complex. I can stuff that code back in if it makes the PR clearer.

simon3z:
> @simon3z, I only extracted that part because I got a warning from Code Climate that the host method is too complex. I can stuff that code back in if it makes the PR clearer.

@msufa no, on the contrary I am all in favor of that but in the scope of this PR it makes it harder to review for relevant changes. Just split it to another patch and we're good.

simon3z:
This is working for my use-cases, @abonas can you review?

abonas:
Didn't see that this comment was addressed: https://github.com/abonas/kubeclient/pull/142#issuecomment-161709995
Did I miss it?

lib/kubeclient/common.rb
```diff
@@ -24,6 +24,10 @@ def initialize_client(
         password: nil,
         bearer_token: nil,
         bearer_token_file: nil
+      },
+      socket_options: {
```
abonas: I am probably missing something here, but why is there a need to do the socket_options hash initialization both here and in kubeclient.rb line 55? the code looks the same and kubeclient.rb calls common.rb anyway

msufa: I agree @abonas, I was just trying to be consistent with the way `ssl_options` & `auth_options` are handled. I can clean them all up either here or in a different PR, whichever you fancy.

abonas: > I agree @abonas, I was just trying to be consistent with the way ssl_options & auth_options are handled. I can clean them all up either here or in a different PR, whichever you fancy.

imo no need to handle it in this PR, a separate one can be totally fine. @simon3z ?
And @msufa , thanks for your work!

msufa:
> Didn't see that this comment was addressed: #142 (comment)
Did I miss it?

@abonas, indeed this comment was addressed, please have a look here https://github.com/zendesk/kubeclient/commit/ac7ac64c53710e9367ee9e6f09305d5e62be2c5c
I'll wait until this PR is merged and revert that commit to bring back the refactoring if that's ok.

lib/kubeclient/common.rb
abonas: I can't see the build details for some reason, but Travis does fail...

abonas:
> @abonas, indeed this comment was addressed, please have a look here zendesk@ac7ac64
> I'll wait until this PR is merged and revert that commit to bring back the refactoring if that's ok.

@simon3z was the one doing the more thorough review of this PR given that he implemented the watches area, I'll let him decide on the approach here.

lib/kubeclient/common.rb
msufa: Travis is failing on a Rubocop check, I mentioned it here https://github.com/abonas/kubeclient/pull/142#issuecomment-161314282

simon3z: > imo no need to handle it in this PR, a separate one can be totally fine. @simon3z ?

:+1: (separate PR is better)

simon3z:
@msufa please rebase and squash `move parsing of the HTTP response back into the #each method` in `update readme to include info about non-blocking socket support`

lib/kubeclient/common.rb
```diff
@@ -337,19 +342,18 @@ def net_http_options(uri)
       }
 
       if uri.scheme == 'https'
-        options.merge!(
-          use_ssl: true,
+        options[:ssl] = {
```
simon3z: @msufa why is this change required?

msufa: http.rb takes SSL options from an `ssl` sub hash under main options, see https://github.com/httprb/http/blob/1836ae8d7986d4ce39ab417bdfd20b9098800edc/lib/http/options.rb#L47

Furthermore, `use_ssl` is not required for http.rb, it's a net/http concept, see http://ruby-doc.org/stdlib-2.2.4/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-HTTPS

msufa:
@simon3z PR rebased, commits squashed.

lib/kubeclient/watch_stream.rb
```diff
@@ -1,54 +1,63 @@
...
 module Kubeclient
   module Common
     # HTTP Stream used to watch changes on entities
     class WatchStream
       def initialize(uri, http_options, format: :json)
         @uri = uri
-        @http = nil
-        @http_options = http_options.merge(read_timeout: nil)
+        @http_client = nil
+        @http_options = http_options
```
simon3z: @msufa you removed the `read_timeout` but apparently you haven't introduced anything similar for the new `http` gem, don't we need something like:

http://www.rubydoc.info/gems/http/0.9.8/HTTP/Timeout/Global

lib/kubeclient/watch_stream.rb
```diff
@@ -1,54 +1,63 @@
...
 
       def each
         @finished = false
-        @http = Net::HTTP.start(@uri.host, @uri.port, @http_options)
 
-        buffer = ''
-        request = generate_request
+        @http_client = build_client
+        response = @http_client.request(:get, @uri, build_client_options)
+        unless response.code < 300
```
simon3z: @msufa how is this related to "Use non-blocking IO"? It's a good improvement but it should be evaluated separately, so if it's not part of this PR please remove it.

simon3z:
@msufa I think that to pass rubocop you have to add in `.rubocop.yml`:

    Metrics/ParameterLists:
      Max: 8

`8` seems a good compromise to me. Feel free to suggest another value, cc @abonas 

Otherwise @msufa you'll have to comply to current metric which is `5` and refactor `initialize_client` (it would be too much confusing in this PR though).

abonas:
> @msufa I think that to pass rubocop you have to add in .rubocop.yml:
> 
> Metrics/ParameterLists:
>   Max: 8
> 8 seems a good compromise to me. Feel free to suggest another value, cc @abonas
> 
> Otherwise @msufa you'll have to comply to current metric which is 5 and refactor initialize_client (it would be too much confusing in this PR though).

no objections to change rubocop.yml threshold

lib/kubeclient/watch_stream.rb
msufa: @simon3z, `HTTP` does not time out by default so no explicit setup is needed:
 - https://github.com/httprb/http#timeouts
 - https://github.com/httprb/http/blob/v0.9.8/lib/http/options.rb#L10

lib/kubeclient/watch_stream.rb
msufa: @simon3z, unfortunately `HTTP` does not have a fancy `Success` wrapper class and deals only with raw integer codes. In case of `net/http` the trick is to have all the 2xx codes inherit from `Net::HTTPSuccess`, see [here](https://github.com/ruby/ruby/blob/trunk/lib/net/http/responses.rb#L36). So in our case the only difference is that we treat 1xx responses as success as well. I can fix this if you think it makes a difference.

msufa:
I've relaxed the rubocop parameter list restriction to 8.
I've also fixed the `http.rb` dependency to version 0.9.8 for now, as the Celluloid::IO support was temporarily removed from 1.0, see [here](https://github.com/httprb/http/wiki/Parallel-requests-with-Celluloid::IO) for details.
/cc @simon3z @abonas

abonas:
thanks @msufa , @simon3z please review/merge

msufa:
Bump...

abonas:
@msufa I pinged @simon3z and asked him to confirm/merge the PR - given that he is more familiar with the modified area and led the review. sorry for the delays.

abonas:
@msufa if you could clarify the following to me - as that wiki states, Celluloid is removed from 1.0 of http (it's not clear from the link why it was removed), and setting hardcoded http gem version dependency in kubeclient to have Celluloid support can create issues for some kubeclient users who require other versions of the http gem at the moment, and can prevent from upgrading the gem version to enjoy other bug fixes, etc. Your thoughts on the topic?

msufa:
@abonas, please have a look here for more details on `Celluloid::IO` support in `http.rb`: https://github.com/httprb/http/issues/272
Although there are many other benefits from using `http.rb`, the most significant use case for us is to utilise the non-blocking IO in watchers, implemented as Celluloid Actors, which constantly observe the state of the cluster. We believe that this is something that other users of `kubeclient` could benefit from.
I don't see any crucial changes in `http.rb` when comparing v0.9.8 with v1.0.0. I've also asked for some more details about the future of Celluloid support. Ultimately, upgrading the dependency to a version other than v0.9.8 should not break any code, as the extra socket parameters should just be ignored by the http library. This would be totally transparent for `kubeclient` users which don't rely on Celluloid support.

abonas:
> @abonas, please have a look here for more details on Celluloid::IO support in http.rb: httprb/http#272
> Although there are many other benefits from using http.rb, the most significant use case for us is to utilise the non-blocking IO in watchers, implemented as Celluloid Actors, which constantly observe the state of the cluster. We believe that this is something that other users of kubeclient could benefit from.
> I don't see any crucial changes in http.rb when comparing v0.9.8 with v1.0.0. I've also asked for some more details about the future of Celluloid support. Ultimately, upgrading the dependency to a version other than v0.9.8 should not break any code, as the extra socket parameters should just be ignored by the http library. This would be totally transparent for kubeclient users which don't rely on Celluloid support.

ack. thanks!

PR #143, opened 2015-12-11
Add in the Deployment resource type even though it's still currently in beta

henders:
The Deployment API is a great improvement to the current ReplicationController mechanism of deploying your app. (https://github.com/kubernetes/kubernetes/blob/master/docs/user-guide/deployments.md).

This change just adds in the new class and methods to use for interacting with the Deployment API. I put an example of how to use it in the README.

cc @msufa @jonmoter

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@
 [![Dependency Status](https://gemnasium.com/abonas/kubeclient.svg)](https://gemnasium.com/abonas/kubeclient)
 
 A Ruby client for Kubernetes REST api.
-The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas, limit ranges, endpoints, persistent volumes, persistent volume claims, component statuses and service accounts.
-The client currently supports Kubernetes REST api version v1.
+The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, deployments, namespaces, resource quotas, limit ranges, endpoints, persistent volumes, persistent volume claims, component statuses and service accounts.
+The client currently supports Kubernetes REST api version v1 (and limited selection of extension APIs).
 
 ## Installation
 
@@ -36,6 +36,11 @@ Or without specifying version (it will be set by default to "v1")
 
 ```ruby
 client = Kubeclient::Client.new 'http://localhost:8080/api/'
+```
+
+Or to access the extension api groups
+```ruby
+client = Kubeclient::Client.new 'http://localhost:8080/apis', 'extensions/v1beta1'
 ```
 
 Another option is to initialize the client with URI object:
@@ -131,7 +136,7 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , 'v1',
 ## Examples:
 
 #### Get all instances of a specific entity type
-Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`, `get_resource_quotas`, `get_limit_ranges`, `get_persistent_volumes`, `get_persistent_volume_claims`, `get_component_statuses`, `get_service_accounts`
+Such as: `get_pods`, `get_secrets`, `get_services`, `get_nodes`, `get_replication_controllers`, `get_deployments`, `get_resource_quotas`, `get_limit_ranges`, `get_persistent_volumes`, `get_persistent_volume_claims`, `get_component_statuses`, `get_service_accounts`
 
 ```ruby
 pods = client.get_pods
@@ -155,7 +160,7 @@ pods = client.get_pods(label_selector: 'name=redis-master,app=redis')
 ```
 
 #### Get a specific instance of an entity (by name)
-Such as: `get_service "service name"` , `get_pod "pod name"` , `get_replication_controller "rc name"`, `get_secret "secret name"`, `get_resource_quota "resource quota name"`, `get_limit_range "limit range name"` , `get_persistent_volume "persistent volume name"` , `get_persistent_volume_claim "persistent volume claim name"`, `get_component_status "component name"`, `get_service_account "service account name"`
+Such as: `get_service "service name"` , `get_pod "pod name"` , `get_replication_controller "rc name"`, `get_deployment "deployment name"`, `get_secret "secret name"`, `get_resource_quota "resource quota name"`, `get_limit_range "limit range name"` , `get_persistent_volume "persistent volume name"` , `get_persistent_volume_claim "persistent volume claim name"`, `get_component_status "component name"`, `get_service_account "service account name"`
 
 The GET request should include the namespace name, except for nodes and namespaces entities.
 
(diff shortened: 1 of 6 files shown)


test/test_deployment.rb
```diff
@@ -0,0 +1,25 @@
...
+
+# Replication Controller entity tests
+class TestDeployment < MiniTest::Test
+  def test_get_from_json_v1
+    stub_request(:get, %r{/deployment})
+      .to_return(body: open_test_file('deployment.json'),
+                 status: 200)
+
+    Kubeclient::Deployment.new
+    client = Kubeclient::Client.new 'http://localhost:8080/apis', 'extensions/v1beta1'
```
simon3z: @henders at first glance I'd say that we would like to formalize better how to reach extensions and their api versioning (according to the new api groups in kubernetes). This line works but has the feeling of a temporary workaround.

README.md
```diff
@@ -204,6 +204,41 @@ service.metadata.labels.role = 'slave'
 client.create_service service`
 ```
 
+Below example is for the extension API
```
simon3z: @henders IIRC there are other places in the README that should be updated to mention deployments.

test/test_deployment.rb
henders: @simon3z , yup, I didn't want to mess with the current code too much, and I figured the test should change anyway when 'Deployments' gets merged to the official api, but I'm up for suggestions.

README.md
henders: Added some more mentions. A lot of the places seem like they just list examples of what you can do, and don't necessarily need to mention all possible operations.

henders:
I've just realized the all_entities() method is broken as you can't enumerate all the v1 objects with the extension objects with the current method of configuring the Kubeclient::Client.

This may need some more discussion, but the only way I see around this is to build awareness of which objects live in which api group.

E.g.
```ruby
V1 = %w(Pod Service ReplicationController Node Event Endpoint
                      Namespace Secret ResourceQuota LimitRange PersistentVolume
                      PersistentVolumeClaim ComponentStatus ServiceAccount)
EXTENSIONS_V1BETA1 = %w(Deployment)
VERSIONS = %i(V1 EXTENSIONS_V1BETA1)
```
Then use the name of the containing Array to dictate the URL sent to K8s, e.g.
```ruby
api_version = VERSIONS.find {|ver| ver.any? {|obj| obj == resource_name(entity_type) } }
api_url = api_version.sub('_', '/').lowercase
```

henders:
Or maybe create a different client type to match what you want to do?
```ruby
class Client
  ENTITY_TYPES = %w(Pod Service .....)
end
class ClientExtensionV1beta1
  ENTITY_TYPES = %w(Deployment ....)
end
...
...
```

simon3z:
@henders I sketched a quick description of the feature in #144

abonas:
Hi @henders , thanks for bringing up this issue!
I definitely think we should have a discussion on handling different api versions and extensions, and I'll add my thoughts on the topic shortly in https://github.com/abonas/kubeclient/issues/144.
Regardless, I'd like to understand better the functional and technical perspective of the Deployment object from @smarterclayton, given that such object already exists in OpenShift:
https://docs.openshift.org/latest/architecture/core_concepts/deployments.html#deployments-and-deployment-configurations.
(I'd like to understand whether looking forward it will be removed from OpenShift api and move to Kubernetes core API, so it will be available on /api and not oapi, so won't need to be handled in  https://github.com/abonas/openshift_client

test/test_deployment.rb
abonas: > @henders at first glance I'd say that we would like to formalize better how to reach extensions and their api versioning (according to the new api groups in kubernetes). This line works but has the feeling of a temporary workaround.

I agree, we need a better solution to work with multiple versions and experimental features.

smarterclayton:
DeploymentConfigs will be around for a while in OpenShift.  The current plan is for all Kube deployments to show up under the deployment config api.  The extra features OpenShift exposes will only be visible under the deploymentconfig endpoint, and there will be some limits on how k8s deployments work in the system.  To get feature parity with OpenShift is probably around the 1.3 or 1.4 kube timeframe (end of this next year?) so I wouldn't expect it to go away any time soon.

nyxcharon:
Any updates on  getting deployments added to kubeclient?

simon3z:
> Any updates on getting deployments added to kubeclient?

@moolitayer I have the feeling that your #185 should cover this as well right?

moolitayer:
> @moolitayer I have the feeling that your #185 should cover this as well right?

Yes, please review

moolitayer:
@simon3z @henders This should be closed through #185

onbjerg:
I can't get #185 work with extensions/v1beta1. Something is really messed up.

```
~|⇒ irb -rkubeclient
irb(main):001:0> client = Kubeclient::Client.new 'http://localhost:8001/apis/extensions', 'v1beta1'
=> #<Kubeclient::Client:0x007fd59c867878 @api_endpoint=#<URI::HTTP http://localhost:8001/apis/extensions>, @api_version="v1beta1", @headers={}, @ssl_options={:client_cert=>nil, :client_key=>nil, :ca_file=>nil, :cert_store=>nil, :verify_ssl=>1}, @auth_options={:username=>nil, :password=>nil, :bearer_token=>nil, :bearer_token_file=>nil}, @socket_options={:socket_class=>nil, :ssl_socket_class=>nil}>
irb(main):002:0> client.get_deployments
NoMethodError: undefined method `get_deployments' for #<Kubeclient::Client:0x007fd59c867878>
Did you mean?  get_events
	from (irb):2
	from /usr/local/bin/irb:11:in `<main>'
irb(main):003:0> client.discover
NoMethodError: undefined method `discover' for #<Kubeclient::Client:0x007fd59c867878>
	from (irb):3
	from /usr/local/bin/irb:11:in `<main>'
irb(main):004:0> 
```

It still works fine with api/v1 though

```
~|⇒ irb -rkubeclient
irb(main):001:0> client = Kubeclient::Client.new 'http://localhost:8001/api', 'v1'
=> #<Kubeclient::Client:0x007fa84d1d0bd0 @api_endpoint=#<URI::HTTP http://localhost:8001/api>, @api_version="v1", @headers={}, @ssl_options={:client_cert=>nil, :client_key=>nil, :ca_file=>nil, :cert_store=>nil, :verify_ssl=>1}, @auth_options={:username=>nil, :password=>nil, :bearer_token=>nil, :bearer_token_file=>nil}, @socket_options={:socket_class=>nil, :ssl_socket_class=>nil}>
irb(main):002:0> client.get_pods
=> [...]
```

moolitayer:
OH MY :grimacing: 
Is it working now?

simon3z:
> EDIT D'oh, never mind. Wrong gem version

@onbjerg so this is fixed? If so you can close this issue.

onbjerg:
@moolitayer Yes. It works perfectly fine, I've tried getting all deployments, individual ones and patching them.

@simon3z It is fixed in #185. I can't close the issue.

henders:
Great! 🎉

garo:
I'm sorry, but I'm not getting how this is now supposed to be working? Do I need to create separated clients for both /api/v1 and /apis/extensions/v1beta1?

PR #150, opened 2016-01-06
Fix for nil ns_prefix caused by upgrade to recursive-open-struct 1.0

bkoski:
Not sure if this is caused by ruby 2.3.0, recursive_open_struct 1.0.0, or some combination of the two but – when trying to create services with kubeclient 1.0.0 on ruby 2.3.0, was getting 

```
405 The server does not allow this method on the requested resource
```

Did some debugging and found that `ns_prefix` in `create_resource` was nil, because keys under `entity_config.metadata['table']` were strings rather than symbols.   So

```
entity_config.metadata['table'][:namespace]
```

is nil, but these expressions work:

```
entity_config.metadata['table']['namespace']
entity_config.metadata.namesapce
```

Not sure why this reference was going through `.metadata['table']`, but as far as I understand ROS 1.0.0 you should always be able to hit `entity_config.metadata.namespace`?

Changed this reference in this PR.  In addition, removed the workaround for aetherknight/recursive-open-struct#9 since 1.0.0 solves for this.

Tests pass; this also fixes the 405 error.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -207,17 +207,17 @@ module Kubeclient
     end
 
     def create_entity(entity_type, entity_config, klass)
-      # to_hash should be called because of issue #9 in recursive open
-      # struct
+      # Duplicate the entity_config to a hash so that when we assign
+      # kind and apiVersion, this does not mutate original entity_config obj.
       hash = entity_config.to_hash
 
-      ns_prefix = build_namespace_prefix(entity_config.metadata['table'][:namespace])
+      ns_prefix = build_namespace_prefix(hash[:metadata][:namespace])
 
       # TODO: temporary solution to add "kind" and apiVersion to request
       # until this issue is solved
       # https://github.com/GoogleCloudPlatform/kubernetes/issues/6439
-      hash['kind'] = entity_type
-      hash['apiVersion'] = @api_version
+      hash[:kind]       = entity_type
+      hash[:apiVersion] = @api_version
       response = handle_exception do
         rest_client[ns_prefix + resource_name(entity_type)]
         .post(hash.to_json, @headers)
@@ -227,14 +227,11 @@ module Kubeclient
     end
 
     def update_entity(entity_type, entity_config)
-      name = entity_config.metadata.name
-      # to_hash should be called because of issue #9 in recursive open
-      # struct
-      hash = entity_config.to_hash
-      ns_prefix = build_namespace_prefix(entity_config.metadata['table'][:namespace])
+      name      = entity_config[:metadata][:name]
+      ns_prefix = build_namespace_prefix(entity_config[:metadata][:namespace])
       handle_exception do
         rest_client[ns_prefix + resource_name(entity_type) + "/#{name}"]
-          .put(hash.to_json, @headers)
+          .put(entity_config.to_h.to_json, @headers)
       end
     end
 
--- a/test/test_service.rb
+++ b/test/test_service.rb
@@ -23,9 +23,9 @@ class TestService < MiniTest::Test
     assert_equal our_service.metadata.labels.name,
                  hash[:metadata][:labels][:name]
 
-    stub_request(:post, %r{namespaces/staging/services})
-      .to_return(body: open_test_file('created_service.json'),
-                 status: 201)
+    expected_url = 'http://localhost:8080/api/v1/namespaces/staging/services'
+    stub_request(:post, expected_url)
+      .to_return(body: open_test_file('created_service.json'), status: 201)
 
     client = Kubeclient::Client.new 'http://localhost:8080/api/'
     created = client.create_service our_service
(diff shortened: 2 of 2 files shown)


lib/kubeclient/common.rb
```diff
@@ -223,13 +219,10 @@ def create_entity(entity_type, entity_config, klass)
 
     def update_entity(entity_type, entity_config)
       name = entity_config.metadata.name
-      # to_hash should be called because of issue #9 in recursive open
-      # struct
-      hash = entity_config.to_hash
-      ns_prefix = build_namespace_prefix(entity_config.metadata['table'][:namespace])
+      ns_prefix = build_namespace_prefix(entity_config.metadata.namespace)
```
simon3z: @bkoski please don't remove the access to `namespace` using `['table'][:namespace]`, it's going to create a regression in applications where `.namespace` is already overridden.

Was that required for 2.3.0?

simon3z: @bkoski let's use `entity_config.metadata['table']['namespace']` if that works.

simon3z:
@abonas @bkoski why the tests passed in the first place? If we don't have any test covering this part we should probably add them (with this PR possibly).

lib/kubeclient/common.rb
```diff
@@ -202,20 +202,16 @@ def delete_entity(entity_type, name, namespace = nil)
...
 
       # TODO: temporary solution to add "kind" and apiVersion to request
       # until this issue is solved
       # https://github.com/GoogleCloudPlatform/kubernetes/issues/6439
-      hash['kind'] = entity_type
-      hash['apiVersion'] = @api_version
+      entity_config['kind'] = entity_type
+      entity_config['apiVersion'] = @api_version
```
simon3z: @bkoski you haven't mentioned this change in behavior: you're now changing the original object passed in (`entity_config`).

abonas: @bkoski I also wonder why is this change needed for this PR, and why the original object is changed (entity_config)

abonas:
> @abonas @bkoski why the tests passed in the first place? If we don't have any test covering this part we should probably add them (with this PR possibly).

I'm puzzled about the tests passing as well

abonas:
@bkoski could you perhaps share a code example of how you were constructing a service?

bkoski:
Did some digging, the error shows up when you assign `.metadata` using a Hash with string keys (rather than symbols, or directly assigning the properties through deep accessors).

In my particular case, I was setting up the service instance after parsing a JSON payload:

```
payload = JSON.parse(params[:service])
service = Kubeclient::Service.new
service.metadata = service_data['metadata']
service.spec     = service_data['spec']
```

It seems the error was caused not by Ruby 2.3, but instead by the ROS 1.0 upgrade.  In ROS 1.0, [all keys are cast to symbols internally](https://github.com/aetherknight/recursive-open-struct/commit/c5c00f5bf0f690d6f8b236651151b3335005cc71#diff-04c6e90faac2675aa89e2176d2eec7d8R25).

Updated the PR to:
  * include a test that exposes the error
  * go back to the original setup, where `entity_config` is converted to a hash to avoid mutating the original param
  * access the namespace as `hash[:metadata][:namespace]`, which appears to work no matter how you initialize the service

abonas:
@bkoski please squash the commits

bkoski:
Just squashed, and added a couple more tests to exercise creation and updates w/ string/symbol keys.

lib/kubeclient/common.rb
```diff
@@ -207,34 +207,31 @@ def delete_entity(entity_type, name, namespace = nil)
...
       # until this issue is solved
       # https://github.com/GoogleCloudPlatform/kubernetes/issues/6439
-      hash['kind'] = entity_type
-      hash['apiVersion'] = @api_version
+      hash[:kind]       = entity_type
+      hash[:apiVersion] = @api_version
       response = handle_exception do
         rest_client[ns_prefix + resource_name(entity_type)]
-        .post(hash.to_json, @headers)
+        .post(entity_config.to_json, @headers)
```
abonas: shouldn't it be hash that is sent to the server? 
the entity_config doesn't contain kind and apiVersion set on lines 219 and 220

bkoski:
Good call, fixed this and added some some tests that:
  - check payload sent to kubernetes api to make sure it's valid JSON with complete data
  - assert that original entity_confg is not modified

lib/kubeclient/common.rb
```diff
@@ -227,14 +227,11 @@ def create_entity(entity_type, entity_config, klass)
     end
 
     def update_entity(entity_type, entity_config)
-      name = entity_config.metadata.name
-      # to_hash should be called because of issue #9 in recursive open
-      # struct
-      hash = entity_config.to_hash
-      ns_prefix = build_namespace_prefix(entity_config.metadata['table'][:namespace])
+      name      = entity_config[:metadata][:name]
```
abonas: did the access with metadata.name stopped working?

test/test_service.rb
```diff
@@ -33,6 +33,81 @@ def test_construct_our_own_service
...
+
+    # Check that original entity_config is not modified by kind/apiVersion patches:
+    assert_equal(our_service.kind, nil)
+
+    assert_requested(:post, expected_url, times: 1) do |req|
+      data = JSON.parse(req.body)
+      data['kind'] == 'Service' &&
+        data['apiVersion'] == 'v1' &&
+        data['metadata']['name'] == 'guestbook' &&
+        data['metadata']['namespace'] == 'staging'
```
abonas: :+1:

abonas:
can you please change the PR title to a more relevant one? the mention of ruby 2.3 is misleading as the issue is eventually not related to that.

lib/kubeclient/common.rb
bkoski: `metadata.name` was working, but my thought was that all references into `entity_config` should be consistent.  In the original version of the PR the reference to ns_prefix was
```
build_namespace_prefix(entity_config.metadata.namespace)
```
but seemed like [the preference](https://github.com/abonas/kubeclient/pull/150#discussion_r49056553) was towards using `[]` rather than method references.

bkoski:
Just updated title.

lib/kubeclient/common.rb
abonas: > but seemed like the preference was towards using [] rather than method references

yes, because we are aware with override specific issues of the 'namespace' part :)
but it's ok.

test/test_service.rb
```diff
@@ -111,26 +186,49 @@ def test_get_service
...
+    service.metadata = {}
+    service.metadata.name      = name
+    service.metadata.namespace = 'development'
+
+    expected_url = "http://localhost:8080/api/v1/namespaces/development/services/#{name}"
+    stub_request(:put, expected_url)
+      .to_return(body: open_test_file('service_update.json'), status: 201)
+
+    client = Kubeclient::Client.new 'http://localhost:8080/api/', 'v1'
+    client.update_entity entity, service
```
abonas: preferred way to do it is: client.update_service serviceObj

PR #152, opened 2016-02-23
Add content-type header = application/json for create and update actions

abonas:
Following PR  Split codec from scheme #17922 on k8s server.
https://github.com/kubernetes/kubernetes/pull/17922

Without sending content-type, the requests fail.
cc @simon3z

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -218,6 +218,7 @@ module Kubeclient
       # https://github.com/GoogleCloudPlatform/kubernetes/issues/6439
       hash[:kind]       = entity_type
       hash[:apiVersion] = @api_version
+      @headers['Content-Type'] = 'application/json'
       response = handle_exception do
         rest_client[ns_prefix + resource_name(entity_type)]
         .post(hash.to_json, @headers)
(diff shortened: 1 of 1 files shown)


simon3z:
@abonas I think that `post` and `put` support `:content_type` in the request. E.g.

    rest_client[ns_prefix + resource_name(entity_type)].post(hash.to_json, @headers, :content_type => 'application/json')

It looks just slightly better (because closer to the `to_json` transformation), but I don't know if it works.
(Not sure if it's worth investigating, up to you)

Should we do something for what format we expect in the response? E.g. passing the `Accept: application/json` header to be sure to receive json?

abonas:
> @abonas I think that post and put support :content_type in the request. E.g.
> 
> rest_client[ns_prefix + resource_name(entity_type)].post(hash.to_json, @headers, :content_type => 'application/json')
> It looks just slightly better (because closer to the to_json transformation), but I don't know if it works.
> (Not sure if it's worth investigating, up to you)

I thought about it too, saw some examples that use it. What bothered me is that eventually content type is a header, and there is already a headers object passed. Differentiating it can confuse the reader since it will create 2 different ways of passing headers. 
Thoughts?

> Should we do something for what format we expect in the response? E.g. passing the Accept: application/json header to be sure to receive json?

It does work without passing Accept, and I remember there were some exceptions that were sent in response as text instead of json from k8s server ( a bug, but still...) , so passing Accept can actually prevent the client from getting an exception from the server.
So I tend to say not to send Accept at the moment.

simon3z:
OK :+1:

PR #153, opened 2016-02-24
Fix #81 kind of an endpoint in creation should be Endpoints

### LINKED ISSUE #81: Endpoints entity_type is not plural

abonas:
cc @simon3z

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -216,7 +216,13 @@ module Kubeclient
       # TODO: temporary solution to add "kind" and apiVersion to request
       # until this issue is solved
       # https://github.com/GoogleCloudPlatform/kubernetes/issues/6439
-      hash[:kind]       = entity_type
+      # TODO: #2 solution for
+      # https://github.com/kubernetes/kubernetes/issues/8115
+      if entity_type.eql? 'Endpoint'
+        hash[:kind] = resource_name(entity_type).capitalize
+      else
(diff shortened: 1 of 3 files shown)


lib/kubeclient/common.rb
```diff
@@ -216,7 +216,11 @@ def create_entity(entity_type, entity_config, klass)
       # TODO: temporary solution to add "kind" and apiVersion to request
       # until this issue is solved
       # https://github.com/GoogleCloudPlatform/kubernetes/issues/6439
-      hash[:kind]       = entity_type
+      if entity_type.eql? 'Endpoint'
```
simon3z: @abonas do we need also another comment with the reference to this specific issue? (Remind me... is the issue on their side still open?)

simon3z:
LGTM :+1:

lib/kubeclient/common.rb
abonas: > @abonas do we need also another comment with the reference to this specific issue? (Remind me... is the issue on their side still open?)

I'll add a comment, it doesn't cost money :)
yes, the issue on server side is still open.

PR #154, opened 2016-03-03
Remove json is part of the ruby stdlib now

cored:


abonas:
@cored
please add a reference/comment in the commit message - since which ruby version json was added as part of ruby.
 thank you! 
(also, I'd be really interested to hear if you are willing to elaborate how/where you are using kubeclient, just to learn more about the gem usage)

cored:
@abonas I was considering using the gem but still not sure. JSON is part of the stdlib since 1.9. Take a look at this blog post. 
http://www.mikeperham.com/2016/02/09/kill-your-dependencies/

abonas:
> @abonas I was considering using the gem but still not sure. JSON is part of the stdlib since 1.9. Take a look at this blog post. 
> http://www.mikeperham.com/2016/02/09/kill-your-dependencies/

very nice article, thank you

PR #156, opened 2016-03-07
Add content-type header = application/json for create and update actions

simon3z:
Following PR  Split codec from scheme #17922 on k8s server.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -194,6 +194,7 @@ module Kubeclient
       # https://github.com/GoogleCloudPlatform/kubernetes/issues/6439
       hash['kind'] = entity_type
       hash['apiVersion'] = @api_version
+      @headers['Content-Type'] = 'application/json'
       response = handle_exception do
         rest_client[ns_prefix + resource_name(entity_type)]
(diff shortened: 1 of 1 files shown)


simon3z:
@abonas this is a serious issue (backported from master) that we want to fix in v0.8.z releasing a v0.8.1.

abonas:
> @abonas this is a serious issue (backported from master) that we want to fix in v0.8.z releasing a v0.8.1.

what is the reason for the need in 0.8.z and not taking the last gem release containing the fix?
btw did you check the watches area? perhaps it is of a need in a similar fix too (I saw briefly someone commenting on content type and watches in the k8s side of things)

simon3z:
> what is the reason for the need in 0.8.z and not taking the last gem release containing the fix?

@abonas because between v0.8.0 and v1.1.1 there are 654 insertions and 96 deletions spread over 17 files. Also, the new dependency on `recursive-open-struct = 1.0.0` and the new `http` gem.

You don't want to bring all those changes to applications that are already in production and they just need this fix.

abonas:
agreed, you have my blessing :)

PR #160, opened 2016-03-22
Add HTTP headers to requests for benefit of the server

garethr:
While testing out the client with OpenShift latest (including kubernetes
v1.2.0-origin) I hit an issue where requests where rejected because of
missing headers. This patch resolved it.

I've tested this out with 1.1 as well and it appears not to cause any
issues.


abonas:
@garethr have you tried the latest gem version 1.1.2 ?

garethr:
@abonas ha, thanks. Perfect timing. This looks fixed in 1.1.2.

PR #161, opened 2016-03-23
Add support for Patch method

zakiva:


simon3z:
@zakiva can you check why this is failing. Thanks.

zakiva:
@simon3z fixed

abonas:
I'd like to understand why this functionality is needed and several aspects around it:
1. the client already provides the update_X methods with PUT, why the patch is needed in addition to it?
2. do all the entities support PATCH on k8s client side? (didn't see a definitive answer in the docs)
3. how the users will differ update_X from patch_X and know when to use which? maybe better names are required, if indeed this functionality shall be added to the client.

zakiva:
@abonas Almost all entities support Patch on Kubernetes side (except for ComponentStatus which does not support other methods as well). Patch may be useful for [late initialization](https://github.com/kubernetes/kubernetes/blob/master/docs/devel/api-conventions.md#late-initialization) or when pushing annotations on entities, maybe @simon3z can point out another use case. 
I agree that having both update and patch may be confusing. One option is to change "update" to "replace" as done in kubectl commands: https://github.com/kubernetes/kubernetes/pull/10468 https://github.com/kubernetes/kubernetes/pull/10374

simon3z:
> 1. the client already provides the update_X methods with PUT, why the patch is needed in addition to it?

http://restcookbook.com/HTTP%20Methods/patch/

1. when you only need to update one field of the resource, PUTting a complete resource representation might be cumbersome and utilizes more bandwidth (you have to GET  and PUT)

2. patching is done atomically on the server side (no risk for races with other concurrent GET/PUT)

3. kubernetes and openshift are supporting it both server side and on their clients (`kubectl patch` and `oc patch`), we should close this gap here.

> 2. do all the entities support PATCH on k8s client side? (didn't see a definitive answer in the docs)

As any other rest-api call it can fail server-side for several reasons. What if it's supported but the user has no permissions to update the entity. We shouldn't enforce any policy on the client side.
If an entity has no patch support it will fail.

> 3. how the users will differ update_X from patch_X and know when to use which? maybe better names are required, if indeed this functionality shall be added to the client.

Update is replacing the entire entity and patch is just a diff that gets applied atomically. The user will use pick the one which is relevant for his use-case.

"Patch" is called `patch` in rest-api, in rest-client, in kubernetes and openshift (both server and client) so I don't think we should change its name.

On the other hand, what we call "update" is called "replace" in kubernetes:

```
$ kubectl help
...
  create         Create a resource by filename or stdin
  replace        Replace a resource by filename or stdin.
  patch          Update field(s) of a resource using strategic merge patch.
```

Although I am fine to keep the existing method with the current name "update".

abonas:
> the client already provides the update_X methods with PUT, why the patch is needed in addition to it?
> http://restcookbook.com/HTTP%20Methods/patch/
> 
> when you only need to update one field of the resource, PUTting a complete resource representation might be cumbersome and utilizes more bandwidth (you have to GET and PUT)
> 
> patching is done atomically on the server side (no risk for races with other concurrent GET/PUT)
> 
> kubernetes and openshift are supporting it both server side and on their clients (kubectl patch and oc patch), we should close this gap here.
> 
> do all the entities support PATCH on k8s client side? (didn't see a definitive answer in the docs)
> As any other rest-api call it can fail server-side for several reasons. What if it's supported but the user has no permissions to update the entity. We shouldn't enforce any policy on the client side.
> If an entity has no patch support it will fail.

yes that is obvious that the server can fail any request for any particular reason and the client doesn't enforce it. my point was that if only one entity supports it on server side, I don't see any reason adding a "massive" support for it for all entities.
 
> how the users will differ update_X from patch_X and know when to use which? maybe better names are required, if indeed this functionality shall be added to the client.
> Update is replacing the entire entity and patch is just a diff that gets applied atomically. The user will use pick the one which is relevant for his use-case.
> 
> "Patch" is called patch in rest-api, in rest-client, in kubernetes and openshift (both server and client) so I don't think we should change its name.
> 
> On the other hand, what we call "update" is called "replace" in kubernetes:
> 
> $ kubectl help
> ...
>   create         Create a resource by filename or stdin
>   replace        Replace a resource by filename or stdin.
>   patch          Update field(s) of a resource using strategic merge patch.
> Although I am fine to keep the existing method with the current name "update".

yes, I would not rename the existing update methods not to break existing users.

I am missing tests and additions to readme in this patch.

lib/kubeclient/common.rb
```diff
@@ -109,6 +109,10 @@ def self.define_entity_methods(entity_types)
         define_method("update_#{entity_name}") do |entity_config|
           update_entity(entity_type, entity_config)
         end
+
+        define_method("patch_#{entity_name}") do |name, namespace = nil, patch|
```
moolitayer: better move the namespace = nil to the end of argument list, otherwise users always have to call
('somename', nil, {somevalues})

zakiva: @moolitayer I think you can still call ('somename', {somevalues}) it worked for me

abonas: @zakiva I agree with @moolitayer that it's better to move the namespace to be the last parameter.
it's also more logical from the perspective of flow: "what to patch, what is the patch content and in which namespace"

zakiva: It sounds to me that according to this flow, name and namespace are both part of the "what to patch". This order is also consistent with the other methods (Get and Delete), where namespace comes right after name. Changing it might be confusing for the user. @simon3z  what do you think?

simon3z: @zakiva :+1: to move `namespace` at the end. The consistency for me is "`namespace` is optional, therefore is always at the end", so for me it's consistent to `get` and `delete` as well.

simon3z:
@zakiva don't forget @abonas's ask for documentation and examples.

zakiva:
@simon3z @abonas done

test/test_service.rb
```diff
@@ -229,4 +229,33 @@ def test_update_service_with_string_keys
...
+    service = Kubeclient::Service.new
+    name = 'my_service'
+
+    service.metadata = {}
+    service.metadata.name      = name
+    service.metadata.namespace = 'development'
+
+    expected_url = "http://localhost:8080/api/v1/namespaces/development/services/#{name}"
+    stub_request(:patch, expected_url)
+      .to_return(body: open_test_file('service_patch.json'), status: 201)
```
abonas: a PATCH method returning 201 (created) ? are you sure this is not 200 or 204?

test/test_service.rb
```diff
@@ -229,4 +229,33 @@ def test_update_service_with_string_keys
...
+
+    patch = {
```
abonas: where is this patch in the service_patch json file?

README.md
```diff
@@ -227,6 +227,19 @@ The below example is for v1
 client.update_service service1
 ```
 
+#### Patch an entity (by name)
+For example: `patch_service "service name"`, `patch_pod "pod name"`, `patch_replication_controller "rc name"`, `patch_secret "secret name"`, `patch_resource_quota "resource quota name"`, `patch_limit_range "limit range name"`, `patch_persistent_volume "persistent volume name"`, `patch_persistent_volume_claim "persistent volume claim name"`, `patch_component_status "component name"`, `patch_service_account "service account name"`
+
+The PATCH request should include the namespace name, except for nodes and namespaces entities.
+
+Input parameter - patch (hash) to be applied to the resource.
```
abonas: this is not the only input parameter, you have also the name of the resource/entity to be updated as well as optionally namespace name as well.

simon3z:
@abonas ack?

README.md
```diff
@@ -227,6 +227,19 @@ The below example is for v1
 client.update_service service1
 ```
 
+#### Patch an entity (by name)
+For example: `patch_service "service name"`, `patch_pod "pod name"`, `patch_replication_controller "rc name"`, `patch_secret "secret name"`, `patch_resource_quota "resource quota name"`, `patch_limit_range "limit range name"`, `patch_persistent_volume "persistent volume name"`, `patch_persistent_volume_claim "persistent volume claim name"`, `patch_component_status "component name"`, `patch_service_account "service account name"`
```
abonas: this section here is incorrect - the patch methods receive as a bare minimum 2 parameters, potentially 3. so the examples above are misleading. Also this section seems a bit unnecessary and too long.

abonas:
> @abonas ack?

@simon3z 
I left a comment

README.md
zakiva: @abonas, changed it to method name only, while the required and optional parameters are specified below (same as the description for update entity). If you still find it unnecessary I can remove it too.  cc @simon3z

simon3z:
@abonas can you check if everything is ok now? Thanks.

abonas:
@simon3z LGTM. have you reviewed it?

simon3z:
:+1:

PR #164, opened 2016-05-01
some cleanup

grosser:
@abonas 
 - reduce argument list size
 - avoid throwaway option hashes
 - avoid repeating of argument defaults

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -2,7 +2,28 @@ require 'json'
 require 'rest-client'
 module Kubeclient
   # Common methods
+  # this is mixed in by other gems
   module ClientMixin
+    DEFAULT_SSL_OPTIONS = {
+      client_cert: nil,
+      client_key:  nil,
+      ca_file:     nil,
+      cert_store:  nil,
+      verify_ssl:  OpenSSL::SSL::VERIFY_PEER
+    }.freeze
+
+    DEFAULT_AUTH_OPTIONS = {
+      username:          nil,
+      password:          nil,
+      bearer_token:      nil,
+      bearer_token_file: nil
+    }.freeze
+
+    DEFAULT_SOCKET_OPTIONS = {
+      socket_class:     nil,
+      ssl_socket_class: nil
+    }.freeze
+
     attr_reader :api_endpoint
     attr_reader :ssl_options
     attr_reader :auth_options
@@ -11,24 +32,10 @@ module Kubeclient
     def initialize_client(
       uri,
       path,
-      version = nil,
-      ssl_options: {
-        client_cert: nil,
-        client_key: nil,
-        ca_file: nil,
-        cert_store: nil,
-        verify_ssl: OpenSSL::SSL::VERIFY_PEER
-      },
-      auth_options: {
-        username: nil,
-        password: nil,
-        bearer_token: nil,
-        bearer_token_file: nil
-      },
-      socket_options: {
-        socket_class: nil,
-        ssl_socket_class: nil
-      }
+      version,
+      ssl_options: DEFAULT_SSL_OPTIONS,
+      auth_options: DEFAULT_AUTH_OPTIONS,
+      socket_options: DEFAULT_SOCKET_OPTIONS
     )
       validate_auth_options(auth_options)
       handle_uri(uri, path)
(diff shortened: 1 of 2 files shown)


lib/kubeclient/common.rb
```diff
@@ -11,24 +11,10 @@ module ClientMixin
...
-        bearer_token_file: nil
-      },
-      socket_options: {
-        socket_class: nil,
-        ssl_socket_class: nil
-      }
+      version,
+      ssl_options: nil,
+      auth_options: nil,
+      socket_options: nil
```
simon3z: @grosser I have the feeling that the defaults were repeated here as well because other libraries (e.g. `openshift_client`) are calling this and they expect the defaults to be there. It's probably safer to put them back in (now that's it's much simpler with the constants you defined).

lib/kubeclient/common.rb
```diff
@@ -50,21 +36,23 @@ def initialize_client(
...
-      fail ArgumentError, 'Missing uri' if uri.nil?
-      @api_endpoint = (uri.is_a? URI) ? uri : URI.parse(uri)
-      @api_endpoint.path = path if @api_endpoint.path.empty?
-      @api_endpoint.path = @api_endpoint.path.chop \
-                         if @api_endpoint.path.end_with? '/'
+      fail ArgumentError, 'Missing uri' unless uri
+      @api_endpoint = (uri.is_a?(URI) ? uri : URI.parse(uri))
+      if @api_endpoint.path.empty?
+        @api_endpoint.path = path
+      elsif @api_endpoint.path.end_with? '/'
```
abonas: This looks as a logic change versus the previous code revision - before this change, the path was assigned it if was empty and then tested for the existence of "/" so it can be chopped, regardless whether it was empty and assigned on the previous line or not.
In the new code revision it seems that due to the if/elsif, it's only one of them going to happen - if path was empty it will be assigned, but then it won't be tested for "/" existence.

lib/kubeclient/common.rb
abonas: so it seems to me as well. see here: https://github.com/abonas/openshift_client/blob/master/lib/openshift_client.rb

lib/kubeclient/common.rb
grosser: I think this was an unnecessary check since an empty path contain a /

grosser:
updated

kubeclient.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler', '~> 1.6'
   spec.add_development_dependency 'rake', '~> 10.0'
   spec.add_development_dependency 'minitest'
-  spec.add_development_dependency 'webmock'
+  spec.add_development_dependency 'webmock', '~> 1.24.2'
```
abonas: this is no longer needed here in this patch

lib/kubeclient/common.rb
abonas: the chopping was done on a path whether it was passed from external param or not. so not just for empty path. unless I'm missing something between the lines.
@simon3z ?

kubeclient.gemspec
grosser: resolved through rebase

lib/kubeclient/common.rb
grosser: no need to call chop on an empty string :)

grosser:
@abonas this good to go too ?

lib/kubeclient/common.rb
grosser: @simon3z / @abonas good to go ?

simon3z: @grosser I agree with @abonas here, you slipped a change in behavior (and I don't understand why you need that).
If someone calls `handle_uri('http://localhost:8443', '/my-path/')` then you wouldn't reach the `chop` part.
Previous code seemed clean enough already.

grosser: yeah true ... somehow got that mixed up and thought it would only call chop on the empty string

grosser: updated ... looks good ?

simon3z:
LGTM :+1:
@abonas LGTY?

abonas:
LGTM

PR #165, opened 2016-05-01
lock down webmock to pass tests

grosser:
@abonas CI is broken / and tests fail locally :(

PR #166, opened 2016-05-01
parallelize rubocop and normal tests for speed and clarity on what broke

grosser:
@abonas

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,7 @@ rvm:
   - "2.3.0"
 sudo: false
 cache: bundler
+script: bundle exec rake $TASK
+env:
+ - TASK=test
+ - TASK=rubocop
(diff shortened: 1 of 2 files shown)


grosser:
screen shot 2016-05-01 at 10 38 23 am

grosser:
(test errors are fixed with https://github.com/abonas/kubeclient/pull/165)

grosser:
@abonas rebased this one ... should be green now

abonas:
how much time does this save? the gem's tests shouldn't take too long to run in the first place.
cc @simon3z

grosser:
takes it from 45s to 30s for ruby 2.0-2.2 ... but I think the main win is visibility, it's now obvious that rubocop failed and not the tests itself

abonas:
I don't have any objections/concerns regarding this change.
@simon3z wdyt?

simon3z:
I am definitely in favor of the `.travis.yml` file change.
On the other hand I don't think we need the complexity introduced in `Rakefile` (but don't block on me for this one).

abonas:
@simon3z yes, I also had thoughts about the Rakefile.

grosser:
updated Rakefile

PR #169, opened 2016-05-05
adding proxy support

enoodle:
Adding a new optional argument ```http_proxy_uri``` that will be passed to ```RestClient``` as the proxy option.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -24,9 +24,12 @@ module Kubeclient
       ssl_socket_class: nil
     }.freeze
 
+    DEFAULT_HTTP_PROXY_URI = nil
+
     attr_reader :api_endpoint
     attr_reader :ssl_options
     attr_reader :auth_options
+    attr_reader :http_proxy_uri
     attr_reader :headers
 
     def initialize_client(
@@ -35,7 +38,8 @@ module Kubeclient
       version,
       ssl_options: DEFAULT_SSL_OPTIONS,
       auth_options: DEFAULT_AUTH_OPTIONS,
-      socket_options: DEFAULT_SOCKET_OPTIONS
+      socket_options: DEFAULT_SOCKET_OPTIONS,
+      http_proxy_uri: DEFAULT_HTTP_PROXY_URI
     )
       validate_auth_options(auth_options)
       handle_uri(uri, path)
@@ -45,6 +49,7 @@ module Kubeclient
       @ssl_options = ssl_options
       @auth_options = auth_options
       @socket_options = socket_options
+      @http_proxy_uri = http_proxy_uri.to_s if http_proxy_uri
 
       if auth_options[:bearer_token]
         bearer_token(@auth_options[:bearer_token])
@@ -134,6 +139,7 @@ module Kubeclient
         verify_ssl: @ssl_options[:verify_ssl],
         ssl_client_cert: @ssl_options[:client_cert],
         ssl_client_key: @ssl_options[:client_key],
+        proxy: @http_proxy_uri,
         user: @auth_options[:username],
         password: @auth_options[:password]
       }
@@ -361,7 +367,8 @@ module Kubeclient
       options = {
         basic_auth_user: @auth_options[:username],
         basic_auth_password: @auth_options[:password],
-        headers: @headers
+        headers: @headers,
+        http_proxy_uri: @http_proxy_uri
       }
 
       if uri.scheme == 'https'
--- a/test/test_kubeclient.rb
+++ b/test/test_kubeclient.rb
@@ -38,6 +38,18 @@ class KubeClientTest < MiniTest::Test
     assert_equal 'http://localhost:8080/api/v1', rest_client.url.to_s
   end
 
+  def test_pass_proxy
+    uri = URI::HTTP.build(host: 'localhost', port: 8080)
+    proxy_uri = URI::HTTP.build(host: 'myproxyhost', port: 8888)
+    client = Kubeclient::Client.new(uri, http_proxy_uri: proxy_uri)
+    rest_client = client.rest_client
+    assert_equal proxy_uri.to_s, rest_client.options[:proxy]
+
+    watch_client = client.watch_pods
+    assert_equal(watch_client.send(:build_client_options)[:proxy][:proxy_address], proxy_uri.host)
+    assert_equal(watch_client.send(:build_client_options)[:proxy][:proxy_port], proxy_uri.port)
+  end
+
   def test_exception
     stub_request(:post, %r{/services})
       .to_return(body: open_test_file('namespace_exception.json'),
--- a/README.md
+++ b/README.md
@@ -140,6 +140,14 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , 'v1',
                                 socket_options: socket_options
 ```
 
+You can also use kubeclient with an http proxy server such as tinyproxy. It can be entered as a string or a URI object
+For example:
+```ruby
+proxy_uri = URI::HTTP.build(host: "myproxyhost", port: 8443)
+client = Kubeclient::Client.new('https://localhost:8443/api/',
+                                :http_proxy_uri => proxy_uri)
+```
+
 ## Examples:
 
 #### Get all instances of a specific entity type
--- a/lib/kubeclient/watch_stream.rb
+++ b/lib/kubeclient/watch_stream.rb
@@ -47,8 +47,23 @@ module Kubeclient
         end
       end
 
+      def using_proxy
+        proxy = @http_options[:http_proxy_uri]
+        return nil unless proxy
+        p_uri = URI.parse(proxy)
+        {
+          proxy_address: p_uri.hostname,
+          proxy_port: p_uri.port,
+          proxy_username: p_uri.user,
+          proxy_password: p_uri.password
+        }
+      end
+
       def build_client_options
-        client_options = { headers: @http_options[:headers] }
+        client_options = {
+          headers: @http_options[:headers],
+          proxy: using_proxy
+        }
         if @http_options[:ssl]
           client_options[:ssl] = @http_options[:ssl]
           client_options[:ssl_socket_class] =


abonas:
1. why is proxy part of ssl options? proxy doesn't have to be ssl related
2. tests?
3. how can it be clear to the user what/which format should be put in the new proxy field?
4. readme updates?

enoodle:
@abonas do you think ```socket_options``` is a better location? or do you want me to add a new ```options``` to pass around?
I think 3 will be clarified when I fix 4, the readme. It is the same format that RestClient expects. (URI for the proxy).

abonas:
why does it belong to socket_options?

abonas:
@simon3z please also review. thanks

enoodle:
I moved it to its own place instead of piggybacking the existing options. Added a test as well.

abonas:
perhaps the proxy should be treated in the input parameters similar to the url itself. iirc it could be passed as both uri or string and then the client takes care of it. but I don't have a strong opinion here.
cc @simon3z

enoodle:
@abonas updated this handling of the proxy uri.

simon3z:
@enoodle what about watches, is there any issue there? (are we depending on any proxy requirement?) Have you tried those? (Especially in regard to timeouts of the connection).

simon3z:
> perhaps the proxy should be treated in the input parameters similar to the url itself. iirc it could be passed as both uri or string and then the client takes care of it. but I don't have a strong opinion here.

I prefer to be consistent, so probably yes we should support both.

enoodle:
@simon3z I am not sure what are watches but the proxy access itself is implemented in RestClient that itself is just passing it down to Net::HTTP (https://github.com/rest-client/rest-client/blob/68edb26431d2e7702a4b2cf6f3dca3f486dea08a/lib/restclient/request.rb#L327).
I used tshark to debug my setup (3 machines with TinyProxy in the middle between the client and the OpenShift host) does that count as a watch?

simon3z:
@enoodle watches are in: https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/watch_stream.rb

enoodle:
@abonas  @simon3z Ok, Thank you. I updated to support watches as well.

lib/kubeclient/common.rb
```diff
@@ -38,6 +39,7 @@ def initialize_client(
       @ssl_options = ssl_options
       @auth_options = auth_options
       @socket_options = socket_options
+      @http_proxy_uri = (http_proxy_uri.is_a? URI) ? http_proxy_uri.to_s : http_proxy_uri
```
simon3z: @enoodle `kind_of?` is slightly preferred.

enoodle: @simon3z rubocop tells me to go with ```is_a?```. do you want it changed anyway?
```
lib/kubeclient/common.rb:42:41: C: Prefer Object#is_a? over Object#kind_of?.
      @http_proxy_uri = (http_proxy_uri.kind_of? URI) ? http_proxy_uri.to_s : http_proxy_uri
```

simon3z:
@enoodle can you check if there's a way to test the proxy with `stub_request`. Current test is only checking that we're setting the proxy, not that it's actually used (maybe it's enough but if we can ensure that it's actually used for the request it would be better).

enoodle:
@simon3z , currently it doesn't work:
```
Run options: --seed 698

# Running:

..........................................E..............................

Finished in 0.517860s, 140.9648 runs/s, 519.4456 assertions/s.

  1) Error:
KubeClientTest#test_proxy_enabled:
WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: GET http://kubehost:8080/api/v1/pods with headers {'Accept'=>'*/*; q=0.5, application/xml', 'Accept-Encoding'=>'gzip, deflate', 'User-Agent'=>'Ruby'}

You can stub this request with the following snippet:

stub_request(:get, "http://kubehost:8080/api/v1/pods").
  with(:headers => {'Accept'=>'*/*; q=0.5, application/xml', 'Accept-Encoding'=>'gzip, deflate', 'User-Agent'=>'Ruby'}).
  to_return(:status => 200, :body => "", :headers => {})

============================================================
    /home/efreiber/.rvm/gems/ruby-2.2.4@miq/gems/webmock-1.24.6/lib/webmock/http_lib_adapters/net_http.rb:114:in `request'
    /home/efreiber/.rvm/gems/ruby-2.2.4@miq/gems/rest-client-1.8.0/lib/restclient/request.rb:270:in `net_http_do_request'
    /home/efreiber/.rvm/gems/ruby-2.2.4@miq/gems/rest-client-1.8.0/lib/restclient/request.rb:418:in `block in transmit'
    /home/efreiber/.rvm/gems/ruby-2.2.4@miq/gems/webmock-1.24.6/lib/webmock/http_lib_adapters/net_http.rb:123:in `start_without_connect'
    /home/efreiber/.rvm/gems/ruby-2.2.4@miq/gems/webmock-1.24.6/lib/webmock/http_lib_adapters/net_http.rb:150:in `start'
    /home/efreiber/.rvm/gems/ruby-2.2.4@miq/gems/rest-client-1.8.0/lib/restclient/request.rb:413:in `transmit'
    /home/efreiber/.rvm/gems/ruby-2.2.4@miq/gems/rest-client-1.8.0/lib/restclient/request.rb:176:in `execute'
    /home/efreiber/.rvm/gems/ruby-2.2.4@miq/gems/rest-client-1.8.0/lib/restclient/request.rb:41:in `execute'
    /home/efreiber/.rvm/gems/ruby-2.2.4@miq/gems/rest-client-1.8.0/lib/restclient/resource.rb:51:in `get'
    /home/efreiber/kubeclient/lib/kubeclient/common.rb:181:in `block in get_entities'
    /home/efreiber/kubeclient/lib/kubeclient/common.rb:53:in `handle_exception'
    /home/efreiber/kubeclient/lib/kubeclient/common.rb:179:in `get_entities'
    /home/efreiber/kubeclient/lib/kubeclient/common.rb:85:in `block (2 levels) in define_entity_methods'
    /home/efreiber/kubeclient/test/test_kubeclient.rb:60:in `test_proxy_enabled'

73 runs, 269 assertions, 0 failures, 1 errors, 0 skips
```

I think that webmock doesn't support the proxy option.

abonas:
@enoodle where is the test_proxy_enabled test in this PR?

enoodle:
@abonas I forgot to push it before because it doesn't work. Added it now.

test/test_kubeclient.rb
```diff
@@ -38,6 +38,28 @@ def test_no_version_passed
     assert_equal 'http://localhost:8080/api/v1', rest_client.url.to_s
   end
 
+  def test_pass_proxy
+    uri = URI::HTTP.build(port: 8080)
+    uri.hostname = 'localhost'
```
abonas: please note that here the "hostname" property is used when building the uri, but on line 44, the "host" property is used for building another uri. can you explain why?

abonas:
@enoodle the travis failures are around VCR exception from what I briefly saw.
also, there are rubocop violations as well.

test/test_kubeclient.rb
```diff
@@ -38,6 +38,28 @@ def test_no_version_passed
...
+    rest_client = client.rest_client
+    assert_equal proxy_uri.to_s, rest_client.options[:proxy]
+
+    watch_client = client.watch_pods
+    assert_equal(watch_client.send(:build_client_options)[:proxy][:proxy_address],
+                 proxy_uri.hostname)
+    assert_equal watch_client.send(:build_client_options)[:proxy][:proxy_port], proxy_uri.port
+  end
+
+  def test_proxy_enabled
```
abonas: Can you please explain what is this method testing?

test/test_kubeclient.rb
enoodle: they are the same thing. I will change to use only ```host```

```
[40] pry(main)> tri = URI::HTTP.build(port: 8888)
=> #<URI::HTTP http::8888>
[41] pry(main)> tri
=> #<URI::HTTP http::8888>
[42] pry(main)> tri.host
=> nil
[43] pry(main)> tri.hostname
=> nil
[44] pry(main)> tri.hostname = "erez"
=> "erez"
[45] pry(main)> tri.host
=> "erez"
```

test/test_kubeclient.rb
enoodle: It's not working.  I uploaded it to show that it fails. It seems like webmock / vcr can't catch the use of the CONNECT method with the proxy.

abonas: @enoodle does this work without webmock? does rest client send the request correctly to the proxy?

enoodle: @abonas yes. It works without webmock.

enoodle:
@simon3z @abonas I removed the test that I couldn't make work. I looked at ```RestClient``` and was that it was relaying on ```net::http``` to implement the proxy and it only tests that those parameters are passed on: https://github.com/rest-client/rest-client/blob/f0e2dc4327e91002187541de6baa0c35b125c549/spec/unit/request_spec.rb#L434

lib/kubeclient/common.rb
simon3z: @enoodle no... but it's funny, in ManageIQ it's backward :smile:

simon3z:
@enoodle what's missing to move this out of WIP? (Beside the fact that we can't test if the proxy is actually used with the stub).

enoodle:
@simon3z nothing. Its working as intended.

simon3z:
@jonmoter @grosser @msufa can you review this?

enoodle:
rebased.

simon3z:
@moolitayer can you review this one? Thanks.

lib/kubeclient/common.rb
```diff
@@ -24,6 +24,8 @@ module ClientMixin
       ssl_socket_class: nil
     }.freeze
 
+    DEFAULT_HTTP_PROXY_URI = nil
+
     attr_reader :api_endpoint
     attr_reader :ssl_options
     attr_reader :auth_options
```
moolitayer: please add the http proxy to the exposed attributes. I'ts quite handy for different use cases like debugging or when users of the client want to know of these properties. (I think we better add socket_options in a different patch)

lib/kubeclient/common.rb
moolitayer: @enoodle always calling ```to_s``` should have the same result. I'm in favor of that since it removes handling of a special case

enoodle: @moolitayer I dont understand

lib/kubeclient/common.rb
enoodle: Added

lib/kubeclient/common.rb
enoodle: Ok, I got it. Changed.

lib/kubeclient/common.rb
```diff
@@ -45,6 +49,7 @@ def initialize_client(
       @ssl_options = ssl_options
       @auth_options = auth_options
       @socket_options = socket_options
+      @http_proxy_uri = http_proxy_uri.to_s if http_proxy_uri
```
moolitayer: :+1:

moolitayer:
LGTM :+1:

simon3z:
LGTM :+1: cc @abonas

test/test_kubeclient.rb
```diff
@@ -38,6 +38,18 @@ def test_no_version_passed
...
 
+  def test_pass_proxy
+    uri = URI::HTTP.build(host: 'localhost', port: 8080)
+    proxy_uri = URI::HTTP.build(host: 'myproxyhost', port: 8888)
+    client = Kubeclient::Client.new(uri, http_proxy_uri: proxy_uri.to_s)
+    rest_client = client.rest_client
+    assert_equal proxy_uri.to_s, rest_client.options[:proxy]
+
+    watch_client = client.watch_pods
+    assert_equal(watch_client.send(:build_client_options)[:proxy][:proxy_address], proxy_uri.host)
```
abonas: please use the same style on lines 48 and 49 - either assert_equal with brackets, or without.

test/test_kubeclient.rb
```diff
@@ -38,6 +38,18 @@ def test_no_version_passed
     assert_equal 'http://localhost:8080/api/v1', rest_client.url.to_s
   end
 
+  def test_pass_proxy
+    uri = URI::HTTP.build(host: 'localhost', port: 8080)
+    proxy_uri = URI::HTTP.build(host: 'myproxyhost', port: 8888)
+    client = Kubeclient::Client.new(uri, http_proxy_uri: proxy_uri.to_s)
```
abonas: misleading that the parameter is suffixed with "uri" however the parameter value is with to_s

abonas:
I had cpl of comments inline in the code, but in addition, please clarify the following:
readme file documents that kubeclient needs to be initialized with this parameter:
`:http_proxy => proxy_uri`
However kubeclient.rb file is not modified in this patch, so I'm unclear on how this is supposed to pass the proxy details to common.rb.
Also,[ in the test](https://github.com/abonas/kubeclient/pull/169/files#diff-1efd65c010e2f18aadd1dd73e84c68fdR44), another parameter name is used that is different from the readme (:http_proxy_uri)

enoodle:
@abonas I fixed your comments on the test file and the type in the README. we dont need to change anything in ```kubeclient.rb``` becase the ```**options``` passes the ```http_proxy_uri``` to ```initialize_client``` in ```lib/kubeclient/common.rb```

PR #179, opened 2016-06-23
rudimentary Kubeclient::Config documentation

quavmo:
We dug up the `Kubeclient::Config`, and this is how we are using it.


abonas:
:+1: 
@simon3z

simon3z:
:+1:

Issue #144, opened 2015-12-11
Support new group API


simon3z:
Kubernetes documentation:

- http://kubernetes.io/v1.1/docs/proposals/api-group.html#server-side-implementation

following the kubernetes description the entities can be reached using the `prefix`, `group` and `version` values.

- for `/api/v1`, we set the `prefix="api"`, `group=""`, `version="v1"`, so the URL remains to be `/api/v1`

A similar approach (client-side) could be used in kubeclient.

- `prefix` and `group` should be preconfigured per-entity (e.g. for all current entities should be `prefix="api"`, `group=""`)
- `version` should be more easily configurable per request

For example:

    client.get_pods
    # prefix="api", group="", version="v1"
    # => /api/v1/pods

For group api (e.g. deployments) we could use just a plain form in the API:

- `client.get_deployments`

but I believe that it would be better to scope them in some way:

- `client.extensions.get_deployments`
- `client.extensions_get_deployments`

and for example:

    client.extensions.get_deployments
    # prefix="apis", group="extensions", version="v1beta1"
    # => /apis/extensions/v1beta1/depoyments

simon3z:
cc @abonas @jimmidyson @msufa @jonmoter

henders:
I see, so you could represent all the entity types like?
```ruby
ENTITY_TYPES = {
  Pod: { prefix: 'api', group: nil, version: 'v1' },
  ReplicationController: { prefix: 'api', group: nil, version: 'v1' },
  Service: { prefix: 'api', group: nil, version: 'v1' },
  Deployment: { prefix: 'apis', group: 'extensions', version: 'v1beta1' },
  ...
}

**Comment by abonas** [comment]
IMO hardcoding entity types in the client like we do today, whether it is done with group/api version or without, is not the optimal design approach.
What I would like to see, is the ability to query the server for api versions and the entity types supported in each api version - because the server "knows best" what does it support. That will allow the flexibility needed and will reduce the number of hardcoded things in the client. @thockin, your thoughts on the topic?

**Comment by henders** [comment]
I'd +1 to that idea of loading the supported APIs from K8s.

In a Rails app, the user could call a method in an initializer to cause Kubeclient to connect to K8s to interrogate K8s over which APIs it supports.
Further advantages is auto-creating methods like ```client.pods.log``` or ```client.pods.exec``` as K8s returns these more specific resource methods in that API list as well:
```
[...]

PR #180, opened 2016-07-26
Added ConfigMap to entity types

zeari:
Kubernetes config maps described here:  http://kubernetes.io/docs/user-guide/configmap/

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -20,7 +20,7 @@ module Kubeclient
     # and especially since currently the class body is empty
     ENTITY_TYPES = %w(Pod Service ReplicationController Node Event Endpoint
                       Namespace Secret ResourceQuota LimitRange PersistentVolume
-                      PersistentVolumeClaim ComponentStatus ServiceAccount).map do |et|
+                      PersistentVolumeClaim ComponentStatus ServiceAccount ConfigMap).map do |et|
       clazz = Class.new(RecursiveOpenStruct) do
         def initialize(hash = nil, args = {})
           args.merge!(recurse_over_arrays: true)
--- a/test/json/config_map_list.json
+++ b/test/json/config_map_list.json
@@ -1,0 +1,24 @@
+{
+  "kind": "ConfigMapList",
+  "apiVersion": "v1",
+  "metadata": {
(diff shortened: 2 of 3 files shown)


zeari:
@abonas Which tests are expected to be added here?

abonas:
@simon3z I am not aware of this feature/requirement, I guess you should have more background on this. Please answer @zeari question re testing.

simon3z:
@abonas @zeari I am not aware of this requirement being currently evaluated.

For testing, I think we'll have to add a `test_config_map.rb` as for the rest of the entities.

Anyway until this is not fixed:

```
Gem::InstallError: activesupport requires Ruby version >= 2.2.2.
```

We won't be able to merge anything (unless we overrule the travis).

abonas:
So why is this frature needed if noone is aware of it at the moment? :)

jonmoter:
I'm in favor of having kubeclient support all the entity types that are available in the Kubernetes API. We use ConfigMaps minimally, but I could see using them more in the future.

zeari:
> I'm in favor of having kubeclient support all the entity types that are available in the Kubernetes API.

Thats pretty much it is. There was a missing entity and i thought id add it...

simon3z:
> Thats pretty much what it is. There was a missing entity and i thought id add it...

@zeari yes but IIUC what @moolitayer is working on in #144 is covering these additional entities automatically. Unless we need ConfigMaps ASAP, the switch to the new group API seems a more interesting investment going forward.

What instead is really urgent to fix (cc @jonmoter) is the current failure we have with activesupport requiring Ruby version >= 2.2.2, which is blocking all the PR going forward.

I pinpointed this to be an issue related to rubygems 2.4.5 (with rubygems 2.4.8 the required ruby version is taken in account), but anyway travis only has rubygems 2.4.5.

We'll probably need a workaround for that in our gemspec for the time being.

moolitayer:
@zeari fyi you test config maps on #181:
```
client = Kubeclient::Client.new 'http://localhost:8080/api' , 'v1', ssl_options: ssl_options
client.get_config_maps
```

zeari:
Closing in favor of #181

PR #181, opened 2016-07-27
Support kubernetes group api

### LINKED ISSUE #144: Support new group API

moolitayer:
[kubernetes api-group proposal](https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/api-group.md)
[kubernetes api](https://github.com/kubernetes/kubernetes/blob/master/docs/api.md)

[specification](https://github.com/abonas/kubeclient/issues/144#issuecomment-235254248)

[Some testing code](https://gist.github.com/moolitayer/476ed01d377fc9e9954fea83572ba662)
(Devided into core, groups & OpenShift)

Fixes #144 
Does not handle nested (e.g pods/status) and first level (e.g /healthz) resources

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -77,6 +77,8 @@ module Kubeclient
       @api_endpoint = (uri.is_a?(URI) ? uri : URI.parse(uri))
       @api_endpoint.path = path if @api_endpoint.path.empty?
       @api_endpoint.path = @api_endpoint.path.chop if @api_endpoint.path.end_with? '/'
+      components = @api_endpoint.path.to_s.split('/') # ["", "api"] or ["", "apis", batch]
+      @group = components[2] if components.length > 2
     end
 
     def build_namespace_prefix(namespace)
@@ -230,6 +232,7 @@ module Kubeclient
         hash[:kind] = entity_type
       end
       hash[:apiVersion] = @api_version
+      hash[:apiVersion] = @group + '/' + @api_version if @group
       @headers['Content-Type'] = 'application/json'
       response = handle_exception do
         rest_client[ns_prefix + resource_name(entity_type)]
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -68,6 +68,7 @@ module Kubeclient
         {}
       end
       err_message = json_error_msg['message'] || e.message
+      raise NameError.new if json_error_msg && json_error_msg['reason'] == 'NotFound'
       raise KubeException.new(e.http_code, err_message, e.response)
     end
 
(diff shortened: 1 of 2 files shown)


simon3z:
@enoodle @zeari @yaacov @zakiva please review.

lib/kubeclient/common.rb
```diff
@@ -71,68 +80,52 @@ def handle_exception
...
-          update_entity(entity_type, entity_config)
-        end
-
-        define_method("patch_#{entity_name}") do |name, patch, namespace = nil|
-          patch_entity(entity_type, name, patch, namespace)
+          args = [singular, options]
+        else
+          args = [singular] + args
         end
+        puts "calling: #{method_name}(#{args})"
```
enoodle: Are we sure that we want to have this ```puts``` here?

lib/kubeclient/common.rb
```diff
@@ -267,8 +261,23 @@ def patch_entity(entity_type, name, patch, namespace = nil)
       end
     end
 
-    def new_entity(hash, klass)
-      klass.new(hash)
+    def new_entity(hash, entity_type)
+      resource_class(entity_type).new(hash)
+    end
+
+    def resource_class(entity_type)
```
enoodle: :+1:

lib/kubeclient/common.rb
```diff
@@ -267,8 +261,23 @@ def patch_entity(entity_type, name, patch, namespace = nil)
...
+      # TODO: test binding since there already seems to be a constant with that name
+      Kubeclient.const_get(entity_type)
+      rescue NameError
+        Kubeclient.const_set(
```
enoodle: Can we check if the entity actually exists?

enoodle:
As far as I understand this doesn't break the old API at all. e.g the class definitions can still be used. Am I right?

moolitayer:
> As far as I understand this doesn't break the old API at all. e.g the class definitions can still be used. Am I right?

I'm afraid It does, I've exapalained why it is problematic to maintain that backward compitability in this [comment](https://github.com/abonas/kubeclient/issues/144#issuecomment-235254248) see Discovery

lib/kubeclient/common.rb
moolitayer: > Can we check if the entity actually exists?

When a request is made for a resource that does not exist we raise a [name error](https://github.com/abonas/kubeclient/pull/181/files#diff-104b3d1973988933ffc83feada470da2R71).
That is done per your previous suggestion, to be backward compatible.

moolitayer:
@simon3z @abonas please review

lib/kubeclient/common.rb
```diff
@@ -68,6 +68,7 @@ def handle_exception
         {}
       end
       err_message = json_error_msg['message'] || e.message
+      raise NameError.new if json_error_msg && json_error_msg['reason'] == 'NotFound'
```
simon3z: @moolitayer isn't "NotFound" used also when you look for a specific entity and it's not found? (e.g. `get_pod("my_pod")`).

lib/kubeclient/common.rb
```diff
@@ -76,57 +77,49 @@ def handle_uri(uri, path)
...
+        options = { resource_version: options } unless options.is_a?(Hash)
+        args = [singular, options]
+      else
+        args = [singular] + args
+      end
+      send(method_name.to_sym, *args)
+    end
+
+    def respond_to?(method_sym)
+      resource_method?(method_sym) ? true : super
```
abonas: who would be the `super` here?

lib/kubeclient/common.rb
moolitayer: Right :(
The error message is different. I'm wondering If we could use that?
For non existing types we get:
```"message": "the server could not find the requested resource",```
For non existing entities:
```"message": "nodes \"127.0.0.12\" not found",```

File: lib/kubeclient/common.rb (hunk 5 above)
Comment (moolitayer) [inline lib/kubeclient/common.rb]: The including object;
```
> client.hi
NoMethodError: undefined method `hi' for #<Kubeclient::Client:0x00000003ed5118>
```

simon3z:
@moolitayer can you rebase this? I want to merge this ASAP so that it will be tested by everyone longer.

moolitayer:
This is more then a rebase. The approach would need to change since it relay's on activesupport (with a robust implementation to handle entities we now know nothing about). I think there are solutions and I will be looking into this.

moolitayer:
Now that #184 was merged I'm dropping this solution in favor of #185

PR #183, opened 2016-08-16
Add support for processing templates

zakiva:
This PR adds support for processing templates. 
A template allows the user to generate a list of objects with certain labels and parameters. This list can be then used in order to create these objects. So, we would like to support processing the template and return the list of objects to create (as part of the processed template). 
Template is currently an [Openshift entity](https://docs.openshift.com/enterprise/3.2/dev_guide/templates.html), and will be added to k8s [here](https://github.com/kubernetes/kubernetes/pull/25293). 
cc @abonas @simon3z @zeari

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,6 +381,16 @@ watcher.each do |line|
 end
 ```
 
+#### Process a template
+Returns a processed template containing a list of objects to create.
+Input parameter - template (hash)
+Besides its metadata, the template should include a list of objects to be processed and a list of parameters
+to be substituted. Note that for a required parameter that does not provide a generated value, you must supply a value.
+
+```ruby
+client.process_template template
+```
+
 ## Upgrading
 
 #### past version 1.2.0
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -405,6 +405,15 @@ module Kubeclient
       end
     end
 
+    def process_template(template)
+      ns_prefix = build_namespace_prefix(template[:metadata][:namespace])
+      response = handle_exception do
+        rest_client[ns_prefix + 'processedtemplates']
+        .post(template.to_h.to_json, { 'Content-Type' => 'application/json' }.merge(@headers))
+      end
+      JSON.parse(response)
+    end
+
     def api_valid?
       result = api
       result.is_a?(Hash) && (result['versions'] || []).any? do |group|
--- a/test/test_process_template.rb
+++ b/test/test_process_template.rb
@@ -1,0 +1,44 @@
+require 'test_helper'
+
+# Process Template tests
+class TestProcessTemplate < MiniTest::Test
+  def test_process_template
+    client = Kubeclient::Client.new 'http://localhost:8080/api/', 'v1'
+    template = {}
+    template[:metadata] = {}
+    template[:metadata][:name] = 'my-template'
+    template[:metadata][:namespace] = 'default'
+    template[:kind] = 'Template'
+    template[:apiVersion] = 'v1'
+    service = {}
+    service[:metadata] = {}
+    service[:metadata][:name] = '${NAME_PREFIX}my-service'
+    service[:kind] = 'Service'
+    service[:apiVersion] = 'v1'
+    template[:objects] = [service]
+    param = { name: 'NAME_PREFIX', value: 'test/' }
+    template[:parameters] = [param]
+
+    req_body = "{\"metadata\":{\"name\":\"my-template\",\"namespace\":\"default\"},"\
+    "\"kind\":\"Template\",\"apiVersion\":\"v1\",\"objects\":[{\"metadata\":"\
+    "{\"name\":\"${NAME_PREFIX}my-service\"},\"kind\":\"Service\",\"apiVersion\":\"v1\"}],"\
+    "\"parameters\":[{\"name\":\"NAME_PREFIX\",\"value\":\"test/\"}]}"
+
+    expected_url = 'http://localhost:8080/api/v1/namespaces/default/processedtemplates'
+    stub_request(:post, expected_url)
+      .with(body: req_body, headers: { 'Content-Type' => 'application/json' })
+      .to_return(body: open_test_file('processed_template.json'), status: 200)
+
+    processed_template = client.process_template template
+
+    assert_equal('test/my-service', processed_template['objects'].first['metadata']['name'])
+
+    assert_requested(:post, expected_url, times: 1) do |req|
+      data = JSON.parse(req.body)
+      data['kind'] == 'Template' &&
+        data['apiVersion'] == 'v1' &&
+        data['metadata']['name'] == 'my-template' &&
+        data['metadata']['namespace'] == 'default'
+    end
+  end
+end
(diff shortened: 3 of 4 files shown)


abonas:
are you going to add tests? what is this feature about/link to k8s documentation?

zakiva:
> are you going to add tests? what is this feature about/link to k8s documentation?

yes, I marked it with WIP since tests and updating the Readme are still missing. I'll also update the description above.

zakiva:
@moolitayer @zeari Please review

abonas:
Currently the feature is not merged on k8s side and that PR wasn't updated for a while. 
So either this support should be added to openshift ruby client, or be blocked from merging here until it's merged in k8s - to be on the safe side for users who are using k8s and not openshift.
cc @simon3z

README.md
```diff
@@ -356,6 +356,13 @@ watcher.each do |line|
 end
 ```
 
+#### Process a template
+Returns a processed template containing a list of objects to create.
+Input parameter - template (hash)
+
+```ruby
+client.process_template template
+```
```
abonas: any example/rules how a template should look like? any mandatory fields?

simon3z:
> Currently the feature is not merged on k8s side and that PR wasn't updated for a while. 
> So either this support should be added to openshift ruby client, or be blocked from merging here until it's merged in k8s - to be on the safe side for users who are using k8s and not openshift.

Yes, @zakiva this should be move to the openshift client.

zakiva:
moved to [openshift client](https://github.com/abonas/openshift_client/pull/29), closing.

zakiva:
@simon3z @zeari @moolitayer reopened, please take another look.

lib/kubeclient/common.rb
```diff
@@ -405,6 +405,17 @@ def proxy_url(kind, name, port, namespace = '')
       end
     end
 
+    def process_template(template)
+      ns_prefix = build_namespace_prefix(template[:metadata][:namespace])
+      @headers['Content-Type'] = 'application/json'
+      response = handle_exception do
+        rest_client[ns_prefix + 'processedtemplates']
+        .post(template.to_h.to_json, @headers)
```
enoodle: Question: I dont understand how this is different from `create_template`. They are both using `post` on the received template hash. Can you please explain how this works?

zakiva: Both indeed use ```Post``` on a template, but while ```create_template``` actually creates a template, ```process_template``` does not. It is only used to substitute the template parameters and generate their values (if required), in order to get a list of objects to create from the template. Note that we need here a separated method because although both requests are using ```Post``` they are still different: 

```POST /oapi/v1/namespaces/{namespace}/templates``` for creating a template. 

```POST /oapi/v1/namespaces/{namespace}/processedtemplates``` for processing a template.

Comment (enoodle) [inline lib/kubeclient/common.rb]: :+1: Thank you

File: lib/kubeclient/common.rb [hunk 3]
```diff
@@ -405,6 +405,17 @@ def proxy_url(kind, name, port, namespace = '')
...
+      end
+      result = JSON.parse(response)
+      result
```
moolitayer: style: replace the above two lines with
```
JSON.parse(response)
```

lib/kubeclient/common.rb
```diff
@@ -405,6 +405,17 @@ def proxy_url(kind, name, port, namespace = '')
       end
     end
 
+    def process_template(template)
+      ns_prefix = build_namespace_prefix(template[:metadata][:namespace])
+      @headers['Content-Type'] = 'application/json'
```
moolitayer: Actually we do not want to update @headers except for this request.
Please follow https://github.com/abonas/kubeclient/pull/190 to clean this up.

lib/kubeclient/common.rb
zakiva: done, thanks.

lib/kubeclient/common.rb
zakiva: done, thanks.

lib/kubeclient/common.rb
```diff
@@ -405,6 +405,15 @@ def proxy_url(kind, name, port, namespace = '')
       end
     end
 
+    def process_template(template)
+      ns_prefix = build_namespace_prefix(template[:metadata][:namespace])
+      response = handle_exception do
+        rest_client[ns_prefix + 'processedtemplates']
+        .post(template.to_h.to_json, { 'Content-Type' => 'application/json' }.merge(@headers))
+      end
+      JSON.parse(response)
```
simon3z: @zakiva @moolitayer I thought this was taken care of by the rest client.

moolitayer: No and I don't see rest_client has support for that. Our exception handler parses exceptions but other then that we parse the response manually on all calls but get_pod_log. I recently thought about wrapping some of the rest_client logic since it is repetitive. Maybe that belongs there too. Not in this PR though

moolitayer:
LGTM :+1:

moolitayer:
Since last comments regarding kubeclinet vs openshift_client from 24 days ago some things changed:

- we are looking into replacing openshift_client with a kubeclient with discovery on the openshift endpoint - see openshift_client [here](https://github.com/abonas/kubeclient/pull/185#issue-173817703) 
- The only real functional difference between the two would be the new process_template method as clients were kept functionally similar up until this point (which is great for us)
- Given those fact and that the issue in question is targeted to reach kubernetes at some point our cleanest direction for these two clients would be to do the process in kubeclient

@abonas @jonmoter @jimmidyson @nyxcharon @Fryguy @zeari @yaacov @cben @grosser 

Any comment on this? We'd like to merge this ASAP

simon3z:
> Any comment on this? We'd like to merge this ASAP

LGTM :+1:, I haven't reviewed the code closely, maybe there are nits to fix, but I agree with the direction.

zeari:
LGTM :+1:

test/test_process_template.rb
```diff
@@ -0,0 +1,44 @@
...
+    service[:kind] = 'Service'
+    service[:apiVersion] = 'v1'
+    template[:objects] = [service]
+    param = { name: 'NAME_PREFIX', value: 'test/' }
+    template[:parameters] = [param]
+
+    req_body = "{\"metadata\":{\"name\":\"my-template\",\"namespace\":\"default\"},"\
+    "\"kind\":\"Template\",\"apiVersion\":\"v1\",\"objects\":[{\"metadata\":"\
+    "{\"name\":\"${NAME_PREFIX}my-service\"},\"kind\":\"Service\",\"apiVersion\":\"v1\"}],"\
+    "\"parameters\":[{\"name\":\"NAME_PREFIX\",\"value\":\"test/\"}]}"
```
enoodle: Is this the best way to create the request body?
Do we really need a request body here? I see that in other tests we just mocked without this and it seems to be enough.

salilgupta1:
Might be nice to clarify in the README that process_template is not something that is supported by k8s right now.

simon3z:
> Might be nice to clarify in the README that process_template is not something that is supported by k8s right now.

Right @salilgupta1, can you or @zakiva send a PR to update the README file?

PR #184, opened 2016-08-16
Drop the dependency on activesupport

### LINKED ISSUE #182: activesupport requires Ruby version >= 2.2.2
The `activesupport` dependency [kubeclient has](https://github.com/abonas/kubeclient/blob/3409365eb20171acef57162cf0abc101381474fd/kubeclient.gemspec#L29):

```ruby
# ...
Gem::Specification.new do |spec|
  # ...
  spec.add_dependency 'activesupport'
  # ...
end
```

is breaking travis jobs for Ruby < 2.2.2 on:

```
Gem::InstallError: activesupport requires Ruby version >= 2.2.2.
An error occurred while installing activesupport (5.0.0), and Bundler cannot
continue.
Make sure that `gem install activesupport -v '5.0.0'` succeeds before bundling.
The command "eval bundle install --jobs=3 --retry=3" failed. Retrying, 2 of 3.
```

The gem installation on Ruby < 2.2.2 fails as well (when activesupport is not installed yet):

```
$ ruby -v
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-linux]
$ gem install kubeclient
ERROR:  Error installing kubeclient:
        activesupport requires Ruby version >= 2.2.2.
```

Unless there is a way to have a conditional dependency (`activesupport < 5` for Ruby < 2.2.2) we should probably consider to drop support for Ruby < 2.2.2.

Other possible options and workarounds:

- find out if `activesupport` has any suggestion on how other projects should specify the dependency
- force a specific activesupport (< 5.0) version in the gem file (dependency issues with other projects)
- find an alternative to `activesupport`
- maybe have a specific workaround for travis (although it's not solving the gem installation issue):

```ruby
  if ENV['TRAVIS'] == 'true' && RUBY_VERSION < '2.2.2'
    spec.add_dependency 'activesupport', '< 5'
  else
    spec.add_dependency 'activesupport'
  end
```

cc @zakiva @zeari @moolitayer @abonas @Fryguy @bdunne @jonmoter @jimmidyson

moolitayer:
Fix #182

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -26,6 +26,13 @@ module Kubeclient
 
     DEFAULT_HTTP_PROXY_URI = nil
 
+    SEARCH_ARGUMENTS = {
+      'labelSelector' => :label_selector,
+      'fieldSelector' => :field_selector
+    }.freeze
+
+    WATCH_ARGUMENTS = { 'resourceVersion' => :resource_version }.merge!(SEARCH_ARGUMENTS).freeze
+
     attr_reader :api_endpoint
     attr_reader :ssl_options
     attr_reader :auth_options
@@ -86,7 +93,7 @@ module Kubeclient
 
     def self.define_entity_methods(entity_types)
       entity_types.each do |klass, entity_type|
-        entity_name = entity_type.underscore
+        entity_name = underscore_entity(entity_type)
         entity_name_plural = pluralize_entity(entity_name)
 
         # get all entities of a type e.g. get_nodes, get_pods, etc.
@@ -127,8 +134,11 @@ module Kubeclient
     end
 
     def self.pluralize_entity(entity_name)
-      return entity_name + 's' if entity_name.end_with? 'quota'
-      entity_name.pluralize
+      "#{entity_name}#{entity_name.end_with?('s') ? 'es' : 's'}"
+    end
+
+    def self.underscore_entity(entity_name)
+      entity_name.gsub(/([a-z])([A-Z])/, '\1_\2').downcase
     end
 
     def create_rest_client(path = nil)
@@ -165,10 +175,9 @@ module Kubeclient
       path += "/#{options[:name]}" if options[:name]
       uri = @api_endpoint.merge("#{@api_endpoint.path}/#{@api_version}/#{path}")
 
-      params = options.slice(:label_selector, :field_selector, :resource_version)
-      if params.any?
-        uri.query = URI.encode_www_form(params.map { |k, v| [k.to_s.camelize(:lower), v] })
-      end
+      params = {}
+      WATCH_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
+      uri.query = URI.encode_www_form(params) if params.any?
 
       Kubeclient::Common::WatchStream.new(uri, http_options(uri))
     end
@@ -179,9 +188,7 @@ module Kubeclient
     #   :field_selector - a selector to restrict the list of returned objects by their fields.
     def get_entities(entity_type, klass, options = {})
       params = {}
-      [:label_selector, :field_selector].each do |p|
-        params[p.to_s.camelize(:lower)] = options[p] if options[p]
-      end
+      SEARCH_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
 
       ns_prefix = build_namespace_prefix(options[:namespace])
       response = handle_exception do
@@ -275,9 +282,9 @@ module Kubeclient
       entity_types.each_with_object({}) do |(_, entity_type), result_hash|
         # method call for get each entities
         # build hash of entity name to array of the entities
-        entity_name = ClientMixin.pluralize_entity entity_type.underscore
+        entity_name = ClientMixin.pluralize_entity(ClientMixin.underscore_entity(entity_type))
         method_name = "get_#{entity_name}"
-        key_name = entity_type.underscore
+        key_name = ClientMixin.underscore_entity(entity_type)
         result_hash[key_name] = send(method_name)
       end
     end
@@ -310,7 +317,11 @@ module Kubeclient
     end
 
     def proxy_url(kind, name, port, namespace = '')
-      entity_name_plural = ClientMixin.pluralize_entity(kind.to_s)
+      entity_name_plural = if %w(services pods nodes).include?(kind.to_s)
+                             kind.to_s
+                           else
+                             ClientMixin.pluralize_entity(kind.to_s)
+                           end
       ns_prefix = build_namespace_prefix(namespace)
       # TODO: Change this once services supports the new scheme
       if entity_name_plural == 'pods'
--- a/test/test_common.rb
+++ b/test/test_common.rb
@@ -1,0 +1,60 @@
+require 'test_helper'
+
+# Unit tests for the common module
+class CommonTest < MiniTest::Test
+  def test_pluralize_entity
+    %w(
+      Pod Pods
+      Service Services
+      ReplicationController ReplicationControllers
+      Node Nodes
+      Event Events
+      Endpoint Endpoints
+      Namespace Namespaces
+      Secret Secrets
+      ResourceQuota ResourceQuotas
+      LimitRange LimitRanges
+      PersistentVolume PersistentVolumes
+      PersistentVolumeClaim PersistentVolumeClaims
+      ComponentStatus ComponentStatuses
+      ServiceAccount ServiceAccounts
+      Project Projects
+      Route Routes
+      ClusterRoleBinding ClusterRoleBindings
+      Build Builds
+      BuildConfig BuildConfigs
+      Image Images
+      ImageStream ImageStreams
+    ).each_slice(2) do |singular, plural|
+      assert_equal(Kubeclient::ClientMixin.pluralize_entity(singular), plural)
+    end
+  end
+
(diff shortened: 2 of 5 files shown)


moolitayer:
@simon3z @jonmoter @jimmidyson PTAL

test/test_kubeclient.rb
```diff
@@ -519,15 +519,15 @@ def test_proxy_url
                  client.proxy_url('service', 'srvname', 'srvportname', 'ns'))
 
     assert_equal('http://host:8080/api/v1/namespaces/ns/pods/srvname:srvportname/proxy',
-                 client.proxy_url('pods', 'srvname', 'srvportname', 'ns'))
+                 client.proxy_url('pod', 'srvname', 'srvportname', 'ns'))
```
simon3z: @moolitayer from this it seems that what you changed is not backward compatible

moolitayer: Right but using plural here is against what we say in the README as well as #116 .
We can retain support by adding a condition in proxy_url.

ps https://github.com/kubernetes/kubernetes/issues/18622 is interesting (though not really related)

moolitayer:
Retain backward compatibility for plural entity names in proxy_url and test it.

moolitayer:
PS For reference currently supported entities are:

## kubeclient
Pod
Pods
Service
Services
ReplicationController
ReplicationControllers
Node
Nodes
Event
Events
Endpoint
Endpoints
Namespace
Namespaces
Secret
Secrets
ResourceQuota
ResourceQuotas
LimitRange
LimitRanges
PersistentVolume
PersistentVolumes
PersistentVolumeClaim
PersistentVolumeClaims
ComponentStatus
ComponentStatuses
ServiceAccount
ServiceAccounts

## openshift_client
Project
Projects
Route
Routes
ClusterRoleBinding
ClusterRoleBindings
Build
Builds
BuildConfig
BuildConfigs
Image
Images
ImageStream
ImageStreams

lib/kubeclient/common.rb
```diff
@@ -310,7 +315,7 @@ def watch_pod_log(pod_name, namespace, container: nil)
     end
 
     def proxy_url(kind, name, port, namespace = '')
-      entity_name_plural = ClientMixin.pluralize_entity(kind.to_s)
+      entity_name_plural = kind.to_s.end_with?('s')? kind.to_s : ClientMixin.pluralize_entity(kind.to_s)
```
jonmoter: Would this cause a problem for the `ComponentStatus` object?

moolitayer: I believe that entity does not have a proxy.

lib/kubeclient/common.rb
```diff
@@ -127,8 +127,11 @@ def self.define_entity_methods(entity_types)
     end
 
     def self.pluralize_entity(entity_name)
-      return entity_name + 's' if entity_name.end_with? 'quota'
-      entity_name.pluralize
+      entity_name.end_with?('s') ? entity_name + 'es' : entity_name + 's'
```
Fryguy: Prefer interpolation over `+`.  The latter creates extra strings which aren't needed.

lib/kubeclient/common.rb
```diff
@@ -127,8 +127,11 @@ def self.define_entity_methods(entity_types)
...
+    end
+
+    def self.underscore_entity(entity_name)
+      entity_name.gsub(/([a-z])([A-Z])/, '\1_\2').downcase
```
Fryguy: Be careful as this breaks up something like `AnXMLParser` into `an_xmlparser`.  Might be ok, but .underscore would break it up as `an_xml_parser`

lib/kubeclient/common.rb
```diff
@@ -310,7 +315,11 @@ def watch_pod_log(pod_name, namespace, container: nil)
     end
 
     def proxy_url(kind, name, port, namespace = '')
-      entity_name_plural = ClientMixin.pluralize_entity(kind.to_s)
+      entity_name_plural = if kind.to_s.end_with?('s')
```
Fryguy: Might be better to have a `ClientMixin.plural?` method (or just let pluralize_entity do the check for you).  Note that not all words that end in 's' are plural, so that could be a potential problem.  Ironcally, "Kubernetes" is one of those words, but you don't pluralize that :).

moolitayer: This one has to support a narrower list and plural code provides backward compatibility.
I've changed the code to be very explicit

lib/kubeclient/common.rb
moolitayer: This method has to handle very specific, known in advance, values. I've added a test with supported values.

lib/kubeclient/common.rb
```diff
@@ -310,7 +315,11 @@ def watch_pod_log(pod_name, namespace, container: nil)
     end
 
     def proxy_url(kind, name, port, namespace = '')
-      entity_name_plural = ClientMixin.pluralize_entity(kind.to_s)
+      entity_name_plural = if %w(services pods nodes).include?(kind.to_s)
```
simon3z: @moolitayer this is ugly but I don't have suggestions ATM.

lib/kubeclient/common.rb
```diff
@@ -165,9 +168,12 @@ def watch_entities(entity_type, options = {})
       path += "/#{options[:name]}" if options[:name]
       uri = @api_endpoint.merge("#{@api_endpoint.path}/#{@api_version}/#{path}")
 
-      params = options.slice(:label_selector, :field_selector, :resource_version)
+      params = {}
+      params['labelSelector'] = options[:label_selector] if options[:label_selector]
+      params['fieldSelector'] = options[:field_selector] if options[:field_selector]
+      params['resourceVersion'] = options[:resource_version] if options[:resource_version]
```
simon3z: @moolitayer maybe you could keep a map for this transformation.

lib/kubeclient/common.rb
moolitayer: Previously the code was nicer
```if kind.to_s.end_with?('s')```
But this makes it clearer this is for backward comp that only handles 3 entities.

Comment (simon3z) [inline lib/kubeclient/common.rb]: @moolitayer OK I guess we'll be dropping this in your group api PR.

**Comment by simon3z** [comment]
@moolitayer I think you should just handle the travis failure for now. LGTM :+1:

**Comment by simon3z** [comment]
@jonmoter @abonas can you review / ack?

**Comment by jonmoter** [comment]
Looks good to me. 👍  when the Travis build is 💚

File: lib/kubeclient/common.rb [hunk 8]
```diff
@@ -165,9 +175,10 @@ def watch_entities(entity_type, options = {})
       path += "/#{options[:name]}" if options[:name]
       uri = @api_endpoint.merge("#{@api_endpoint.path}/#{@api_version}/#{path}")
 
-      params = options.slice(:label_selector, :field_selector, :resource_version)
+      params = {}
+      WATCH_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
       if params.any?
-        uri.query = URI.encode_www_form(params.map { |k, v| [k.to_s.camelize(:lower), v] })
+        uri.query = URI.encode_www_form(params.map { |k, v| [k, v] })
```
Fryguy: `params.map { |k, v| [k, v] }` can be simplified to `params.to_a`

Fryguy: Actually, now that I see it, you can pass params directly into the encode_www_form without translation into an intermediate Array:

```ruby
[1] pry(main)> params = {:a => 1, :b => 2}
=> {:a=>1, :b=>2}
[2] pry(main)> URI.encode_www_form(params.to_a)
=> "a=1&b=2"
[3] pry(main)> URI.encode_www_form(params)
=> "a=1&b=2"
```

lib/kubeclient/common.rb
```diff
@@ -26,6 +26,13 @@ module ClientMixin
 
     DEFAULT_HTTP_PROXY_URI = nil
 
+    SEARCH_ARGUMENTS = {
+      'labelSelector' => :label_selector,
+      'fieldSelector' => :field_selector
+    }.freeze
+
+    WATCH_ARGUMENTS = { 'resourceVersion' => :resource_version }.merge(SEARCH_ARGUMENTS).freeze
```
abonas: why not `merge!` ?

test/test_common.rb
```diff
@@ -0,0 +1,60 @@
+require 'test_helper'
+
+# Unit tests for the common module
+class CommonTest < MiniTest::Test
```
abonas: the test looks great, the only thing I wonder about is whether openshift entities should be tested here and not in their  respective client

moolitayer: thanks. It should be tested here since it tests code that is here and we want the test to fail if that code breaks.
this should be dropped as part of the group api patch (that would need to change as it currently relays on pluralize)

simon3z:
I am not sure if we should address the [current outstanding comment](https://github.com/abonas/kubeclient/pull/184#discussion_r75645993).

Overall I'd like to see this merged (unless critical issues) :+1: 
@abonas what do you think? LGTY or should @moolitayer address your comment?

abonas:
> I am not sure if we should address the current outstanding comment.
> 
> Overall I'd like to see this merged (unless critical issues)  
> @abonas what do you think? LGTY or should @moolitayer address your comment?

it's not critical

PR #185, opened 2016-08-29
Support k8s api discovery and group api

### LINKED ISSUE #144: Support new group API

### LINKED ISSUE #189: jobs support
Please support jobs from /apis/extensions/v1beta1/jobs

moolitayer:
#### Creating the client
Currently we support arguments for kubernetes api path and version and guess '/api' and 'v1' if not provided.
Both of these are valid today and create a client for the same endpoint:
```
client = Kubeclient::Client.new('http://localhost:8080/api', 'v1')
client = Kubeclient::Client.new('http://localhost:8080')
```

It is now possible to create a client for a group api:
```
client = Kubeclient::Client.new('http://localhost:8080/apis/batch', 'v1')
```
#### Discovery
Discovery is done lazily on method calls so it should not change behavior. 

It can also  be done explicitly:
```
client = Kubeclient::Client.new('http://localhost:8080/api', 'v1')
client.discover
```

It is possible to check the status of discovery
```
unless client.discovered
  client.discover
end
```

#### Class definitions

It is no longer possible to support the following code:
```
require 'kubeclient'

service = Kubeclient::Service.new
```

Since classes are no longer statically defined they will not be available at require time.

One solution for this added here is to use the generic `Resource`:
```
service = Kubeclient::Resource.new

```
used to fix some of the tests. I suggest to use this everywhere (including the Readme) since specific entity classes hold no logic and the following will work:

```
client.create_pod(Kubeclient::Service.new)
``` 

Class definitions after the discovery would work but I do not suggest to relay on that .

### Resources in need of special attention
Here is an example for how a normal resource looks in discovery:
```
    {
      "name": "componentstatuses",
      "namespaced": false,
      "kind": "ComponentStatus"
    },
```

####  Endpoints
```
    {
      "name": "endpoints",
      "namespaced": true,
      "kind": "Endpoints" # kind is plural!
    },
```
See https://github.com/kubernetes/kubernetes/issues/8115 complete backward compatibility is currently maintained with specific code in discovery and Endpoint creation. 

As a side note, it would be possible to drop the special handling (and change interface) if we follow https://github.com/kubernetes/kubernetes/issues/18622 and oc client and have kubeclient react the same to plural and singular methods. to distinguish between get one and get all we would have to look at the parameters but that seems completely doable.

#### Resources not supporting GET
```
    {
      "name": "bindings",
      "namespaced": true,
      "kind": "Binding"
    },
```
Bindings is an example for a resource that does not support get, all_entities ignores exceptions so that it keeps working. I don't think there is a way to learn that it does not support get in discovery except for queering it. 

#### Non verbs
Some resources do not fall under the logic kubeclient currently support and should be ignored by discovery. These include:
From k8s /apis/extensions/v1beta1

```
    {
      "name": "replicationcontrollers",
      "namespaced": true,
      "kind": "ReplicationControllerDummy"
    },
```

 See also: https://github.com/kubernetes/kubernetes/issues/22413

OpenShift:
 
```
    {
      "name": "generatedeploymentconfigs",
      "namespaced": true,
      "kind": "DeploymentConfig"
    },
...
    {
      "name": "processedtemplates",
      "namespaced": true,
      "kind": "Template"
    },
```

As reference, all the resource currently returned in discovery from k8s core and groups as well as OpenShift can be found [here](https://gist.github.com/moolitayer/c5d831936bef690aa641089e3a55743a)

### openshift_client
Since OpenShift support discovery in the same way, It is possible to create a client for OpenShift using Kubeclient:
```
Kubeclient::Client.new('http://localhost:8443/oapi')
```

So it could be possible in the future to drop openshift_client in favor of kubeclient if that is desirable.

This also brings the possibility of having a forwarding client - just like how oc can work on k8s entities. That would be helpful for some current efforts like generation from templates but is outside the scope of this pr.

### Out of the scope of this pr
- Discovery of nested resources (```namespaces/finalize```)
- Discovery of first level resources (```/healthz```)

This continues the work done in #184 to drop activesupport removing the last method that had lingual logic: pluralize_entities in favor of discovery..

Fixes #144  
Fixes #189

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -12,40 +12,20 @@ module Kubeclient
   # Kubernetes Client
   class Client
     include ClientMixin
-    # Dynamically creating classes definitions (class Pod, class Service, etc.),
-    # The classes are extending RecursiveOpenStruct.
-    # This cancels the need to define the classes
-    # manually on every new entity addition,
-    # and especially since currently the class body is empty
-    ENTITY_TYPES = %w(Pod Service ReplicationController Node Event Endpoint
-                      Namespace Secret ResourceQuota LimitRange PersistentVolume
-                      PersistentVolumeClaim ComponentStatus ServiceAccount).map do |et|
-      clazz = Class.new(RecursiveOpenStruct) do
-        def initialize(hash = nil, args = {})
-          args.merge!(recurse_over_arrays: true)
-          super(hash, args)
-        end
-      end
-      [Kubeclient.const_set(et, clazz), et]
-    end
-
-    ClientMixin.define_entity_methods(ENTITY_TYPES)
-
+    # define a multipurpose resource class, available before discovery
+    ClientMixin.resource_class(Kubeclient, 'Resource')
     def initialize(
       uri,
       version = 'v1',
       **options
     )
       initialize_client(
+        Kubeclient,
         uri,
         '/api',
         version,
         options
       )
-    end
-
-    def all_entities
-      retrieve_all_entities(ENTITY_TYPES)
     end
   end
 end
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -4,6 +4,8 @@ module Kubeclient
   # Common methods
   # this is mixed in by other gems
   module ClientMixin
+    ENTITY_METHODS = %w(get watch delete create update patch)
+
     DEFAULT_SSL_OPTIONS = {
       client_cert: nil,
       client_key:  nil,
@@ -38,8 +40,10 @@ module Kubeclient
     attr_reader :auth_options
     attr_reader :http_proxy_uri
     attr_reader :headers
+    attr_reader :discovered
 
     def initialize_client(
+      class_owner,
       uri,
       path,
       version,
@@ -51,6 +55,9 @@ module Kubeclient
       validate_auth_options(auth_options)
       handle_uri(uri, path)
 
+      @class_owner = class_owner
+      @entities = {}
+      @discovered = false
       @api_version = version
       @headers = {}
       @ssl_options = ssl_options
@@ -66,6 +73,28 @@ module Kubeclient
       end
     end
 
+    def method_missing(method_sym, *args, &block)
+      if discovery_needed?(method_sym)
+        discover
+        send(method_sym, *args, &block)
+      else
+        super
+      end
+    end
+
+    def respond_to_missing?(method_sym, include_private = false)
+      if discovery_needed?(method_sym)
+        discover
+        respond_to?(method_sym, include_private)
+      else
+        super
+      end
+    end
+
+    def discovery_needed?(method_sym)
+      !@discovered && ENTITY_METHODS.any? { |x| method_sym.to_s.start_with?(x) }
+    end
+
     def handle_exception
       yield
     rescue RestClient::Exception => e
@@ -76,6 +105,37 @@ module Kubeclient
       end
       err_message = json_error_msg['message'] || e.message
       raise KubeException.new(e.http_code, err_message, e.response)
+    end
+
+    def discover
+      @entities = {}
+      result = JSON.parse(handle_exception { rest_client.get(@headers) })
+      result['resources'].each do |resource|
+        next if resource['name'].include?('/')
+        entity = ClientMixin.parse_definition(resource['kind'], resource['name'])
+        @entities[entity.method_names[0]] = entity if entity
+      end
+      define_entity_methods
+      @discovered = true
+    end
+
+    def self.parse_definition(kind, name)
+      # "name": "componentstatuses", networkpolicies, endpoints
+      # "kind": "ComponentStatus" NetworkPolicy, Endpoints
+      # maintain pre group api compatibility for endpoints.
+      # See: https://github.com/kubernetes/kubernetes/issues/8115
+      kind = 'Endpoint' if kind == 'Endpoints'
+
+      prefix = kind[0..kind.rindex(/[A-Z]/)] # NetworkP
+      m = name.match(/^#{prefix.downcase}(.*)$/)
+      m && OpenStruct.new(
+        entity_type:   kind, # ComponentStatus
+        resource_name: name, # componentstatuses
+        method_names:  [
+          ClientMixin.underscore_entity(kind),         # component_status
+          ClientMixin.underscore_entity(prefix) + m[1] # component_statuses
+        ]
+      )
     end
 
     def handle_uri(uri, path)
@@ -91,50 +151,58 @@ module Kubeclient
 
     public
 
-    def self.define_entity_methods(entity_types)
-      entity_types.each do |klass, entity_type|
-        entity_name = underscore_entity(entity_type)
-        entity_name_plural = pluralize_entity(entity_name)
+    def self.resource_class(class_owner, entity_type)
+      class_owner.const_get(entity_type, false)
+    rescue NameError
+      class_owner.const_set(
+        entity_type,
+        Class.new(RecursiveOpenStruct) do
+          def initialize(hash = nil, args = {})
+            args[:recurse_over_arrays] = true
+            super(hash, args)
+          end
+        end
+      )
+    end
 
+    def define_entity_methods
+      @entities.values.each do |entity|
+        klass = ClientMixin.resource_class(@class_owner, entity.entity_type)
         # get all entities of a type e.g. get_nodes, get_pods, etc.
-        define_method("get_#{entity_name_plural}") do |options = {}|
-          get_entities(entity_type, klass, options)
+        define_singleton_method("get_#{entity.method_names[1]}") do |options = {}|
+          get_entities(entity.entity_type, klass, entity.resource_name, options)
         end
 
         # watch all entities of a type e.g. watch_nodes, watch_pods, etc.
-        define_method("watch_#{entity_name_plural}") do |options = {}|
+        define_singleton_method("watch_#{entity.method_names[1]}") do |options = {}|
           # This method used to take resource_version as a param, so
           # this conversion is to keep backwards compatibility
           options = { resource_version: options } unless options.is_a?(Hash)
 
-          watch_entities(entity_type, options)
+          watch_entities(entity.resource_name, options)
         end
 
         # get a single entity of a specific type by name
-        define_method("get_#{entity_name}") do |name, namespace = nil|
-          get_entity(entity_type, klass, name, namespace)
+        define_singleton_method("get_#{entity.method_names[0]}") do |name, namespace = nil|
+          get_entity(klass, entity.resource_name, name, namespace)
         end
 
-        define_method("delete_#{entity_name}") do |name, namespace = nil|
-          delete_entity(entity_type, name, namespace)
+        define_singleton_method("delete_#{entity.method_names[0]}") do |name, namespace = nil|
+          delete_entity(entity.resource_name, name, namespace)
         end
 
-        define_method("create_#{entity_name}") do |entity_config|
-          create_entity(entity_type, entity_config, klass)
+        define_singleton_method("create_#{entity.method_names[0]}") do |entity_config|
+          create_entity(entity.entity_type, entity.resource_name, entity_config, klass)
         end
 
-        define_method("update_#{entity_name}") do |entity_config|
-          update_entity(entity_type, entity_config)
+        define_singleton_method("update_#{entity.method_names[0]}") do |entity_config|
+          update_entity(entity.resource_name, entity_config)
         end
 
-        define_method("patch_#{entity_name}") do |name, patch, namespace = nil|
-          patch_entity(entity_type, name, patch, namespace)
+        define_singleton_method("patch_#{entity.method_names[0]}") do |name, patch, namespace = nil|
+          patch_entity(entity.resource_name, name, patch, namespace)
         end
       end
-    end
-
-    def self.pluralize_entity(entity_name)
-      "#{entity_name}#{entity_name.end_with?('s') ? 'es' : 's'}"
     end
 
     def self.underscore_entity(entity_name)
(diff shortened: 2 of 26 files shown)


moolitayer:
@abonas @simon3z please review

simon3z:
@jonmoter @jimmidyson @nyxcharon can you review?

simon3z:
cc @Fryguy

simon3z:
> After a client is created and before it can be used discovery must be made.

@moolitayer can't you just hide that with a lazy initialization?

jonmoter:
We have a lot of logic in our application that does the equivalent of:

```
require 'kubeclient'
service = Kubeclient::Service.new
```

and then we have unit tests that exercise that functionality. I don't want to require a Kubernetes server to be running in order to run my unit tests.

I see that you have some logic to stub out the discovery call and return the contents of `core_api_resource_list.json`.  Could you wrap that logic up in a helper method that could be used when running tests?  Maybe something like:

```
require 'kubeclient'
require 'kubeclient/test'

Kubeclient.stub_core_apis!
# rest of my tests
```

That would make it far easier to cope with the breaking changes this introduces.

/cc @grosser, @irwaters

moolitayer:
> @moolitayer can't you just hide that with a lazy initialization?

now supporting lazy initialization

moolitayer:
@jonmoter could you elaborate on your tests:
are you doing something with the service? e.g 
```
client.create_service(service)
```
if so how are you mocking that?

if not will using 
```
service = Kubeclient::Resource.new
```
solve your problem?

jonmoter:
In our model code, we do stuff like:
```
client.create_service(Kubeclient::Service.new(service_template))
```
and then in our tests, we stub out `create_service` to return some value.

So now that I think about it, we could change our model logic to say `Kubeclient::Resource.new(service_template)`, which would be a minimal change.  That's not too bad.

lib/kubeclient/common.rb
```diff
@@ -51,6 +53,9 @@ def initialize_client(
       validate_auth_options(auth_options)
       handle_uri(uri, path)
 
+      @class_owner = class_owner
+      @entities = nil
```
simon3z: @moolitayer why setting this to `nil` when the first thing you do in `discover` is to set it to an empty hash?

lib/kubeclient/common.rb
```diff
@@ -51,6 +53,9 @@ def initialize_client(
...
+      @discovered = false
```
simon3z: @moolitayer isn't `@discovered` the same as `@entities.empty?`? (provided that you initialize `@entities` to an hash). If you really want to keep it explicit then you can have a method:

```ruby
    def discovered
      @entities.empty?
    end
```

lib/kubeclient/common.rb
moolitayer: The only goal here is to have all instance variables explicitly mentioned at initialization - it serves no functional purpose. 

cc @abonas

simon3z: @moolitayer I wasn't suggesting to remove it, I was suggesting to use the initialization value that you need (empty hash).

lib/kubeclient/common.rb
moolitayer: I noticed in case the returned APIResourceList is empty or has only things that aren't parsed (like nested resources) that could cause an infinite loop in method_missing. 
I would like to have some guard against that

lib/kubeclient/common.rb
moolitayer: That I can do. I would like to keep the same initialization in discover  - so that it purges the list of entities each time it is called.

lib/kubeclient/common.rb
```diff
@@ -78,11 +92,48 @@ def handle_exception
...
+        response = rest_client.get(@headers)
+        result = JSON.parse(response)
+        result['resources'].each do |resource|
+          # mitigation for https://github.com/kubernetes/kubernetes/issues/22413:
+          next if resource['name'].include?('/') || resource['kind'] == 'ReplicationControllerDummy'
+          entity = ClientMixin.parse_definition(resource['kind'], resource['name'])
+          @entities[entity.method_names[0]] = entity if entity
+        end
+        define_entity_methods
+        @discovered = true
```
simon3z: @moolitayer why all these lines are within the `handle_exception` block? Shouldn't just the `rest_client` request be within the `handle_exception` block?

lib/kubeclient/common.rb
```diff
@@ -78,11 +92,48 @@ def handle_exception
...
+      end
+    end
+
+    def self.parse_definition(kind, name)
+      # "name": "componentstatuses", networkpolicies, endpoints
+      # "kind": "ComponentStatus" NetworkPolicy, Endpoints
+      # maintain pre group api compatibility for endpoints.
+      # See: https://github.com/kubernetes/kubernetes/issues/8115
+      kind = 'Endpoint' if kind == 'Endpoints'
```
simon3z: @moolitayer this looks like another exception as the one you have at line 102 (`ReplicationControllerDummy`). I suggest to group these normalizations in a single method you can reuse.

moolitayer: Forgot to remove the exception for ReplicationControllerDummy after implementing a general solution for non verbs.

moolitayer:
Addressed comments @simon3z @abonas PTAL

simon3z:
@jonmoter @jimmidyson @nyxcharon @Fryguy @abonas @zeari @yaacov @cben any comment on this?
We'd like to merge this ASAP (btw we already looked at this in another fashion in #181)

jimmidyson:
Sorry not got time to review right now - I'm sure what you add will be great!

moolitayer:
- Updated README
- Addressed comment about ```@entities``` initialization

lib/kubeclient/common.rb
```diff
@@ -78,11 +92,47 @@ def handle_exception
       raise KubeException.new(e.http_code, err_message, e.response)
     end
 
+    def discover
+      @entities = {}
+      response = handle_exception do
+        rest_client.get(@headers)
+      end
```
simon3z: @moolitayer will a one-liner be nicer here?

```ruby
response = handle_exception { rest_client.get(@headers) }
```

Same for other similar instances in the rest of the PR.

lib/kubeclient/common.rb
```diff
@@ -66,6 +71,15 @@ def initialize_client(
       end
     end
 
+    def method_missing(method_sym, *args, &block)
```
simon3z: @moolitayer I don't think the `super` part is needed. I have the feeling that this logic is simpler and may work as well:

```ruby
def method_missing(method_sym, *args, &block)
  discover unless !@discovered
  send(method_sym, *args, &block)
end
```

lib/kubeclient/common.rb
simon3z: @moolitayer maybe even (if not too clever):

```ruby
result = JSON.parse(handle_exception { rest_client.get(@headers) })
```

lib/kubeclient/common.rb
```diff
@@ -51,6 +53,9 @@ def initialize_client(
       validate_auth_options(auth_options)
       handle_uri(uri, path)
 
+      @class_owner = class_owner
+      @entities = {}
+      @discovered = false
```
simon3z: @moolitayer if you wanted to initialize all the attributes you are missing `@group` I think.

lib/kubeclient/common.rb
moolitayer: That bugged me too.
It is done elsewhere so I submitted:
https://github.com/abonas/kubeclient/pull/186
(For the other one line handler)

and I will fix this one here

lib/kubeclient/common.rb
moolitayer: Causes an infinite loop since method_missing will keep getting called

lib/kubeclient/common.rb
moolitayer: That is done in handle uri (same like @api_endpoint) so I now always init it there.
also renamed group to api_group (like existing api_version) and changed code to rely on initialization.

moolitayer:
@simon3z PTAL

zeari:
@moolitayer I think this looks good but im not too knowledgeable in this code.

enoodle:
:+1: LGTM, I will try to use it now to add deamonset and hopfully I will not have any problems with it.

lib/kubeclient.rb
```diff
@@ -12,40 +12,20 @@ module Kubeclient
...
+    ClientMixin.resource_class(Kubeclient, 'Resource')
     def initialize(
       uri,
       version = 'v1',
       **options
     )
       initialize_client(
+        Kubeclient,
         uri,
         '/api',
```
enoodle: @moolitayer do you think its possible to make this parameter customizable by the user as well? Its needed for example to access beta api: http://kubernetes.io/docs/api/ ( /apis/extentions/<version> )

moolitayer: The user can change this (please give it a try), '/api' is only the default if a user used a uri that includes no path:
```client = Kubeclient::Client.new('http://localhost:8080')```

Comment (enoodle) [inline lib/kubeclient.rb]: :+1: now I see it, Thanks!

File: lib/kubeclient/common.rb [hunk 9]
```diff
@@ -66,6 +71,15 @@ def initialize_client(
       end
     end
 
+    def method_missing(method_sym, *args, &block)
+      if !@discovered && method_sym.to_s.match(/^(get|watch|delete|create|update|patch)_(.*)$/)
```
simon3z: @moolitayer do we need to check for these or is it just an optimization that we could drop? Can we make a constant and simplify the check?

```ruby
ENTITY_METHODS = %w(get watch delete create update patch)

ENTITY_METHODS.any? { |x| method_sym.to_s.start_with?(x) }
```

I am not sure if I like the above better, anyway what I would like to have is a single explicit place where we can add these verbs.

simon3z:
> client = Kubeclient::Client.new('http://localhost:8080/api', 'v1')

@moolitayer we had a way to discover the available versions. Is that still available?

simon3z:
@moolitayer IIRC @zakiva may need a new verb `process`. See #183. Any chance to squeeze that in (another patch but this very PR). That would probably help him.

lib/kubeclient/common.rb
moolitayer: > @moolitayer do we need to check for these or is it just an optimization that we could drop?

Since this is using a basic mechanism of the language I prefer to be as explicit as possible to avoid replacing any unexpected methods. I will extract the list as suggested.

lib/kubeclient/common.rb
```diff
@@ -91,52 +138,60 @@ def build_namespace_prefix(namespace)
...
+          def initialize(hash = nil, args = {})
+            args[:recurse_over_arrays] = true
+            super(hash, args)
+          end
+        end
+      )
+    end
 
+    def define_entity_methods
+      @entities.values.each do |entity|
```
simon3z: @moolitayer once you'll have extracted the verbs (also including the missing `process`), please file an issue (or a trello card) to remember that maybe we can cleanup this method looping over the verbs.

enoodle:
@simon3z I think that `process` only applies for templates so it should be handled separately.

simon3z:
> @simon3z I think that process only applies for templates so it should be handled separately.

@enoodle separate how? code? patch? PR?

enoodle:
@simon3z I thought it would be defined in openshift_client as a sole method `process_template` that will handle this specific case.

lib/kubeclient/common.rb
moolitayer: you mean with the list of verbs? each method has different parameters set so I'm not sure how that would work

moolitayer:
> @moolitayer IIRC @zakiva may need a new verb process. See #183. Any chance to squeeze that in (another patch but this very PR). That would probably help him.

Any reason not to keep it in a separate patch?

simon3z:
> > @moolitayer IIRC @zakiva may need a new verb process. See #183. Any chance to squeeze that in (another patch but this very PR). That would probably help him.
>
> Any reason not to keep it in a separate patch?

@moolitayer, the argument was that `process` for now is just an OpenShift verb. Although there's a proposal here for Kubernetes https://github.com/kubernetes/kubernetes/pull/25293

@enoodle @moolitayer if this PR is covering OpenShift as a group API I feel that having another gem just for the `process` verb is overkill, and anyway the Kubernetes PR will be merged at some point.

simon3z:
LGTM :+1:
@abonas any opinion on this PR?

cc @jonmoter @nyxcharon @Fryguy @zeari @yaacov @cben @grosser @irwaters

lib/kubeclient/common.rb
Fryguy: Always define `respond_to_missing?` when implementing `method_missing`... https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding

moolitayer:
Addressed with the best solution I found. Added a test test_missing_method.rb because this is quite hairy. 
- I've reset the client in the test since it depends on state and I noticed some of the implementations I tested were wrong only on the first call.
- Currently respond_to? will raise an exception if the discovery fails. We could also return false. I'm not sure what is more deceiving for a user...

@Fryguy thanks, please take another look.

Fryguy:
@moolitayer The respond_to_missing? changes look great to me :+1:

zeari:
:hooray:

zeari:
:confetti_ball:

PR #191, opened 2016-09-19
Absolute paths for certs

albertrdixon:
Really simple change here to allow for absolute paths to cert files in your kubeconfig.

### DIFF
--- a/lib/kubeclient/config.rb
+++ b/lib/kubeclient/config.rb
@@ -1,5 +1,6 @@
 require 'yaml'
 require 'base64'
+require 'pathname'
 
 module Kubeclient
   # Kubernetes client configuration class
@@ -63,7 +64,7 @@ module Kubeclient
     private
 
     def ext_file_path(path)
-      File.join(@kcfg_path, path)
+      Pathname(path).absolute? ? path : File.join(@kcfg_path, path)
     end
 
     def fetch_context(context_name)


simon3z:
@albertrdixon I think that there is something more official than `start_with?('/')` to check if a path is absolute. For example: [this](https://ruby-doc.org/stdlib-1.9.3/libdoc/pathname/rdoc/Pathname.html#method-i-absolute-3F). Also, I have the feeling there was even something simpler to use (can you check?).

albertrdixon:
Fair enough. I was looking for something simple without needed a new object, but I suppose `Pathname` is the way to go.

simon3z:
@moolitayer @enoodle can you review?

moolitayer:
@albertrdixon thanks for the patch. 
Can you please provide a code example of what this is fixing.

albertrdixon:
Sure,

let's say you put all your certs in `/etc/ssl/certs/` then in your kubeconfig you might have `certificate-authority: /etc/ssl/certs/ca.pem`. This patch will fix kubeclient incorrectly trying to find `ca.pem` relative to the kubeconfig.

simon3z:
LGTM :+1: can anyone else review/ack as well? (cc @abonas @moolitayer @enoodle )

simon3z:
@albertrdixon I don't recall, do we have tests for this part? Should we add a test for your new code as well?

lib/kubeclient/config.rb
```diff
@@ -63,6 +64,7 @@ def context(context_name = nil)
     private
 
     def ext_file_path(path)
+      return path if Pathname(path).absolute?
```
enoodle: Question: shouldn't this be wrapped with `File` as well? (`return File(path) if Pathname(path).absolute?`

moolitayer: `File.join` below returns a string and that's what we need here as well.

moolitayer:
@albertrdixon thanks. Please squash your two commit;
When someone looks at the log in the future they should see this as one change which this is.

Once you do that I'm good with this change

albertrdixon:
@simon3z no explicit tests for absolute paths, and I'm not sure of an elegant way to add that in. My thinking is that it is sufficient to ensure this does not affect current functionality (relative paths) and - since this is now using `Pathname#absolute?` - assume the `Pathname` will Do The Right Thing.

PR #192, opened 2016-09-23
Bump version to 2.0.0

simon3z:
cc @abonas @jonmoter @zeari @yaacov @cben @grosser @zakiva @moolitayer

### DIFF
--- a/lib/kubeclient/version.rb
+++ b/lib/kubeclient/version.rb
@@ -1,4 +1,4 @@
 # Kubernetes REST-API Client
 module Kubeclient
-  VERSION = '1.2.0'
+  VERSION = '2.0.0'
 end


zakiva:
I think we are still missing the option to set the target API of the client, @moolitayer have you added it already?

moolitayer:
> I think we are still missing the option to set the target API of the client, @moolitayer have you added it already?

The idea was instead of having to recreate the client to change groups to have a method that would
set group & version. That would have to reset discovery & delete defined methods.

cc @zeari @simon3z

zeari:
ping @simon3z

simon3z:
> ping @simon3z

@zeari I discussed this with @moolitayer and @zakiva this morning. Waiting here for either +1's or other PRs before proceeding.

moolitayer:
LGTM :+1: 
Discussed with @zakiva how ManageIQ code should look

simon3z:
Released: https://rubygems.org/gems/kubeclient/versions/2.0.0

PR #204, opened 2016-10-26
bump http lib to remove circular dependencies

grosser:
trying to avoid circular dependency in http library see https://github.com/httprb/http/pull/353

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -28,5 +28,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rubocop', '= 0.47.1'
   spec.add_dependency 'rest-client'
   spec.add_dependency 'recursive-open-struct', '~> 1.0.0'
-  spec.add_dependency 'http', '= 0.9.8'
+  spec.add_dependency 'http', '~> 2.0'
 end


simon3z:
cc @moolitayer

grosser:
same tests are red on master ... so good to go ?

moolitayer:
Thanks @grosser.
Do you have a reproduction of the issue solved here?
I'm generally ok with this fix. @simon3z what do you think?
Anything in particular we need to test for bumping 'http' from 0.9.8 to 2.0.3?

It's the first time i'm seeing this integration failure and I couldn't figure out why it suddenly
happens(gem change?) @simon3z @grosser @jonmoter have you any idea why this happens?

grosser:
reproduction is 

```
ruby -w -r bundler/setup -r http -e 1
```

grosser:
error is

```
assert(context.ssl_options[:cert_store].verify(context.ssl_options[:client_cert]))
```

tried adding timecop because maybe the cert is expired, but that did not work ...

@simon3z can you check the certs are indeed expired ... then update the certs and ideally document how they were created ?

grosser:
alternatively can we set these tests to skip and move on instead of blocking all PRs on it ?

simon3z:
@grosser we'll try to fix this asap.

simon3z:
@grosser I think that you should update this PR according to the agreement in #206

grosser:
agreement was to merge this when next kubeclient major release is done ?

grosser:
alternatively we could say "~> 0.9.8", "~> 2.0.3" ... but that would be lying since we don't test on both versions

simon3z:
> agreement was to merge this when next kubeclient major release is done ?

@grosser I think it was mentioned on #206 to change this to `~> 2.0`, but probably leaving it to `~> 2.0.3` is the same.

I think we could try to merge this ASAP. Is that ok with you @moleksyuk ?

moleksyuk:
Hi @simon3z 
Sorry for a long silence.

Unfortunately it won't work for me, because other third party dependencies required other newer version. It was 2 weeks ago, I upgraded manually to http library in logstash but got other issues.

Also, it's not required for me any more.

But thanks a lot guys!

grosser:
updated to 2.0 ... which is the correct dependency in semver terms ...

simon3z:
@gabe-ochoa is this change OK for you?

gabe-ochoa:
@simon3z Apologies for the _super_ slow reply. This change works for my use case. We modified the usage of the tool chain that we're using `kubeclient` in such that it's isolated enough to not cause the dependency errors we were getting. 

Great to see the dependency being upgraded.

cben:
@msufa Is this OK or would it break your Celluloid + http.rb use case?

msufa:
@cben the use-case was meant for https://github.com/zendesk/samson but has essentially been removed in https://github.com/zendesk/samson/pull/1085

Since I'm no longer involved in that project I'll let @grosser have the final word, though I think this should work fine.

grosser:
updated ... ready to merge ?

cben:
Cool LGTM.

@grosser kubeclient README documents support for celluloid-io, do you think that needs any update?
Reading a bit more it seems http might still support Celluloid, but not officially due to not supporting timeouts [https://github.com/httprb/http/pull/268], but kubeclient uses http with nil (infinite) timeouts, so it might still work...

grosser:
I'd just leave it there until someone complains about what exactly does not
work ...

simon3z:
@cben @moolitayer can you review/approve this?

cben:
We labeled v2.x/no as 0.9.8 is still bundle-installable and aside from runtime warnings passes kubeclient specs, so no need to *force* this bump on people in a minor release.
- [ ] @moolitayer but we should relax dependency on 2.x to *allow* people to get newer version without the runtime warnings.

moolitayer:
@cben so we will need a 2.x only PR, added to our list

PR #205, opened 2016-10-27
make test output more readable by adding color

grosser:


kubeclient.gemspec
```diff
@@ -22,6 +22,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler', '~> 1.6'
   spec.add_development_dependency 'rake', '~> 10.0'
   spec.add_development_dependency 'minitest'
+  spec.add_development_dependency 'minitest-rg'
```
simon3z: @grosser do you need both `minitest` and `minitest-rg`... or the latter is a replacement and you can drop the former?

cben:
LGTM, tested :green_heart:.

PR #209, opened 2016-11-07
Fix undefined method `rindex'

zakiva:
A possible fix for: BZ https://bugzilla.redhat.com/show_bug.cgi?id=1387614

The fix handles a 'nil' `kind`. However, we would expect that each resource would have a well defined `kind`, so I'm not sure if this is the right approach here, asked for more information in the BZ. 

@simon3z @moolitayer @zeari @cben

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -111,7 +111,7 @@ module Kubeclient
       @entities = {}
       result = JSON.parse(handle_exception { rest_client.get(@headers) })
       result['resources'].each do |resource|
-        next if resource['name'].include?('/')
+        next if resource['name'].include?('/') || resource['kind'].nil?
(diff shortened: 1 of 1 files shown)


simon3z:
@moolitayer please review

moolitayer:
If k8s returns something like a missing kind[1] we might get a NoMethodError instead of the current error described in the bz with this fix. I guess this is the best we can do.

Please move the fix to the discover method in the ```next if resource['name'].include?('/')``` line

[1] example: 
```
{
  "name": "bindings",
  "namespaced": true
},
```

zakiva:
@moolitayer done, please take another look

moolitayer:
LGTM :+1: cc @simon3z note might not actually solve the bz[1]

[1] there could be something going wrong (e.g a resource like events with no kind field in openshift 3.1) and we ignore that only to get a NoMethodError later.

PR #211, opened 2016-11-14
Enable "gcp" authentication structure for bearer token in .kube/config

jeremywadsack:
This adds the ability to parse the bearer token from ~/.kube/config
stored by Google Cloud Platform. This token is only good for an hour
since it was last renewed with a `kubectl` command. So if a 401 error
is returned adds a note to the exception message when the token is
expired.

Also updates other auth failure tests that were bypassing exception
parsing in `Kubeclient::ClientMixing#handle_exception`.

Addresses #210

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -104,6 +104,9 @@ module Kubeclient
         {}
       end
       err_message = json_error_msg['message'] || e.message
+      if e.http_code == 401 && @auth_options.key?(:bearer_token_expiry) && @auth_options[:bearer_token_expiry] < Time.now
+        err_message += '. Your token has expired. If you are using `kubectl`, running any command should renew your token.'
+      end
       raise KubeException.new(e.http_code, err_message, e.response)
     end
 
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -104,7 +104,7 @@ module Kubeclient
         {}
       end
 
-      err_message = expired_token_message(
+      err_message = check_token_expiration(
         e.http_code,
         json_error_msg['message'] || e.message
       )
@@ -483,7 +483,7 @@ module Kubeclient
       options.merge(@socket_options)
     end
 
-    def expired_token_message(http_code, message)
+    def check_token_expiration(http_code, message)
       if http_code == 401 &&
          @auth_options.key?(:bearer_token_expiry) &&
          @auth_options[:bearer_token_expiry] < Time.now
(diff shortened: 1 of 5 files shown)


simon3z:
@moolitayer @enoodle can you review?

lib/kubeclient/common.rb
```diff
@@ -477,5 +482,16 @@ def http_options(uri)
 
       options.merge(@socket_options)
     end
+
+    def expired_token_message(http_code, message)
```
moolitayer: What is the error returned from the server in this case? I would prefer to keep mirroring the returned exception.

enoodle: Consider changing the name of the method to something like `check_token_expiration` as it adds to the mesasge.

jeremywadsack: @moolitayer intended to include the original message. That's why it passes in `message` and carries that through. Maybe I misunderstood your comment.

Another option would be to have a `TokenExpiredException` (or something) that takes the original exception as an argument and adds the expiration message. I thought this was a simpler approach.

jeremywadsack: Good suggestion. I wasn't happy with the original name.

lib/kubeclient/common.rb
```diff
@@ -477,5 +482,16 @@ def http_options(uri)
 
       options.merge(@socket_options)
     end
+
+    def check_token_expiration(http_code, message)
+      if http_code == 401 &&
+         @auth_options.key?(:bearer_token_expiry) &&
+         @auth_options[:bearer_token_expiry] < Time.now
+        return message + '. Your token has expired. If you are using `kubectl`, running any '\
+          'command should renew your token.'
```
moolitayer: Maybe I don't understand this but it does not seem like a flow that a normal Kubeclient user would use.
(use library to talk to api while using command line tool to get authentication token). Can we do the same thing kubectl does to renew the token? what is needed for the kubectl to give back a token (meaning do you have to be pre authenticated against www.googleapis.com/auth/cloud-platform ?) 

This seems relevant: https://github.com/kubernetes/kubernetes/blob/3e51ca2e16479763bc832e893e1b9ffa34310b7e/plugin/pkg/client/auth/gcp/gcp.go#L56

jeremywadsack: I opened #213 which provides support for using Google's default application credentials. I think this is a better solution as long as you are fine with a dependency on `googleauth` gem.

jeremywadsack:
Closing this as I don't think it's a reasonable solution. See #213 for a better approach to handling GCP authorizations.

PR #214, opened 2016-11-29
Add backward compatibility for missing kind

zakiva:
Adding static mapping for resources in order to keep compatibility for Openshift 3.1 (kubernetes v1.1.0-origin). 

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1387614

@simon3z @moolitayer @cben @zeari

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,9 @@ You can also load your JSONified config in from an ENV variable (e.g. `KUBE_CONF
 ```
 Kubeclient::Config.new(JSON.parse(ENV['KUBE_CONFIG']), nil)
 ```
+###Supported kubernetes versions
+
+For 1.1 only the core api v1 is supported, all api groups are supported in later versions.
 
 ## Examples:
 
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -425,7 +425,9 @@ module Kubeclient
     def load_entities
       @entities = {}
       fetch_entities['resources'].each do |resource|
-        next if resource['name'].include?('/') || resource['kind'].nil?
+        next if resource['name'].include?('/')
+        resource['kind'] = Kubeclient::Common::MissingKindCompatibility
+                           .resource_kind(resource['name']) if resource['kind'].nil?
         entity = ClientMixin.parse_definition(resource['kind'], resource['name'])
         @entities[entity.method_names[0]] = entity if entity
       end
--- a/test/test_resource_list_without_kind.rb
+++ b/test/test_resource_list_without_kind.rb
@@ -1,0 +1,78 @@
+require 'test_helper'
+
+# Core api resource list without kind tests
+class TestResourceListWithoutKind < MiniTest::Test
+  def test_get_from_json_api_v1
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list_without_kind.json'),
+                 status: 200)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+    client.discover
+
+    [
+      {
+        entity: 'pod',
+        type: 'Pod',
+        name: 'pods',
+        methods: %w(pod pods)
+      },
+      {
+        entity: 'node',
+        type: 'Node',
+        name: 'nodes',
+        methods: %w(node nodes)
+      },
+      {
+        entity: 'service',
+        type: 'Service',
+        name: 'services',
+        methods: %w(service services)
+      }
+    ].each { |h| assert_entities(client.instance_variable_get(:@entities)[h[:entity]], h) }
+
+    assert_requested(:get,
+                     'http://localhost:8080/api/v1',
+                     times: 1)
+  end
+
+  def test_get_from_json_oapi_v1
+    stub_request(:get, %r{/oapi/v1$})
+      .to_return(body: open_test_file('core_oapi_resource_list_without_kind.json'),
+                 status: 200)
+
+    client = Kubeclient::Client.new('http://localhost:8080/oapi/', 'v1')
+    client.discover
+
+    [
+      {
+        entity: 'template',
+        type: 'Template',
+        name: 'templates',
+        methods: %w(template templates)
+      },
+      {
+        entity: 'build',
+        type: 'Build',
+        name: 'builds',
+        methods: %w(build builds)
+      },
+      {
+        entity: 'project',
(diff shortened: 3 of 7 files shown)


lib/kubeclient/common.rb
```diff
@@ -425,6 +425,7 @@ def api
     def load_entities
       @entities = {}
       fetch_entities['resources'].each do |resource|
+        resource = Kubeclient::Common::Compatibility31.compatible_resource(resource)
```
simon3z: @zakiva shouldn't you do this only if `resource['kind'].nil?`

simon3z:
@zakiva please link this on the BZ as well

lib/kubeclient/compatibility_3_1.rb
```diff
@@ -0,0 +1,68 @@
+module Kubeclient
+  module Common
+    # Backward compatibility for 3.1
+    class Compatibility31
```
zakiva: not sure regarding the naming here (and the file name)

lib/kubeclient/common.rb
zakiva: Yes, but I checked it inside the method: https://github.com/abonas/kubeclient/pull/214/files#diff-f1f5af126b7e6ff600fcc39decd74abfR64 thought it would be nicer to hide it. 
Should I check it here instead?

moolitayer:
@zakiva I asked to add a note in the readme about supported kubernetes versions.
Mention for 1.1 we support only core api not apis/extensions. Thanks!
(I think that is the correct k8s version, 3.1 is Openshift)

ps We need to make sure future manageiq code is aware of that, and uses other apis only if available.

lib/kubeclient/common.rb
simon3z: @zakiva it's late here so don't assume that this works (or makes sense :blush:)... but I would have expected something as:

```ruby
fetch_entities['resources'].each do |resource|
  next if resource['name'].include?('/')
  resource['kind'] = Compatibility31.resource_kind(resource['name']) if resource['kind'].nil?
...
```
Instead of `nil?` you can also think to use `empty?` (I always forget the nits... anyway it would be nice to catch also when it's an empty string).

@moolitayer please check if the above makes sense.
For example, the check `next if resource['name'].include?('/')` is fully valid and working on 3.1 as well?

moolitayer: Yeah I'm good with that (.nil? since empty comes from rails) I'd definitely like to see a backward compatibility test.
@zakiva take a look at test/json/core_api_resource_list.json and it's use all over. You would probably want to rename it to core_api_resource_list_v13.json and add a core_api_resource_list_v11.json that you can test in a new file.

README.md
```diff
@@ -198,6 +198,9 @@ You can also load your JSONified config in from an ENV variable (e.g. `KUBE_CONF
 ```
 Kubeclient::Config.new(JSON.parse(ENV['KUBE_CONFIG']), nil)
 ```
+###Supported kubernetes versions
+
+For 1.1 only the core api is supported, apis/extensions are supported in later versions.
```
cben: We could in principle extend the mapping, right?  It's just we don't yet need them?

Saying 1.1 is not precise, when I tried Origin 1.1 I did get `kind` fields.  See my comment on naming.

lib/kubeclient/compatibility_3_1.rb
zakiva: @moolitayer @simon3z I would like to go with @cben's suggestion `MissingKindCompatibility` since we are not sure yet that kind is missing in k8s 1.1, are you OK with that?

README.md
moolitayer: It should be in the mapping AFAIK. Since we have to manually test, review and add tests for everything we support I would rather we scope this change only for core api|oapi.

We would also need to check if we have cases of different apis using same entities.

@zakiva 
- we are supporting v1 only right? please state that here
- can you figure out with @cben which is the currect k8s version to use here?

lib/kubeclient/compatibility_3_1.rb
moolitayer: yes

README.md
zakiva: added 'v1'. @cben the point is that we know that for OSE 3.1 ( - kubernetes v1.1.0-origin ) kind is missing, so anyway we can't guarantee support for 1.1 and that's what I have mentioned in the readme.

zakiva:
@moolitayer added tests, please take another look.

cben:
LGTM

README.md
```diff
@@ -198,6 +198,9 @@ You can also load your JSONified config in from an ENV variable (e.g. `KUBE_CONF
 ```
 Kubeclient::Config.new(JSON.parse(ENV['KUBE_CONFIG']), nil)
 ```
+###Supported kubernetes versions
+
+For 1.1 only the core api v1 is supported, apis/extensions are supported in later versions.
```
moolitayer: apis/extensions is only one group api(the only one in 1.1 but not in later versions) please change to 
`For 1.1 only the core api v1 is supported, all api groups are supported in later versions.`

test/test_resource_list_without_kind.rb
```diff
@@ -0,0 +1,78 @@
+require 'test_helper'
+
+# Core api resource list without kind tests
+class TestResourceListWithoutKind < MiniTest::Test
+  def test_get_from_json_api_v1
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list_without_kind.json'),
+                 status: 200)
+
+    client = Kubeclient::Client.new 'http://localhost:8080/api/', 'v1'
```
moolitayer: parentheses please

test/test_resource_list_without_kind.rb
```diff
@@ -0,0 +1,78 @@
...
+                     'http://localhost:8080/api/v1',
+                     times: 1)
+  end
+
+  def test_get_from_json_oapi_v1
+    stub_request(:get, %r{/oapi/v1$})
+      .to_return(body: open_test_file('core_oapi_resource_list_without_kind.json'),
+                 status: 200)
+
+    client = Kubeclient::Client.new 'http://localhost:8080/oapi/', 'v1'
```
moolitayer: parentheses please

moolitayer:
LGTM :+1: cc @simon3z

simon3z:
Nice! :+1:

PR #221, opened 2017-01-09
Include request info in exceptions to make it easy to see what reques…

grosser:
…t and what server failed

we are doing a lot of requests to a lot of different servers, so a general "This failed" exception is not helpful.

```
Kubernetes::Cluster.last.client.get_secret 'sdfsdfs'
KubeException: HTTP status code 404 the server could not find the requested resource for GET https://k8s-sandbox.consul:8443/api/v1/secrets/sdfsdfs
  from kubeclient/common.rb:107:in `rescue in handle_exception'
```

@abonas

### DIFF
--- a/lib/kubeclient/kube_exception.rb
+++ b/lib/kubeclient/kube_exception.rb
@@ -9,6 +9,10 @@ class KubeException < StandardError
   end
 
   def to_s
-    'HTTP status code ' + @error_code.to_s + ', ' + @message
+    string = "HTTP status code #{@error_code}, #{@message}"
+    if @response.is_a?(RestClient::Response) && @response.request
+      string << " for #{@response.request.method.upcase} #{@response.request.url}"
+    end
+    string
   end
 end
--- a/test/test_kubeclient.rb
+++ b/test/test_kubeclient.rb
@@ -79,6 +79,8 @@ class KubeClientTest < MiniTest::Test
     assert_instance_of(KubeException, exception)
     assert_equal("converting  to : type names don't match (Pod, Namespace)",
                  exception.message)
+
+    assert_includes(exception.to_s, ' for POST http://localhost:8080/api')
     assert_equal(409, exception.error_code)
   end
 


lib/kubeclient/common.rb
```diff
@@ -104,7 +104,7 @@ def handle_exception
         {}
       end
       err_message = json_error_msg['message'] || e.message
-      raise KubeException.new(e.http_code, err_message, e.response)
+      fail KubeException.new(e.http_code, err_message, e.response)
```
grosser: making this consistent with the other usages

lib/kubeclient/kube_exception.rb
```diff
@@ -9,6 +9,10 @@ def initialize(error_code, message, response)
   end
 
   def to_s
```
grosser: ideally I'd like to modify the `@message` in `initialize`, but that would be a bigger change ...

simon3z:
@grosser just a couple of thoughts (to be discussed): could this additional information being reported in one (or more) separate attributes of KubeException? (For example it may be nice to keep the URL as an URL instead of stringify it).
Can we report all this info only when specifically requested? E.g. `ex.verbose_message`? (Naming is probably not good but you get the idea...)

@moolitayer @dkorn @enoodle @zeari please review.

grosser:
I'd prefer more output by default ... it's easy to ignore, but hard to get
when something goes wrong and was not set up to be verbose
separate attributes are not really necessary since all that information is
already available via the request

lib/kubeclient/kube_exception.rb
```diff
@@ -9,6 +9,10 @@ def initialize(error_code, message, response)
...
-    'HTTP status code ' + @error_code.to_s + ', ' + @message
+    string = "HTTP status code #{@error_code} #{@message}"
```
moolitayer: Removed the ',' on purpose?

lib/kubeclient/common.rb
```diff
@@ -265,7 +265,7 @@ def get_entities(entity_type, klass, resource_name, options = {})
       resource_version = result.fetch('resourceVersion', nil)
       if resource_version.nil?
         resource_version =
-            result.fetch('metadata', {}).fetch('resourceVersion', nil)
+          result.fetch('metadata', {}).fetch('resourceVersion', nil)
```
simon3z: @grosser fixing the indentation style (here and below) is not related to this PR. Please send a separate PR (which I suppose will be merged much faster).

grosser:
updated

zeari:
LGTM @simon3z

lib/kubeclient/kube_exception.rb
```diff
@@ -9,6 +9,10 @@ def initialize(error_code, message, response)
   end
 
   def to_s
-    'HTTP status code ' + @error_code.to_s + ', ' + @message
+    string = "HTTP status code #{@error_code}, #{@message}"
+    if @response.is_a?(RestClient::Response) && @response.request
+      string << " for #{@response.request.method.upcase} #{@response.request.url}"
```
simon3z: @grosser should we protect in case any of these are ever `nil`?

This is a pretty dangerous part where to get `NoMethodError` exceptions. cc @cben

grosser: checking that the request is there above ... and method can never be nil ... so I think this is safe

cben: LGTM.  RestClient::Request [normalizes method][1] to always be a string.

[1]: https://github.com/rest-client/rest-client/blob/f62a8a0e/lib/restclient/request.rb#L707-L710

PR #223, opened 2017-01-10
bump rubocop, fix offenses, use 2-space indent

grosser:
@abonas 

fixing weird indents with old rubocop version did not work ... so bumping ...
found a few new things (frozen constants / weird indents / always use raise)

to me these all look more readable now ... let me know which ones you don't like and I'll change them

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -4,7 +4,7 @@ module Kubeclient
   # Common methods
   # this is mixed in by other gems
   module ClientMixin
-    ENTITY_METHODS = %w(get watch delete create update patch)
+    ENTITY_METHODS = %w(get watch delete create update patch).freeze
 
     DEFAULT_SSL_OPTIONS = {
       client_cert: nil,
@@ -133,7 +133,7 @@ module Kubeclient
     end
 
     def handle_uri(uri, path)
-      fail ArgumentError, 'Missing uri' unless uri
+      raise ArgumentError, 'Missing uri' unless uri
       @api_endpoint = (uri.is_a?(URI) ? uri : URI.parse(uri))
       @api_endpoint.path = path if @api_endpoint.path.empty?
       @api_endpoint.path = @api_endpoint.path.chop if @api_endpoint.path.end_with? '/'
@@ -257,16 +257,15 @@ module Kubeclient
       ns_prefix = build_namespace_prefix(options[:namespace])
       response = handle_exception do
         rest_client[ns_prefix + resource_name]
-        .get({ 'params' => params }.merge(@headers))
+          .get({ 'params' => params }.merge(@headers))
       end
 
       result = JSON.parse(response)
 
-      resource_version = result.fetch('resourceVersion', nil)
-      if resource_version.nil?
-        resource_version =
-            result.fetch('metadata', {}).fetch('resourceVersion', nil)
-      end
+      resource_version =
+        result.fetch('resourceVersion') do
+          result.fetch('metadata', {}).fetch('resourceVersion', nil)
+        end
 
       # result['items'] might be nil due to https://github.com/kubernetes/kubernetes/issues/13096
       collection = result['items'].to_a.map { |item| new_entity(item, klass) }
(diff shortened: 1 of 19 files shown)


simon3z:
@cben can you review this? Thanks.

lib/kubeclient/common.rb
```diff
@@ -426,8 +422,8 @@ def load_entities
       @entities = {}
       fetch_entities['resources'].each do |resource|
         next if resource['name'].include?('/')
-        resource['kind'] = Kubeclient::Common::MissingKindCompatibility
-                           .resource_kind(resource['name']) if resource['kind'].nil?
+        resource['kind'] ||=
```
cben: This will also use the falback mapping if it's `false`.  Should never be a boolean, either a string or absent.  And if it could be `false`, using mapping is better.  => OK

lib/kubeclient/common.rb
```diff
@@ -257,16 +257,15 @@ def get_entities(entity_type, klass, resource_name, options = {})
...
 
-      resource_version = result.fetch('resourceVersion', nil)
-      if resource_version.nil?
-        resource_version =
-            result.fetch('metadata', {}).fetch('resourceVersion', nil)
-      end
+      resource_version =
+        result.fetch('resourceVersion') do
+          result.fetch('metadata', {}).fetch('resourceVersion', nil)
+        end
```
simon3z: @cben can you review the change above for correctness. Thanks.

cben: There is a mild difference:
given `result == {'resourceVersion' => nil, 'metadata' => {'resourceVersion' => 10}}`,
old code would take the 10, while new code returns nil — `fetch` runs its block only when the key is entirely absent.

Don't know if this matters (can both resourceVersion and metadata.resource be present but one be nil?), and don't know which behavior is better.
Anyway I think this would match old logic exactly:
```
resource_version = result.fetch('resourceVersion', nil)
resource_version ||= result.fetch('metadata', {}).fetch('resourceVersion', nil)
```
[Well not *exactly* exactly :-)  This still would still differ in handing of `false`, but it sounds safe to assume it's not a boolean.]

test/test_kubeclient.rb
```diff
@@ -368,17 +340,19 @@ def test_api_bearer_token_with_params_success
 
   def test_api_bearer_token_success
     stub_request(:get, %r{/api/v1$})
-      .to_return(body: open_test_file('core_api_resource_list.json'),
-                 status: 200)
+      .to_return(
+        body: open_test_file('core_api_resource_list.json'), status: 200
+      )
```
cben: I must say I find the previous style in this particular case more readable.
2 lines became 3 but the actual arguments got crammed on one line.

IMHO the natural progression as lines get too long and require splitting is from:
```
func(arg1, arg2)
```
to
```
func(long_argument1,
     long_argument2)
```
to
```
a_long.function_call(
  long_argument1,
  long_argument2
)
```

IIUC many of your changes are forced by Rubocop forbidding the 2nd option, so you compacted to 1st or expanded to 3rd, right?
I don't feel strongly about this (and I like most of your changes!) but I wonder if *enforcing* this is beneficial.  Is there a Rubocop setting that would make it accept the 2nd form, leaving it to human discretion?

test/test_kubeclient.rb
```diff
@@ -459,107 +428,100 @@ def test_api_basic_auth_failure
...
+        auth_options: { user: 'username' }
+      )
     end
     assert_equal(expected_msg, exception.message)
   end
 
   def test_init_username_and_bearer_token
     expected_msg = 'Invalid auth options: specify only one of username/password,' \
-                   ' bearer_token or bearer_token_file'
+      ' bearer_token or bearer_token_file'
```
cben: consider dropping the start of the string to a new line (as you did elsewhere), to keep alignment.

cben:
LGTM.  Thanks for the big cleanup!

simon3z:
@cben can you please double-check that any extra change (other than indentation) is correct? (E.g. as you did for the one about `result.fetch('resourceVersion') do`).

If everything is alright then we'll merge. Thanks.

cben:
Already did.  I saw only 2 non-formatting changes, and I think they're both
OK.
(unless you think we might both resourceVersion and
metadata.resourceVersion simultaneously,
with one being nil?)

PR #225, opened 2017-02-07
Remove limited entity list from Readme

moolitayer:
We are specifying support of a limited list of objects in the Readme whilst supporting all of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,9 @@
 [![Dependency Status](https://gemnasium.com/abonas/kubeclient.svg)](https://gemnasium.com/abonas/kubeclient)
 
 A Ruby client for Kubernetes REST api.
-The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas, limit ranges, endpoints, persistent volumes, persistent volume claims, component statuses and service accounts.
+The client supports GET, POST, PUT, DELETE on all the entities available in kubernetes in both the core and group apis.
 The client currently supports Kubernetes REST api version v1.
+To learn more about groups and versions in kubernetes refer to [k8s docs](https://kubernetes.io/docs/api/)
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -262,7 +262,7 @@ Input parameter - object of type `Service`, `Pod`, `ReplicationController`.
 The below example is for v1
 
 ```ruby
-service = Service.new
+service = Kubeclient::Resource.new
 service.metadata = {}
 service.metadata.name = "redis-master"
 service.metadata.namespace = 'staging'
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ client = Kubeclient::Client.new 'https://localhost:8443/api/' , 'v1',
                                 auth_options: auth_options
 ```
 
-You can find information about token in [this guide](http://kubernetes.io/docs/user-guide/accessing-the-cluster/) and in [this reference](http://kubernetes.io/docs/admin/authentication/).
+You can find information about tokens in [this guide](http://kubernetes.io/docs/user-guide/accessing-the-cluster/) and in [this reference](http://kubernetes.io/docs/admin/authentication/).
 
 You can also use kubeclient with non-blocking sockets such as Celluloid::IO, see [here](https://github.com/httprb/http/wiki/Parallel-requests-with-Celluloid%3A%3AIO) 
 for details. For example:


moolitayer:
@hridyeshpant @jonmoter could you please review the Readme change?

hridyeshpant:
@moolitayer looks good.
also does below lines are correct to create Service object? 
**The below example is for v1
service = Service.new**

but when i looked the test code, we are using something like below?
our_service = Kubeclient::Resource.new
    our_service.metadata = {}
    our_service.metadata.name = 'guestbook'
    our_service.metadata.namespace = 'staging'
    our_service.metadata.labels = {}
    our_service.metadata.labels.name = 'guestbook'

simon3z:
@hridyeshpant @jonmoter can you review? Thanks.

hridyeshpant:
@simon3z the README is still not clear about creating the service,deployment or others.
if i used Kubeclient::Service.new(), i am getting error
uninitialized constant Kubeclient::Service (NameError)

if i used service = Service.new as per README , this definitely not work

i have my own deployment json/ service.json file i want to create deployment and service based on these files.
but i am not sure how to pass these json content to create object Kubeclient::Service.new() or Kubeclient::Deployment.new().
i want to avoid creating data like below, rather then want to pass json data to Service or Deployment .
service.metadata = {}
service.metadata.name = "redis-master"
service.metadata.namespace = 'staging'
service.spec = {}
service.spec.ports = [{ 'port' => 6379,
                                'targetPort' => 'redis-server'
                              }]
service.spec.selector = {}
service.spec.selector.name = "redis"
service.spec.selector.role = "master"
service.metadata.labels = {}
service.metadata.labels.app = 'redis'
service.metadata.labels.role = 'slave 

i want somethings 
@k8_group_client.create_deployment(json), where our_service is my currently json file data.
I am not sure if this is possible.
i tried to use  deployment = Kubeclient::Resource.new(json)
and use @k8_group_client.create_deployment(deployment). i am getting 
/Users/.rvm/gems/ruby-2.3.1/gems/kubeclient-2.3.0/lib/kubeclient/common.rb:107:in `rescue in handle_exception': the server does not allow this method on the requested resource (KubeException)

is there way i can create service and deployment from json file.

simon3z:
> @k8_group_client.create_deployment(json), where parameter is json file data.
> I am not sure if this is possible.

@hridyeshpant just load the file regularly:

```ruby
service_json = JSON.parse(File.read('service.json'))
@k8_group_client.create_deployment(service_json)
```

hridyeshpant:
@simon3z yeah i tried but it is not working.
/Users/.rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/json/common.rb:156:in `initialize': no implicit conversion of Hash into String (TypeError)

and when i tried  with below is my code, i am able to create service but not deployment 
Step 1 :
@k8_group_client = Kubeclient::Client.new( 'https://api.kubetnets.dev.XYZ.com/apis',
'extensions/v1beta1',
{
ssl_options: @k8_config.context.ssl_options,
auth_options: @k8_config.context.auth_options
}
)
Step 2 :
json = "{:apiVersion=>"extensions/v1beta1", :kind=>"Deployment", :metadata=>{:name=>"docker-ecs-service"}, :spec=>{:replicas=>2, :template=>{:metadata=>{:labels=>{:app=>"docker-ecs-service"}}, :spec=>{:containers=>[{:name=>"docker-ecs-service", :image=>"docker-registry.XYZ.com/library/docker-ecs-service:c15978d253f6faf8e7fea00270e9014e7b17656d", :ports=>[{:containerPort=>8080}], :env=>[{:name=>"APP_NAME", :value=>"specimen-springboot-docker-ecs-service"}, {:name=>"VERSION", :value=>"c15978d253f6faf8e7fea00270e9014e7b17656d"}], :livenessProbe=>{:httpGet=>{:path=>"/buildInfo", :port=>8080}, :initialDelaySeconds=>45, :timeoutSeconds=>10}}], :imagePullSecrets=>[{:name=>"docker-registry"}]}}}}"

Step 3 : deploy = Kubeclient::Resource.new(json)
Step 4 : @k8_group_client.create_deployment(deploy)

getting error

Users/.rvm/gems/ruby-2.3.1/gems/kubeclient-2.3.0/lib/kubeclient/common.rb:107:in `rescue in handle_exception': the server does not allow this method on the requested resource (KubeException)
	from /Users/.rvm/gems/ruby-2.3.1/gems/kubeclient-2.3.0/lib/kubeclient/common.rb:99:in `handle_exception'
	from /Users/.rvm/gems/ruby-2.3.1/gems/kubeclient-2.3.0/lib/kubeclient/common.rb:315:in `create_entity'
	from /Users/.rvm/gems/ruby-2.3.1/gems/kubeclient-2.3.0/lib/kubeclient/common.rb:191:in `block (2 levels) in define_entity_methods'```

hridyeshpant:
@simon3z here is the actual issue
rest client object should have "**https://api.kubetnets.us-west-2.dev.expedia.com/apis/extensions/v1beta1/namespaces/default/deployments**" end point. but 
rest client is using **https://api.kubetnets.us-west-2.dev.expedia.com/apis/extensions/v1beta1** 

so the below code 
rest_client[ns_prefix + resource_name]
        .post(hash.to_json, { 'Content-Type' => 'application/json' }.merge(@headers))

is failing because the rest_client objet using https://api.kubetnets.us-west-2.dev.expedia.com/apis/extensions/v1beta1.
below is the debug values while making the post call for create_deployment()
hash[:apiVersion] : extensions/v1beta1 
resource_name = deployments
rest_client = https://api.kubetnets.us-west-2.dev.expedia.com/apis/extensions/v1beta1

hridyeshpant:
so after using namespace: 'default' in metadata section, things are working fine.
seems namespace is mandatory

moolitayer:
@hridyeshpant I replaced the reference to the Service class with Kubeclient::Resource.
We are encouraging users to use only Kubeclient::Resource and might deprecate the other  Kubeclient::X classes going forward.

moolitayer:
@hridyeshpant the need to specify namespace is by design. 
It makes sense to support 'default' in the cli but less so in a client library.
If you think the Docs should state that more clearly please suggest a change

moolitayer:
@hridyeshpant @jonmoter can you please approve this change if that looks good to you?

moolitayer:
@enoodle @zeari @cben @zakiva could one of you please review this small readme change?

README.md
```diff
@@ -261,7 +262,7 @@ Input parameter - object of type `Service`, `Pod`, `ReplicationController`.
 The below example is for v1
 
 ```ruby
-service = Service.new
+service = Kubeclient::Resource.new
```
cben: what makes it a service then?  do you need `.kind`?  or is it implicit in passing it to `.create_service()` in the end?
IOW, have you verified the example works?

README.md
```diff
@@ -6,8 +6,9 @@
 [![Dependency Status](https://gemnasium.com/abonas/kubeclient.svg)](https://gemnasium.com/abonas/kubeclient)
 
 A Ruby client for Kubernetes REST api.
-The client supports GET, POST, PUT, DELETE on nodes, pods, secrets, services, replication controllers, namespaces, resource quotas, limit ranges, endpoints, persistent volumes, persistent volume claims, component statuses and service accounts.
+The client supports GET, POST, PUT, DELETE on all the entities available in kubernetes in both the core and group apis.
```
cben: https://kubernetes.io/docs/api/ doesn't talk of "entities", it says "resources" aka "objects".  OTOH "entities" terminology is all around Kubeclient, even baked into the  `client.all_entities` API...
 Maybe say "on all entities (nodes, pods, etc.)" just to be clear.

Consider also linking to https://kubernetes.io/docs/api-reference/v1/operations/ which actually lists the `/api` resources, and/or https://kubernetes.io/docs/reference/ which lists other `/apis` groups.

P.S. not necessarily now but would be nice to have a better example of group api use, not just constructing `/apis/batch` but also using it (`get_jobs` ?)

README.md
moolitayer: The create_service call. Our long term goal is to deprecate all Kubeclient::X classes since all the implementation does with them is get their hash

README.md
moolitayer: Since the entities vs resources was not introduced here, I prefer to leave that out of this pr.
(currently the only mention of resources here is the Resource class)

I agree and would appreciate a fix for that. Maybe event alias all_entities with all_resources
(however before working on that I would consult with @abonas or @simon3z to see if there is a reason entities is used)

moolitayer:
@cben please review

README.md
simon3z: +1 to move in the long term to the Kubernetes/OpenShift nomenclature, of course it's going to break the API (unless we'll support both) so we should consider it for a major version change.

cben:
LGTM

PR #228, opened 2017-03-01
raise KeyError as .fetch would do to behave more intuitive

grosser:
I want to show an error message when a user asks for an unknown context, so catching a KeyError would be much nicer than RuntimeError ... also seem like the correct exception to me when it basically behaves like a .fetch

### DIFF
--- a/lib/kubeclient/config.rb
+++ b/lib/kubeclient/config.rb
@@ -72,13 +72,13 @@ module Kubeclient
         break x['context'] if x['name'] == context_name
       end
 
-      raise "Unknown context #{context_name}" unless context
+      raise KeyError, "Unknown context #{context_name}" unless context
 
       cluster = @kcfg['clusters'].detect do |x|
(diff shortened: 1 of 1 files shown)


simon3z:
LGTM :+1: 
@moolitayer can you review?

simon3z:
@grosser it seems that Travis is failing and it has some troubles displaying the relevant logs about the failures. Maybe you can look after this (check if failures are legitimate, fix if needed, restart travis job by closing/reopening this PR, etc.). Thanks.

grosser:
fixed

moolitayer:
@grosser please take a look [here](https://github.com/grosser/kubeclient/blob/0ad44c42742740a426fa8974803623c240fe2a91/lib/kubeclient/config.rb#L75) 

Looks really similar and probably should have the same exception

grosser:
fixed

simon3z:
@moolitayer can you review/ack?

Issue #230, opened 2017-03-02
relax requirement on recursive-open-struct


jordimassaguerpla:
Hi,
kubeclient requires recursive-open-struct to be 1.0.0. However, there has been, on recursive-open-struct, 2 minor updates (1.0.2) which fix 2 issues:

https://github.com/aetherknight/recursive-open-struct/blob/master/CHANGELOG.md

Could you relax the requirement so that we can use kubeclient with recursive-open-struct 1.0.2? I would suggest to go for "~>1.0.0"

simon3z:
@jordimassaguerpla I am ok with `~>1.0.0`. I think you can safely try to invest some time to send a PR for that and if @moolitayer agrees we can merge it.

simon3z:
Closed by #231

PR #231, opened 2017-03-03
relax requirement on recursive-open-struct

jordimassaguerpla:
kubeclient requires recursive-open-struct to be 1.0.0. However, there
has been, on recursive-open-struct, 2 minor updates (1.0.2) which fix 2
issues:

https://github.com/aetherknight/recursive-open-struct/blob/master/CHANGELOG.md

Let's make the requirement be ~>1.0.0 so that we can take the minor
version fixes

Fixes#230

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -26,6 +26,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'vcr'
   spec.add_development_dependency 'rubocop', '= 0.46.0'
   spec.add_dependency 'rest-client'
-  spec.add_dependency 'recursive-open-struct', '= 1.0.0'
+  spec.add_dependency 'recursive-open-struct', '~> 1.0.0'
   spec.add_dependency 'http', '= 0.9.8'
 end


simon3z:
@moolitayer @cben can you review?

moolitayer:
fixes #231
(space missing above)

cben:
Fixes #230 [PR can't fix itself :)]

LGTM on `~>` not hardcoding minor version.
If the fixes in 1.0.2 are desirable, why not require at least 1.0.2?

BTW do we want to stop before 1.1.0 or before 2.0.0?  If they follow semver, 1.y.z should all be backward-compatible, but with this rate of releases, just covering 1.0.z is fine and safer.
According to https://robots.thoughtbot.com/rubys-pessimistic-operator, if we want >= 1.0.2 and < 1.1.0, the syntax is `~>1.0.2`.

moolitayer:
> Fixes #230 [PR can't fix itself :)]

:smile: 

> LGTM on ~> not hardcoding minor version.
> If the fixes in 1.0.2 are desirable, why not require at least 1.0.2?

Unless there is a specific problem you think we should mitigate, using `~> 1.0.0` would give us the latest available version and also not break users that have a `= 1.0.0` dependency on ROS

> BTW do we want to stop before 1.1.0 or before 2.0.0? If they follow semver, 1.y.z should all be backward-compatible, but with this rate of releases, just covering 1.0.z is fine and safer.

:+1: for not assuming minor version won't break us.

PR #232, opened 2017-03-10
Fix Minitest deprecation

kirs:
Fixes:

```
Use assert_nil if expecting nil from test/test_service.rb:43:in `test_construct_our_own_service'. This will fail in MT6.
```

cc @simon3z

### DIFF
--- a/test/test_service.rb
+++ b/test/test_service.rb
@@ -40,7 +40,7 @@ class TestService < MiniTest::Test
     assert_equal(created.spec.ports.size, our_service.spec.ports.size)
 
     # Check that original entity_config is not modified by kind/apiVersion patches:
-    assert_equal(our_service.kind, nil)
+    assert_nil(our_service.kind)
 
     assert_requested(:post, expected_url, times: 1) do |req|
       data = JSON.parse(req.body)


simon3z:
@moolitayer can you review?

test/test_service.rb
```diff
@@ -40,7 +40,7 @@ def test_construct_our_own_service
     assert_equal(created.spec.ports.size, our_service.spec.ports.size)
 
     # Check that original entity_config is not modified by kind/apiVersion patches:
-    assert_equal(our_service.kind, nil)
+    assert_nil our_service.kind
```
moolitayer: @kirs our style dictates here:
`assert_nil(our_service.kind)`

kirs: I've ran rubocop and it said that `assert_nil our_service.kind` is totally fine. Perhaps we need to update the rubocop config?

kirs: Otherwise it may be confusing for contributors.

simon3z: @kirs IIRC we didn't find a way to enforce that at rubocop level, can you double-check?

kirs: Actually there's something related to that in the Ruby community style guide (https://github.com/bbatsov/ruby-style-guide#no-dsl-decorating):

> Omit both the outer braces and parentheses for methods that are part of an internal DSL.

With `assert_nil` being a DSL method I think it's a good point. Not trying to make you reconsider through.

simon3z: @kirs in this PR I'd go with the parenthesis for consistency, we can reconsider the guidelines going forward.

kirs: Sure thing. I updated the PR

PR #236, opened 2017-03-15
bump rubocop to enforce what our style dictates

grosser:
@moolitayer @simon3z

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,3 +14,9 @@ Metrics/CyclomaticComplexity:
   Max: 8
 Metrics/ModuleLength:
   Enabled: false
+Style/MethodCallWithArgsParentheses:
+  Enabled: true
+  IgnoredMethods: [require, raise, include, attr_reader, refute, assert]
+  Exclude: [Gemfile, Rakefile, kubeclient.gemspec]
+Security/MarshalLoad:
+  Exclude: [test/**/*]
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'minitest'
   spec.add_development_dependency 'webmock', '~> 1.24.2'
   spec.add_development_dependency 'vcr'
-  spec.add_development_dependency 'rubocop', '= 0.46.0'
+  spec.add_development_dependency 'rubocop', '= 0.47.1'
   spec.add_dependency 'rest-client'
   spec.add_dependency 'recursive-open-struct', '~> 1.0.0'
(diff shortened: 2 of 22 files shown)


abonas:
@grosser please elaborate why are all those style changes needed and why rubocop needs to be bumped for that. thanks

grosser:
style changes are needed because they are "what our style dictates" according to @moolitayer in https://github.com/abonas/kubeclient/pull/195 ... and we need a newer rubocop version to enforce that

abonas:
> style changes are needed because they are "what our style dictates" according to @moolitayer in #195 ... and we need a newer rubocop version to enforce that

I can't find in that issue the discussion regarding rubocop. do you have the exact link? cc @moolitayer

grosser:
screen shot 2017-03-28 at 8 14 37 am

moolitayer:
@abonas me and @simon3z have been enforcing this for a while on incoming PRs. Where do you stand on this?

abonas:
> @abonas me and @simon3z have been enforcing this for a while on incoming PRs. Where do you stand on this?

@moolitayer 

I'm OK with it as long as:
1. indeed all past code is aligned to the style - which looks like what this PR should achieve
2. there is a clear pointer for the required style in the readme so it is crystal clear to whoever makes code changes what is required
3. openshift client is aligned and enforced to it as well

I'm not clear why rubocop version bump is needed (not objecting to it, just want to understand) - are those new rules that you are interested to enforce?

grosser:
newer rubocop version has this check

all styles got adjusted ... rubocop will alert new contributors.

grosser:
let's start here and then I can make the same PR for openshift client

abonas:
> let's start here and then I can make the same PR for openshift client

it was brought to my attention that openshift client has much lower priority than I thought, so it probably won't be needed there.

moolitayer:
The only issue I have with this change is that the Readme file code examples should also have parenthesis on method calls. 

Thanks for doing this @grosser

grosser:
looked at the readme, tons of calls and not part of this codebase ... also this PR is big enough.
can we please get this merged now ?

moolitayer:
I'm find with leaving the README for next iteration. @abonas has the final call

abonas:
The readme is definitely part of this codebase (unless I misunderstood what that comment meant), but I am OK with updating it in a separate PR as long as it happens ASAP

grosser:
Endless PR discussions and waiting around is demotivating for me, merge if
it's better then before.

I'd not consider the Readme part of the codebase since it is not 'our' code
/ not examples on how to write this gem, but examples for other people to
copy-paste.
I'll still do the conversion since it's nice to be in sync.

PR #244, opened 2017-04-24
Add requests timeouts option

cben:
I have a use case for extending the http timeout (a big openshift cluster whose /oapi/v1/images takes >2min to compute before it starts sending data).
This PR adds new options to control the 2 timeouts rest-client exposes.

- I wasn't sure if the options should also affect watches, which currently have infinite timeout.  Watch data arrives at unpredictable intervals, I think just adding ability to timeout would not make a useful API, so decided to punt and wait for a use case.

- [x] Backward compatible: Works with rest-client 1.8 and 2.0 (extended travis matrix to test both).
  Default open timeout will depend on ruby version, which is exactly the previous behavior (added tests), users that want ruby-independent behavior can simply specify :open timeout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ uri = URI::HTTP.build(host: "somehostname", port: 8080)
 client = Kubeclient::Client.new(uri)
 ```
 
+### SSL
+
 It is also possible to use https and configure ssl with:
 
 ```ruby
@@ -93,6 +95,8 @@ client = Kubeclient::Client.new(
   'https://localhost:8443/api/', 'v1', ssl_options: ssl_options
 )
 ```
+
+### Authentication
 
 If you are using basic authentication or bearer tokens as described
 [here](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/authentication.md) then you can specify one
@@ -148,6 +152,8 @@ client = Kubeclient::Client.new(
 
 You can find information about tokens in [this guide](http://kubernetes.io/docs/user-guide/accessing-the-cluster/) and in [this reference](http://kubernetes.io/docs/admin/authentication/).
 
+### Non-blocking IO
+
 You can also use kubeclient with non-blocking sockets such as Celluloid::IO, see [here](https://github.com/httprb/http/wiki/Parallel-requests-with-Celluloid%3A%3AIO)
 for details. For example:
 
@@ -162,7 +168,11 @@ client = Kubeclient::Client.new(
 )
 ```
 
-You can also use kubeclient with an http proxy server such as tinyproxy. It can be entered as a string or a URI object
+This affects only `.watch_*` sockets, not one-off actions like `.get_*`, `.delete_*` etc.
+
+### Proxies
+
+You can also use kubeclient with an http proxy server such as tinyproxy. It can be entered as a string or a URI object.
 For example:
 ```ruby
 proxy_uri = URI::HTTP.build(host: "myproxyhost", port: 8443)
@@ -170,6 +180,28 @@ client = Kubeclient::Client.new(
   'https://localhost:8443/api/', http_proxy_uri: proxy_uri
 )
 ```
+
+
+### Timeouts
+
+Watching never times out.
+
+One-off actions like `.get_*`, `.delete_*` have a configurable timeout:
+```ruby
+timeouts = {
+  open: 10,  # unit is seconds
+  read: nil  # nil means never time out
+}
+client = Kubeclient::Client.new(
+  'https://localhost:8443/api/', timeouts: timeouts
+)
+```
+
+Default timeouts match `Net::HTTP` and `RestClient`, which unfortunately depends on ruby version:
+- open was infinite up to ruby 2.2, 60 seconds in 2.3+.
+- read is 60 seconds.
+
+If you want ruby-independent behavior, always specify `:open`.
 
 ### Discovery
 
@@ -287,7 +319,7 @@ service.metadata = {}
 service.metadata.name = "redis-master"
 service.metadata.namespace = 'staging'
 service.spec = {}
-service.spec.ports = [{ 
+service.spec.ports = [{
   'port' => 6379,
   'targetPort' => 'redis-server'
 }]
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -26,6 +26,12 @@ module Kubeclient
       ssl_socket_class: nil
     }.freeze
 
+    DEFAULT_TIMEOUTS = {
+      # These do NOT affect watch, watching never times out.
+      open: Net::HTTP.new('127.0.0.1').open_timeout, # depends on ruby version
+      read: Net::HTTP.new('127.0.0.1').read_timeout
+    }.freeze
+
     DEFAULT_HTTP_PROXY_URI = nil
 
     SEARCH_ARGUMENTS = {
@@ -50,6 +56,7 @@ module Kubeclient
       ssl_options: DEFAULT_SSL_OPTIONS,
       auth_options: DEFAULT_AUTH_OPTIONS,
       socket_options: DEFAULT_SOCKET_OPTIONS,
+      timeouts: DEFAULT_TIMEOUTS,
       http_proxy_uri: DEFAULT_HTTP_PROXY_URI
     )
       validate_auth_options(auth_options)
@@ -63,6 +70,9 @@ module Kubeclient
       @ssl_options = ssl_options
       @auth_options = auth_options
       @socket_options = socket_options
+      # Allow passing partial timeouts hash, without unspecified
+      # @timeouts[:foo] == nil resulting in infinite timeout.
+      @timeouts = DEFAULT_TIMEOUTS.merge(timeouts)
       @http_proxy_uri = http_proxy_uri.to_s if http_proxy_uri
 
       if auth_options[:bearer_token]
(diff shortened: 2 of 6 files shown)


cben:
Opened https://github.com/rest-client/rest-client/issues/591 but I don't think outcome there affects decisions here.

cben:
Oh, I thought it's hard to auto-detect Net::HTTP's builtin timeouts but  `Net::HTTP.new('example.com').open_timeout` works nicely.  So RestClient could preserve ruby-dependent defaults yet report them, if you think that's better.

simon3z:
@cben please fix travis.

README.md
```diff
@@ -148,7 +152,9 @@ client = Kubeclient::Client.new(
 
 You can find information about tokens in [this guide](http://kubernetes.io/docs/user-guide/accessing-the-cluster/) and in [this reference](http://kubernetes.io/docs/admin/authentication/).
 
-You can also use kubeclient with non-blocking sockets such as Celluloid::IO, see [here](https://github.com/httprb/http/wiki/Parallel-requests-with-Celluloid%3A%3AIO)
+### Non-blocking IO
```
moolitayer: :+1:

lib/kubeclient/common.rb
```diff
@@ -26,6 +26,12 @@ module ClientMixin
       ssl_socket_class: nil
     }.freeze
 
+    DEFAULT_TIMEOUTS = {
+      # These do NOT affect watch, watching never times out.
+      connect: 60,
+      read: 60
+    }
```
moolitayer: freeze

moolitayer:
@cben I don't think this forces us to bump major, from your description it sounds like we were inconsistent (between ruby versions) and you are fixing that.

We both should test all the verbs thoroughly. 

cc @grosser (regarding https://github.com/abonas/kubeclient/pull/204)

simon3z:
@cben on the BZ (https://bugzilla.redhat.com/show_bug.cgi?id=1440950) you mentioned that you are considering to reduce this to something easier. Are you working on that now? Any idea on when it will be ready?

cben:
- found way to support older rest-client.
- I'll switch to passing only the timeouts user gives, and will document default open timeout changed in ruby 2.3.  Users that require consistent behavior ≤2.2 can just always pass `:open`.
hope to push it today, maybe tomorrow it test gives me trouble.

cben:
@simon3z @moolitayer Reworked, PTAL.  Now completely compatible, with tests for that.  Updated PR description.

.Gemfile-rest-client-1.8.rb
```diff
@@ -0,0 +1,11 @@
+# Allow travis to test rest-client 1.x.
```
moolitayer: Why are we doing this? is our version not available?

cben: I'm testing both, see .travis.yml.
I'm all for bumping to rest-client >= 2.0 and dropping these at some later point.

cben: Just to double-check this works: 
https://travis-ci.org/abonas/kubeclient/jobs/226372922 run with `Using rest-client 1.8.0`
(using .Gemfile-rest-client-1.8.rb)
https://travis-ci.org/abonas/kubeclient/jobs/226372920 run with `Using rest-client 2.0.2`
(using Gemfile uses gemspec, which doesn't specify rest-client version, so we get latest)

README.md
```diff
@@ -171,6 +181,27 @@ client = Kubeclient::Client.new(
 )
 ```
 
+
+### Timeouts
+
+Watching never times out.
+One-off actions like `.get_*`, `.delete_*` have a timeout, can be changed:
```
moolitayer: s/have a timeout, can be changed/have a configurable timeout/ ?

cben: I wanted to say both that they have a finite timeout (in constrast to "Watching never times out") and that it's configurable.  Yeah, awkward.
I'll adopt "have a configurable timeout", I explain defaults below anyway.

cben: 👍 adopted "have a configurable timeout".

simon3z:
@cben I don't think `.Gemfile-rest-client-1.8.rb` should be an hidden file. Any problem to leave it visible?

.Gemfile-rest-client-1.8.rb
moolitayer: @cben Can you create a pr to bump rest-client ? we should discuss that desperately
(I think we should do it, rest client < 2.0 has not seen updates in over two years)

cben:
I thought it's noise for local development (like .travis.yml).  No problem,
done.

lib/kubeclient/common.rb
```diff
@@ -416,6 +428,18 @@ def api
       JSON.parse(response)
     end
 
+    def self.restclient_read_timeout_option
```
moolitayer: private?

do you mind putting this method next to our exiting validate_X methods?

lib/kubeclient/common.rb
```diff
@@ -42,14 +48,15 @@ module ClientMixin
     attr_reader :headers
     attr_reader :discovered
 
-    def initialize_client(
+    def initialize_client( # rubocop:disable Metrics/ParameterLists
```
moolitayer: I guess we can live with that. We should make travis warn and not fail on rubocop at some point

lib/kubeclient/common.rb
moolitayer: What happens if we pass in both options?

cben: couldn't private because used by test too.
but yeah, this actually exposes it, not nice :-(

- use `send` to bypass private in test?
- move to separate class (Kubeclient::RestClientCompatibility)?

lib/kubeclient/common.rb
```diff
@@ -63,6 +70,9 @@ def initialize_client(
       @ssl_options = ssl_options
       @auth_options = auth_options
       @socket_options = socket_options
+      # Allow passing partial timeouts hash, without unspecified
+      # @timeouts[:foo] == nil resulting in infinite timeout.
```
moolitayer: This comment isn't telling me much. Do what you feel

lib/kubeclient/common.rb
moolitayer: :+1: for send in test

lib/kubeclient/common.rb
cben: see rubocop's PR description.
I believe 8 positional `f(a, b, c, d, e, f, g, h)` would be insane, 
but `f(a, b, c, d, options)` is not so bad, and `f(a, b, c, d, x:, y:, z:, w:)` is strictly cleaner than unpacking `options` inside the function, shouldn't be penalized.

AFAICT rubocop can't make this distinction, but this function is only place >5 total.

lib/kubeclient/common.rb
cben: With other options, you either get FOO_DEFAULTS or replace it entirely.
If you pass less keys than  FOO_DEFAULTS had, it's treated as nil, which is generally buggy, in this case would *silently* give infinite timeout which I felt too subtle, so broke the pattern.

I'll send separate PR to do same with other options (and drop this comment).

lib/kubeclient/common.rb
cben: can't make private at all, as it's a class method called from normal method.
I want the methods probing to get cached once, not per instance.
could make it a const but that feels even more public...

lib/kubeclient/common.rb
moolitayer: :+1: for code.
not sure we should merge all of the other defaults - lets discuss there

why would `@timeouts[:foo] == nil` result in infinite timeout?

cben:
PTAL.
Placated rubocop wrt. Gemfile-rest-client-1.8.rb.
Don't see a way to make `restclient_read_timeout_option` private (that I wouldn't have to explain to rubocop ;-)

lib/kubeclient/common.rb
cben: nil represents infinite timeout to RestClient & Net::HTTP

moolitayer: why would an unexpected key be considered? why is merge solving it?

lib/kubeclient/common.rb
cben: > What happens if we pass in both options?
￼ 
Interesting, you mean:
```
        open_timeout: @timeouts[:open],
        read_timeout: @timeouts[:read], # rest-client 2.x
        timeout: @timeouts[:read]       # rest-client 1.x
```
Yes, that works AFAICT!  I've attempted something similar but not exactly this.

However, while our code would look simpler what actually happens will be harder to reason about.
It'll also be harder to test, due to https://github.com/rest-client/rest-client/issues/591 I discovered while trying.  I even tried testing it upstream in rest-client itself but gave up (https://github.com/rest-client/rest-client/pull/595).

=> Bottom line: I'd prefer the method probing.

cben:
@moolitayer @simon3z responded to all points, believe this is as good as I can make it.

lib/kubeclient/common.rb
moolitayer: Yeah me too actually + we can remove when we bump rest-client

lib/kubeclient/common.rb
moolitayer: ?

lib/kubeclient/common.rb
moolitayer: BTW make sure other parameters weren't renamed as well

lib/kubeclient/common.rb
cben: if you pass just one, `timeouts:{open: 10}` or `timeouts: {read: 20}`.  merge ensures `@timeouts` always has both.

cben:
Found rubocop knob to not penalize keyword args, dropped rubocop:disable 🎉 
Removed dot from `.Gemfile-rest-client-1.8.rb` in one more place in .travis.yml.
Re-squashed commits cleanly.

simon3z:
@cben since for master we're planning another major version, please clean this up (in another PR) requiring the `rest-client` gem and removing the backward compatibility. Thanks.

PR #253, opened 2017-06-14
Updated outdated dependencies

Abdulwahaab710:
### In this PR I have updated the following outdated dependencies:
- http
- rake
- recursive-open-struct
- rubocop
- webmock


abonas:
hi, @Abdulwahaab710 .
please squash your commits, they have a similar name, and they have the same purpose.
also, travis is failing, please check its error messages. mostly I believe it's due to version bump of rubocop and the fix of [this](https://github.com/bbatsov/rubocop/issues/4039).

Abdulwahaab710:
Done :)

simon3z:
@moolitayer can you review? (mostly for new dependencies versions implications)

cben:
- dev dependencies are safe to bump.
- https://github.com/aetherknight/recursive-open-struct/blob/master/CHANGELOG.md
1.0.3 adds ruby 2.4 compat, but stops worrying about 2.0, 2.1 which kubeclient still supports.  However 2.0, 2.1 still pass tests.  I hope they'll bump to at least 1.1.0 when it actually breaks 2.0, 2.1.
- https://github.com/httprb/http/blob/master/CHANGES.md looks sane.

moolitayer:
> I hope they'll bump to at least 1.1.0 when it actually breaks ruby 2.0...

I doubt if they will even notice when that happens...

We probably want to enjoy @jrafanie's https://github.com/aetherknight/recursive-open-struct/pull/52 with all the performance work going around, so I would consider dropping support of ruby 2.0 & 2.1 and bumping kubeclient's version. I'm not sure how/if that is problematic to users.

@kirs @grosser @albertrdixon care to comment on dropping ruby 2.0, 2.1 support ?

grosser:
everything up to 2.1 is unsupported https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/ so I'm in favor of getting rid of it ... make the new gem have a `require_ruby_version` and it should not break for people stuck on older versions ...

jrafanie:
> We probably want to enjoy @jrafanie's aetherknight/recursive-open-struct#52 with all the performance work going around,

@moolitayer I doubt that would help most use cases... I would think you would need an enormous recursive structure to see much difference.  I could be wrong though. 

I'm all for dropping unsupported rubies in a non-patch release 👍

grosser:
FYI ROS removal issue https://github.com/abonas/kubeclient/issues/251

kirs:
👍 for dropping anything < 2.1

simon3z:
@Abdulwahaab710 @grosser @moolitayer @jrafanie we may want to change the `.travis.yml` as well at some point (now is running against `2.0` and `2.1` as well).

simon3z:
@Abdulwahaab710 if you squash your 2 commits in 1 I think we're good to merge.

.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: ruby
 rvm:
-  - "2.0"
-  - "2.1"
```
Abdulwahaab710: @simon3z I have removed rvm 2.0 and rvm 2.1

cben:
We forgot to update `required_ruby_version` in gemspec, it's still 2.0.0 but we now have non-2.0 syntax.

Issue #237, opened 2017-03-15
switch http client to something that does not have a native extension


grosser:
faraday or httparty would come to my mind ...

```
 gem i rest-client
Fetching: unf_ext-0.0.7.2.gem (100%)
Building native extensions.  This could take a while...
Successfully installed unf_ext-0.0.7.2
Fetching: unf-0.1.4.gem (100%)
Successfully installed unf-0.1.4
Fetching: domain_name-0.5.20170223.gem (100%)
Successfully installed domain_name-0.5.20170223
Fetching: http-cookie-1.0.3.gem (100%)
Successfully installed http-cookie-1.0.3
Fetching: mime-types-data-3.2016.0521.gem (100%)
Successfully installed mime-types-data-3.2016.0521
Fetching: mime-types-3.1.gem (100%)
Successfully installed mime-types-3.1
Fetching: netrc-0.11.0.gem (100%)
Successfully installed netrc-0.11.0
Fetching: rest-client-2.0.1.gem (100%)
Successfully installed rest-client-2.0.1
8 gems installed
```

cben:
Oh, I thought Kubeclient is purely rest-client but turns out we also use [http gem](https://github.com/httprb/http) for `WatchStream` functionality.  And http also requires native extenstions in `nokogiri`.

I propose any replacement should:
- support streaming so it can replace both rest-client & http.
- use HTTP persistent connections.  rest-client opens new socket (and TLS handshake) for every request :-( https://github.com/rest-client/rest-client/issues/453

And old but detailed comparison: https://bibwild.wordpress.com/2012/04/30/ruby-http-performance-shootout-redux/
=> [httpclient gem](https://github.com/nahi/httpclient) came on top there, pure ruby (**zero deps :1st_place_medal:**), has persistent connections + thread safe.  TOCHECK:
- [ ] rerun those benchmarks, there is a newer fork but also needs some updating
- [ ] does httpclient support GET streaming?  `get_content` yields chunks of String, `get_async` returns [HTTPClient::Connection](http://www.rubydoc.info/gems/httpclient/2.1.5.2/HTTPClient/Connection) — sounds good on paper.

https://jvns.ca/blog/2016/03/04/whats-up-with-ruby-http-libraries/ sumarises nahi's epic table https://www.slideshare.net/HiroshiNakamura/rubyhttp-clients-comparison, with a crucial point that at the bottom everything builds on just 4 libs: Net:HTTP, Excon, httpclient and native libcurl.
[...]

PR #254, opened 2017-06-28
Process JSON Body of ListObjects using streaming json parser

ilackarms:
This PR changes `Kubeclient::get_entities` as follows:

1. Use `RestClient::Request.execute` in place of `RestClient::Resource.get` to read HTTP Response body in chunks.
2. Use [`Json::Streamer`](https://github.com/thisismydesign/json-streamer) to parse chunks as a JSON stream.
3. If the caller passes a block, we yield entities as they are parsed from the stream of json. Otherwise we collect entities and return them as a `Kubeclient::Common::EntityList` as before.

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -30,4 +30,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'rest-client', '~> 2.0'
   spec.add_dependency 'recursive-open-struct', '~> 1.0.4'
   spec.add_dependency 'http', '~> 2.2.2'
+  spec.add_dependency 'json-streamer', '~> 2.0.0'
 end
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -1,5 +1,7 @@
 require 'json'
 require 'rest-client'
+require 'json/streamer'
+
 module Kubeclient
   # Common methods
   # this is mixed in by other gems
@@ -113,9 +115,7 @@ module Kubeclient
       rescue JSON::ParserError
         {}
       end
-      err_message = json_error_msg['message'] || e.message
-      error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
-      raise error_klass.new(e.http_code, err_message, e.response)
+      raise_http_err(json_error_msg, e)
     end
 
     def discover
@@ -176,8 +176,8 @@ module Kubeclient
       @entities.values.each do |entity|
         klass = ClientMixin.resource_class(@class_owner, entity.entity_type)
         # get all entities of a type e.g. get_nodes, get_pods, etc.
-        define_singleton_method("get_#{entity.method_names[1]}") do |options = {}|
-          get_entities(entity.entity_type, klass, entity.resource_name, options)
+        define_singleton_method("get_#{entity.method_names[1]}") do |options = {}, &block|
+          get_entities(entity.entity_type, klass, entity.resource_name, options, &block)
         end
 
         # watch all entities of a type e.g. watch_nodes, watch_pods, etc.
@@ -218,19 +218,7 @@ module Kubeclient
 
     def create_rest_client(path = nil)
       path ||= @api_endpoint.path
-      options = {
-        ssl_ca_file: @ssl_options[:ca_file],
-        ssl_cert_store: @ssl_options[:cert_store],
-        verify_ssl: @ssl_options[:verify_ssl],
-        ssl_client_cert: @ssl_options[:client_cert],
-        ssl_client_key: @ssl_options[:client_key],
-        proxy: @http_proxy_uri,
-        user: @auth_options[:username],
-        password: @auth_options[:password],
-        open_timeout: @timeouts[:open],
-        read_timeout: @timeouts[:read]
-      }
-      RestClient::Resource.new(@api_endpoint.merge(path).to_s, options)
+      RestClient::Resource.new(@api_endpoint.merge(path).to_s, rest_client_options)
     end
 
     def rest_client
@@ -267,7 +255,8 @@ module Kubeclient
     #
     #   Default response type will return a collection RecursiveOpenStruct
     #   (:ros) objects, unless `:as` is passed with `:raw`.
-    def get_entities(entity_type, klass, resource_name, options = {})
+    def get_entities(entity_type, klass, resource_name, options = {}, &block)
+      return get_entities_async(klass, resource_name, options, &block) if block_given?
       params = {}
       SEARCH_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
 
@@ -443,6 +432,96 @@ module Kubeclient
 
     private
 
+    def do_rest_request(args)
+      uri = @api_endpoint.merge("#{@api_endpoint.path}/#{@api_version}/#{args[:path]}")
+      options = {
+        url: uri.to_s,
+        method: args[:method],
+        headers: (args[:headers] || {}).merge('params' => args[:params])
+      }.merge(rest_client_options)
+      options[:block_response] = args[:block] if args[:block]
+      resp = RestClient::Request.execute(options)
+      return unless resp.instance_of?(Net::HTTPClientError)
+      code = resp.code
+      code = code.to_i if code.instance_of?(String)
+      raise HttpError.new(code, resp.message, resp)
+    end
+
+    def rest_client_options
+      {
+        ssl_ca_file: @ssl_options[:ca_file],
+        ssl_cert_store: @ssl_options[:cert_store],
+        verify_ssl: @ssl_options[:verify_ssl],
+        ssl_client_cert: @ssl_options[:client_cert],
+        ssl_client_key: @ssl_options[:client_key],
+        proxy: @http_proxy_uri,
+        user: @auth_options[:username],
+        password: @auth_options[:password],
+        open_timeout: @timeouts[:open],
+        read_timeout: @timeouts[:read]
+      }
+    end
+
+    def get_entities_async(klass, resource_name, options = {})
+      if options[:as] == :raw
+        raise ArgumentError('Cannot pass block to get_entities when requesting options[:as] == raw')
+      end
+
+      params = {}
+      SEARCH_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
+
+      ns_prefix = build_namespace_prefix(options[:namespace])
+
+      conditions = Json::Streamer::Conditions.new
+      conditions.yield_object = lambda do |aggregator:, object:|
+        aggregator.level.eql?(2) && aggregator.key_for_level(1).eql?('items')
+      end
+
+      entity_streamer = Json::Streamer::JsonStreamer.new
+      # result['items'] might be nil due to https://github.com/kubernetes/kubernetes/issues/13096
+      entity_streamer.get_with_conditions(conditions) do |object|
+        entity = new_entity(object, klass)
+        yield(entity)
+      end
+
+      resource_version = nil
+      resource_version_streamer = Json::Streamer::JsonStreamer.new
+      resource_version_streamer.get(key: 'resourceVersion') do |value|
+        resource_version = value.to_i unless resource_version
+      end
+
+      kind = nil
+      kind_streamer = Json::Streamer::JsonStreamer.new
+      kind_streamer.get(key: 'kind') do |value|
+        kind = value unless kind
+      end
+
+      block = proc do |response|
+        if response.instance_of?(Net::HTTPNotFound)
+          raise ResourceNotFoundError.new(404,
+                                          'Not Found',
+                                          response)
+        end
+        response.read_body do |chunk|
+          entity_streamer.parser << chunk
+          resource_version_streamer.parser << chunk unless resource_version
+          kind_streamer.parser << chunk unless kind
+        end
+      end
+
+      handle_exception do
+        # disable compression by setting Accept-Encoding to empty string
+        do_rest_request(path: ns_prefix + resource_name,
+                        block: block, method: :get,
+                        headers: @headers.merge('Accept-Encoding' => ''), params: params)
+      end
+
+      {
+        resourceVersion: resource_version,
+        kind: kind
+      }
+    end
+
     def load_entities
       @entities = {}
       fetch_entities['resources'].each do |resource|
@@ -507,6 +586,12 @@ module Kubeclient
       end
 
       options.merge(@socket_options)
+    end
+
+    def raise_http_err(json_error_msg, e)
+      err_message = json_error_msg['message'] || e.message
+      error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
+      raise error_klass.new((e.http_code || 404), err_message, e.response)
     end
   end
 end
--- a/test/test_kubeclient.rb
+++ b/test/test_kubeclient.rb
@@ -209,7 +209,57 @@ class KubeclientTest < MiniTest::Test
     assert_equal(404, exception.error_code)
   end
 
-  def test_entity_list
+  def test_entity_list_with_block
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
+    stub_request(:get, %r{/services})
+      .to_return(body: open_test_file('entity_list.json'), status: 200)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+    services = []
+    metadata_object = client.get_services do |service|
+      services << service
+    end
+
+    assert_equal(metadata_object, resourceVersion: 59, kind: 'ServiceList')
+    refute_empty(services)
+    assert_equal(2, services.size)
+    assert_instance_of(Kubeclient::Service, services[0])
+    assert_instance_of(Kubeclient::Service, services[1])
+
+    assert_requested(:get, 'http://localhost:8080/api/v1/services', times: 1)
+  end
+
+  def test_entity_list_with_block_404
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
+    stub_request(:get, %r{/services})
+      .to_return(status: 404)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+
+    exception = assert_raises(Kubeclient::ResourceNotFoundError) do
+      client.get_services {}
+    end
+    assert(exception.message.include?('Not Found'))
+    assert_equal(404, exception.error_code)
+  end
+
+  def test_entity_list_with_block_http_err
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
+    stub_request(:get, %r{/services})
+      .to_return(body: '{"err": "I\'m a teapot!"}', status: 418)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+
+    exception = assert_raises(Kubeclient::HttpError) do
+      client.get_services {}
+    end
+    assert_equal(418, exception.error_code)
+  end
+
+  def test_entity_list_without_block
     stub_request(:get, %r{/api/v1$})
       .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
     stub_request(:get, %r{/services})
@@ -573,10 +623,13 @@ class KubeclientTest < MiniTest::Test
       auth_options: { user: 'username', password: 'password' }
     )
 
-    pods = client.get_pods
+    pods = []
+    client.get_pods do |pod|
+      pods << pod
+    end
 
-    assert_equal('Pod', pods.kind)
     assert_equal(1, pods.size)
+    assert_equal('Kubeclient::Pod', pods[0].class.to_s)
     assert_requested(:get, 'http://localhost:8080/api/v1/pods', times: 1)
   end
 
@@ -704,10 +757,14 @@ class KubeclientTest < MiniTest::Test
       auth_options: { bearer_token_file: file }
     )
 
-    pods = client.get_pods
+    pods = []
+    metadata_object = client.get_pods do |pod|
+      pods << pod
+    end
 
-    assert_equal('Pod', pods.kind)
+    assert_equal(metadata_object, resourceVersion: 1315, kind: 'PodList')
     assert_equal(1, pods.size)
+    assert_equal('Kubeclient::Pod', pods[0].class.to_s)
   end
 
   def test_proxy_url
--- a/README.md
+++ b/README.md
@@ -304,6 +304,22 @@ pods = client.get_pods as: :raw
 node = client.get_node "127.0.0.1", as: :raw
 ```
 
+### Yielding lists entity-by-entity
+ By passing a block to any `get_*` method that returns a list (e.g.: `get_pods`, `get_nodes`, `get_namespaces`), entities will be yielded one-by-one to the passed block rather than returned as an array. This permits handling entities in large lists asynchronously while parsing takes place.
+   
+ For example:
+ ```ruby
+ client.get_services do |service, resource_version|
+   puts service
+   puts resource_version
+ end
+    
+ #<Kubeclient::Service metadata={:name=>"kubernetes", :namespace=>"default", :selfLink=>"/api/v1/services/kubernetes?namespace=default", :uid=>"016e9dcd-ce39-11e4-ac24-3c970e4a436a", :resourceVersion=>"6", :creationTimestamp=>"2015-03-19T15:08:16+02:00", :labels=>{:component=>"apiserver", :provider=>"kubernetes"}}, spec={:port=>443, :protocol=>"TCP", :selector=>nil, :clusterIP=>"10.0.0.2", :containerPort=>0, :sessionAffinity=>"None"}, status={}>
+ 59
+ #<Kubeclient::Service metadata={:name=>"kubernetes-ro", :namespace=>"default", :selfLink=>"/api/v1/services/kubernetes-ro?namespace=default", :uid=>"015b78bf-ce39-11e4-ac24-3c970e4a436a", :resourceVersion=>"5", :creationTimestamp=>"2015-03-19T15:08:15+02:00", :labels=>{:component=>"apiserver", :provider=>"kubernetes"}}, spec={:port=>80, :protocol=>"TCP", :selector=>nil, :clusterIP=>"10.0.0.1", :containerPort=>0, :sessionAffinity=>"None"}, status={}>
+ 59
+ ```
+
 #### Delete an entity (by name)
 
 For example: `delete_pod "pod name"` , `delete_replication_controller "rc name"`, `delete_node "node name"`, `delete_secret "secret name"`


ilackarms:
@agrare please review; @simon3z cc

moolitayer:
Just noticed `ClientMixin` uses `rest-client` while `WatchStream` uses `http`.
Can we get rid of that one of the two? Also it seams `WatchStream` was created for stream processing, maybe the new code should live there?

ilackarms:
@moolitayer it's certainly possible to use `RestClient` in place of `HTTP:Client` in `WatchStream`. However I think that is outside the scope of this PR

ilackarms:
@moolitayer `kubeclient.get_entities` parses the nested json inside of a single large kubernetes list object (as a single json object), while `watchstream.each` parses json objects as strings separated by newlines. there's probably some overlapping functionality that can be refactored there, but they will require different implementations

lib/kubeclient/common.rb
```diff
@@ -115,7 +117,7 @@ def handle_exception
       end
       err_message = json_error_msg['message'] || e.message
       error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
-      raise error_klass.new(e.http_code, err_message, e.response)
+      raise error_klass.new((e.http_code || 404), err_message, e.response)
```
NickLaMuro: To fix some of the rubocop errors you are getting you might want to consider a helper method for this:

```diff
-  error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
-  raise error_klass.new(e.http_code, err_message, e.response)
+  raise http_error(err_message, e.response, e.http_code)

+def http_error(err_message, e.response, http_code = 404)
+  error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
+  error_klass.new(e.http_code, err_message, e.response)
+end
```

agrare:
@ilackarms @abonas The interface here is great, without a block this behaves exactly as it used to and with a block I can parse the pods one at a time.

lib/kubeclient/common.rb
```diff
@@ -508,5 +553,11 @@ def http_options(uri)
 
       options.merge(@socket_options)
     end
+
+    def http_err(json_error_msg, e)
+      err_message = json_error_msg['message'] || e.message
+      error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
+      raise error_klass.new((e.http_code || 404), err_message, e.response)
+    end
```
moolitayer: The next code change here is 
```
- raise error_klass.new(e.http_code, err_message, e.response)
+ raise error_klass.new((e.http_code || 404), err_message, e.response)
```
looks good

kubeclient.gemspec
```diff
@@ -30,4 +30,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'rest-client'
   spec.add_dependency 'recursive-open-struct', '~> 1.0.4'
   spec.add_dependency 'http', '~> 2.2.2'
+  spec.add_dependency 'json-streamer'
```
abonas: this library has 1 contributor and from my impression low activity/users/forks etc.
why was this specific one selected?

ilackarms: it's the only one I'm aware of that provides the high-level utility of parsing whole objects out of streaming json bodies. I agree that it's a small project and preferable to use something with a bit of a community around it.

there is https://github.com/dgraham/json-stream which has more stars (if that's an indicator or anything). a bit lower level than `json-streamer`, it parses tokens from a streaming body. However in order to use it, we'd effectively have to duplicate the work that `json-streamer` takes care of for us (filtering objects by nesting level, key name, etc.).

I'm open to changing it to a more active/popular project. Do you have any suggestions?

abonas: I'm unfamiliar so can't recommend anything specific. But I think that it should be taken into consideration shall fixes/bugs will be needed in that lib, and there's only one contributor (so there's a risk that he/she also stop developing it in any moment)

lib/kubeclient/common.rb
```diff
@@ -508,5 +553,11 @@ def http_options(uri)
 
       options.merge(@socket_options)
     end
+
+    def http_err(json_error_msg, e)
+      err_message = json_error_msg['message'] || e.message
+      error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
```
abonas: is there a way not to hardcode 404 here and in the row below?

lib/kubeclient/version.rb
```diff
@@ -1,4 +1,4 @@
 # Kubernetes REST-API Client
 module Kubeclient
-  VERSION = '2.3.0'.freeze
+  VERSION = '2.4.0'.freeze
```
abonas: version bumps should be a separate change

lib/kubeclient/common.rb
```diff
@@ -233,6 +233,27 @@ def create_rest_client(path = nil)
       RestClient::Resource.new(@api_endpoint.merge(path).to_s, options)
     end
 
+    def do_rest_request(args)
```
moolitayer: This one looks really similar to create_rest_client. Is there anyway to reuse it?

Gemfile
```diff
@@ -1,4 +1,6 @@
 source 'https://rubygems.org'
 
+gem 'json-streamer', :git => "git://github.com/thisismydesign/json-streamer"
```
moolitayer: why are we depending on source and not on a gem @ilackarms

abonas: https://github.com/abonas/kubeclient/pull/254/commits/6d46d6d82cd709f975cff97279a6a8bfeca94060 I saw it was removed in another commit (which also looks redundant)

lib/kubeclient/common.rb
```diff
@@ -114,8 +116,8 @@ def handle_exception
         {}
       end
       err_message = json_error_msg['message'] || e.message
-      error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
-      raise error_klass.new(e.http_code, err_message, e.response)
+      error_klass = e.http_code == 404 || e.instance_of?(RestClient::NotFound) ? ResourceNotFoundError : HttpError
```
moolitayer: this is getting a little too complicated for a ternary expression. 
Consider switching it to a dumb if

test/test_kubeclient.rb
```diff
@@ -440,10 +460,13 @@ def test_api_basic_auth_back_comp_success
...
 
-    pods = client.get_pods
+    pods = []
+    client.get_pods do |pod|
+      pods << pod
+    end
 
-    assert_equal('Pod', pods.kind)
     assert_equal(1, pods.size)
+    assert_equal('Kubeclient::Pod', pods[0].class.to_s)
```
moolitayer: why do we need this changes?

ilackarms: a kubernetes `PodList` object has a field `Kind` which is essentially its class name. Since we're no longer returning a `PodList` (if `block_given?`), `Kind` is no longer available to us. Since i removed the line `assert_equal('Pod', pods.kind)`,  I wanted to preserve the logic.

ilackarms:
@abonas is this ok for merge?

abonas:
> @abonas is this ok for merge?

I did not see comments from  8 days ago adressed, so I don't think it's ready for merge yet.

kubeclient.gemspec
ilackarms: what would you recommend?

lib/kubeclient/common.rb
ilackarms: when using `Restclient::Request.execute` (line 254), `RestClient` doesn't throw errors on 404 / Resource Not Found. I think looking for a 404 status is the most consistent way to check that the requested resource is not found here. Other than checking for `Not Found` in the response string (which may or may not be present).

lib/kubeclient/common.rb
```diff
@@ -268,21 +289,45 @@ def get_entities(entity_type, klass, resource_name, options = {})
...
-      response = handle_exception do
-        rest_client[ns_prefix + resource_name]
-          .get({ 'params' => params }.merge(@headers))
+
+      resource_version = nil
+      entity_streamer = Json::Streamer::JsonStreamer.new
+      resource_version_streamer = Json::Streamer::JsonStreamer.new
+
+      resource_version_streamer.get(key: 'resourceVersion') do |value|
+        resource_version = value unless resource_version
```
moolitayer: We should probably avoid the streamer in requests that don't ask for it.
(for example if it is broken it should not effect other requests)

ilackarms: you mean, use the old model (synchronous `GET`)? how should the caller ask for it?

moolitayer: If a user did not pass a block we probably want a synchronous request.
See: https://github.com/abonas/kubeclient/issues/263#issuecomment-324091004

In that case we probably want the resource version code to behave as before

ilackarms: @moolitayer updated

lib/kubeclient/common.rb
moolitayer: Extracting the common options should be enough actually (`common_options`?)

Also consider keeping the pattern we have in this file of passing path and headers later on and memoize the client if possible.

```
rest_client[ns_prefix + resource_name]
  .get({ 'params' => params }.merge(@headers))
end
```

lib/kubeclient/common.rb
```diff
@@ -286,6 +308,45 @@ def get_entities(entity_type, klass, resource_name, options = {})
...
+        entity = new_entity(object, klass)
+        yield(entity, resource_version)
+      end
+
+      block = proc do |response|
+        if response.instance_of?(Net::HTTPNotFound)
+          raise ResourceNotFoundError.new(404,
+                                          'Not Found',
+                                          response)
+        end
```
moolitayer: can we add a test for the negative flow?

lib/kubeclient/common.rb
```diff
@@ -286,6 +308,45 @@ def get_entities(entity_type, klass, resource_name, options = {})
...
+          entity_streamer.parser << chunk
+        end
+      end
+
+      handle_exception do
+        do_rest_request(path: ns_prefix + resource_name,
+                        block: block, method: :get, headers: @headers, params: params)
+      end
+
+      return if block_given?
```
moolitayer: isn't this true always?

lib/kubeclient/common.rb
moolitayer: @abonas was talking about extracting 404 to a constant if I understand correctly. I'm okay with both ways.

moolitayer:
@ilackarms can you add a readme note?

kubeclient.gemspec
kirs: From `json-streamer` description:

> The gem uses JSON::Stream's events in the background. It was chosen because it's a pure Ruby parser. A similar implementation can be done using the ~10 times faster Yajl::FFI gem that is dependent on the native YAJL library.

I think its quality is good enough, and considering an alternative of depending on a native library I'd much prefer using `json-streamer` 👍

lib/kubeclient/common.rb
```diff
@@ -286,6 +308,45 @@ def get_entities(entity_type, klass, resource_name, options = {})
       Kubeclient::Common::EntityList.new(entity_type, resource_version, collection)
     end
 
+    def get_entities_async(klass, resource_name, options = {})
```
kirs: Is this method really async? Or it's more of a lazy execution? Usually `async` is a sign that the procedure is using threads, which I think is not the case here?

kirs:
I know that there were plans to get rid of rest-client (https://github.com/abonas/kubeclient/issues/237). Does this PR ties the gem heavier to rest-client?

ilackarms:
@kirs it would actually be easier to do this without Rest Client. A number of sketchy lines had to be added because Rest Client behaves differently for sync / async requests

lib/kubeclient/common.rb
ilackarms: it's async in that it's non-blocking. entities are `yield`ed to the block passed by the user as they are read from a streaming HTTP response

grosser:
could you add a short explanation as to why users would want this (speed ? saving memory?) ... maybe even to the readme "to do xyz pass a block ..."

kubeclient.gemspec
jrafanie: @moolitayer I don't know.  It hasn't been active for over 1 year.  Maybe it's stable or just not that widely used or both.  Has anyone reviewed the code?  It's very small so maybe if someone feels comfortable with the code, it would be ok to use.  Do we have a way to test it against a large sample set of json reponses to make us feel better about it.

Is this opt-in or a required change?  Will we bump versions to indicate this?

lib/kubeclient/common.rb
cben: can probably switch to using `do_rest_request` everywhere.
except `rest_client`, `create_rest_client` are public, in theory someone might depend on them.

another direction: you only did this to be able to pass `:block_response`, right?
I think RestClient::Resource can pass through any option, including `:block_response`...
https://github.com/rest-client/rest-client/blob/master/lib/restclient/resource.rb#L51

lib/kubeclient/common.rb
```diff
@@ -286,6 +308,45 @@ def get_entities(entity_type, klass, resource_name, options = {})
...
+      end
+
+      block = proc do |response|
+        if response.instance_of?(Net::HTTPNotFound)
+          raise ResourceNotFoundError.new(404,
+                                          'Not Found',
+                                          response)
+        end
+        response.read_body do |chunk|
+          entity_streamer.parser << chunk
```
cben: Shouldn't you also feed it `resource_version_streamer`?  AFAICT it's never fed, so `resource_version` will remain nil.  Are you testing yielded resource_version and do tests pass?

P.S. 2 streamers means everything gets parsed twice?  Or will the `resource_version_streamer` know its work is done as soon as it saw `resource_version`?

lib/kubeclient/common.rb
```diff
@@ -286,6 +308,45 @@ def get_entities(entity_type, klass, resource_name, options = {})
...
+      resource_version = nil
+      entity_streamer = Json::Streamer::JsonStreamer.new
+      resource_version_streamer = Json::Streamer::JsonStreamer.new
+
+      resource_version_streamer.get(key: 'resourceVersion') do |value|
+        resource_version = value unless resource_version
+      end
+
+      # result['items'] might be nil due to https://github.com/kubernetes/kubernetes/issues/13096
+      entity_streamer.get(nesting_level: 2, yield_values: false) do |object|
```
cben: What makes this handle "items" specifically?  is it just the level 2?  (if so why doesn't that also get "metadata" values?)
In principle, one day kubernetes might return other nested fields, I'd prefer to ask for "items" specifically.

lib/kubeclient/common.rb
```diff
@@ -286,6 +308,45 @@ def get_entities(entity_type, klass, resource_name, options = {})
...
+        entity = new_entity(object, klass)
+        yield(entity, resource_version)
```
cben: proposal for future extensibility: `yield(entity, {"resource_version": resource_version)`
or whole metadata_hash...

lib/kubeclient/common.rb
```diff
@@ -113,9 +115,7 @@ def handle_exception
       rescue JSON::ParserError
         {}
       end
-      err_message = json_error_msg['message'] || e.message
-      error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
-      raise error_klass.new(e.http_code, err_message, e.response)
+      http_err(json_error_msg, e)
```
cben: I find it unclear from the "http_err" name what it will do, I'd expect it to just return some kind of err.
Perhaps "raise_http_err"?
[bikeshedding, your call]

lib/kubeclient/common.rb
```diff
@@ -508,5 +569,11 @@ def http_options(uri)
 
       options.merge(@socket_options)
     end
+
+    def http_err(json_error_msg, e)
+      err_message = json_error_msg['message'] || e.message
+      error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
+      raise error_klass.new((e.http_code || 404), err_message, e.response)
```
cben: why did you add the `|| 404`?
I guess http_code can be missing if error happened earlier (socket refused etc), in which case setting 404 is misleading.

lib/kubeclient/common.rb
cben: (to clarify, I don't mind the textual duplication, but I'd rather avoid going through 2 code paths in RestClient.)

kubeclient.gemspec
cben: json-streamer is literally [113 lines](https://github.com/thisismydesign/json-streamer/blob/master/lib/json/streamer/json_streamer.rb).  didn't follow in detail but my impression it's reasonably clean, has 100% test coverage (though tests don't use complicated data).  :+1:

- alternative: https://github.com/bdevel/json_stream_trigger — slightly more complicated, simplistic regexp-based JsonPath parser worries me, 8 commits 2 years ago and dormant since then.

json-stream: not very active but is used by 10 gems.  test look ...long :-) (though again "real world fixture" doc is 1 toy doc). :+1:

=> IMO this dep is fine for an opt-in feature.  It's not as battle-tested as JSON.parse, and slower, so it's good you changed the PR not to always use it.

lib/kubeclient/common.rb
ilackarms: for the synchrounous `get_entities`, we use `RestClient::Resource.new().get` which returns `ResourceNotFoundError` if the HTTP status code is a 404.
for the async `get_entities`, we use `RestClient::Request.execute()` which does **not** return an error on 404. the purpose of this change is to catch the 404 and propagate a `ResourceNotFoundError` (to maintain consistency with the public API of this function)

cben: Hmm, perhaps I see now.
The scenario you talk about is `handle_exceptions` rescuing the `raise ResourceNotFoundError` within `block` on L333?
And that's Kubeclient's ResourceNotFoundError and it has no `.http_code`?  If so, how about raising `RestClient::NotFound` there, with http_code == 404, and letting this code translate it to `raise ResourceNotFoundError`?

lib/kubeclient/common.rb
```diff
@@ -286,6 +308,46 @@ def get_entities(entity_type, klass, resource_name, options = {})
...
+
+      # result['items'] might be nil due to https://github.com/kubernetes/kubernetes/issues/13096
+      entity_streamer.get(nesting_level: 2, yield_values: false) do |object|
+        entity = new_entity(object, klass)
+        yield(entity, resource_version)
+      end
+
+      block = proc do |response|
+        if response.instance_of?(Net::HTTPNotFound)
+          raise ResourceNotFoundError.new(404,
```
cben: I believe RestClient::Resource raises for ANY non-2xx status, not just 404.  Should have similar logic here.
(Consider 5xx errors, or  kubernetes saying 418 I'm a teapot.  Obviously we must support that!)
Also, what about 3xx redirects?

I'd love if we can to converge on just 1 way to call RestClient, either Resource or .execute, for both flows.

kubeclient.gemspec
moolitayer: btw I think https://github.com/thisismydesign/json-streamer/commits/master was last updated in March this year

jrafanie: Thanks, @moolitayer I must have misread it as 2016.  Yeah, I'm fine with this considering how small the project is and if it's opt-in.  Since rubygems doesn't have opt-in dependencies, we'd still have to bump the version if we add the new dependency.

NickLaMuro: Can we make this an optional dependency, and raise a graceful error if this is attempted to be used and isn't available?  Haven't looked at the implementation in this PR recently if this is possible/viable, just a suggestion potentially remove some concerns regarding directly adding a new dependency to this gem.

That all said, going with this approach does then introduce complexity into the codebase for handling this, so it isn't something you would consider doing lightly.  Very much related to the discussion here:  https://github.com/abonas/kubeclient/issues/263

cben: If there are concerns, perhaps Kubeclient can start from offering streamed `as: :raw` interface, and leave it to caller (eg. ManageIQ) to use json-streamer (or whatever).

- Nick mentioned on #263 great example of less trivial use:
  Writing raw stream to temp file (to relieve k8s server ASAP), then take our time to stream-parse & process from that.

However, that way callers needs to know more about k8s response structure, error handling etc :-(
Exposing processing & error handling in reusable way would add a lot of "API surface"...
Whereas "we do everything, just give us a block" is neatly contained. :-)

@ilackarms personally I'm happy to review whatever direction you want to proceed (including current).

moolitayer: @NickLaMuro we had a discussion about introducing optional dependencies back in https://github.com/abonas/kubeclient/pull/213. My view, given the comments above and that this gem does not bring additional dependencies we should be fine. @simon3z/@abonas have final call. (we would need to bump version, as mentioned)

lib/kubeclient/common.rb
ilackarms: @cben not sure how to handle 3xx redirects. easiest solution is to simply return ResourceNotFound for any non 2xx status

lib/kubeclient/common.rb
ilackarms: added some tests to handle exceptions / bad status codes

lib/kubeclient/common.rb
ilackarms: yep, removed

ilackarms:
@moolitayer added readme note

lib/kubeclient/common.rb
ilackarms: @cben I agree that 2 code paths are bad. Unfortunately, you can't pass `:block_response` to `RestClient.get`, et al. (see https://github.com/rest-client/rest-client#block_response-receives-raw-nethttpresponse), so all the other code paths would have to be refactored to use `RestClient::Request.execute` rather than `create_rest_client` as they currently do. Since the scope of this change would be pretty large, I didn't want to include it in my PR.

kubeclient.gemspec
ilackarms: @cben personally i like the way the current patch is designed. one option i would consider is yielding `chunks` to the passed block if `as: :raw` option is specified, so the caller can handle parsing on its own. however I'm a big fan of decoupling; we should allow kubeclient to abstract the kube API as much as possible from callers, so they do not have to concern themselves with "low level" things like parsing JSON, specifying API endpoints, etc. just my $0.02

ilackarms:
@abonas can this be re-reviewed?

lib/kubeclient/common.rb
cben: I see.  Agreed, rewriting everything to use `execute` is riskier change, might eventually be worth it, but not for this PR.
:+1:

lib/kubeclient/common.rb
```diff
@@ -218,7 +218,26 @@ def self.underscore_entity(entity_name)
...
+    def do_rest_request(args)
+      uri = @api_endpoint.merge("#{@api_endpoint.path}/#{@api_version}/#{args[:path]}")
       options = {
+        url: uri.to_s,
+        method: args[:method],
+        headers: (args[:headers] || {}).merge('params' => args[:params])
+      }.merge(common_options)
+      options[:block_response] = args[:block] if args[:block]
+      resp = RestClient::Request.execute(options)
+      return unless resp.instance_of?(Net::HTTPClientError)
```
cben: Not returning the response?
Does Request.execute return a Net::HTTPClientError outside of block_response mode?
=> Both OK as it's the only use case for this method. I'd make it unconditionally expect a block.

[style: early return usually used for problems, with flat fallthrough to happy path.  doesn't matter.]

lib/kubeclient/common.rb
```diff
@@ -291,6 +310,48 @@ def get_entities(entity_type, klass, resource_name, options = {})
...
+      resource_version = nil
+      entity_streamer = Json::Streamer::JsonStreamer.new
+      resource_version_streamer = Json::Streamer::JsonStreamer.new
+
+      resource_version_streamer.get(key: 'resourceVersion') do |value|
+        resource_version = value.to_i unless resource_version
+      end
+
+      # result['items'] might be nil due to https://github.com/kubernetes/kubernetes/issues/13096
+      entity_streamer.get(nesting_level: 2, yield_values: false) do |object|
```
cben: again, can this be narrowed to only 'items'?  if k8s adds another top-level array, it will confuse us.

lib/kubeclient/common.rb
cben: you're now checking both NotFound here and any error in `do_rest_request`.
can you check in one place, e.g. using
`error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError`

test/test_kubeclient.rb
```diff
@@ -209,7 +209,59 @@ def test_nonjson_exception_raw
...
+    end
+    assert(exception.message.include?('Not Found'))
+    assert_equal(404, exception.error_code)
+  end
+
+  def test_entity_list_with_block_http_err
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
+    stub_request(:get, %r{/services})
+      .to_return(body: '{"err": "I\'m a teapot!"}', status: 418)
```
cben: :laughing: :tea:

lib/kubeclient/common.rb
ilackarms: @cben I haven't tested it but I expect that it would; I don't think `block_response` changes the behavior in this regard.

ilackarms: @cben i originally used an `if` block to check for error and fallthrough on the happy path, and rubocop complained that i should replace the `if` block with a guard clause...

lib/kubeclient/common.rb
ilackarms: i tried doing this and discovered (what i believe is) a bug with `json-streamer`: https://github.com/thisismydesign/json-streamer/issues/7

though the bug possibly resides with `json-stream` itself. it may require forking `json-streamer` or re-implementing some of its features as mentioned above :(

kubeclient.gemspec
thisismydesign: @ilackarms @abonas @cben @jrafanie @kirs @moolitayer

Hi fellow developers,

I'm the creator and maintainer of the above mentioned [json-streamer](https://github.com/thisismydesign/json-streamer) gem.

Just wanted to let you know that I'm actively keeping an eye on my open source projects. This one did not require much attention lately hence the last release date is May 2017. I will take a look at the issue you raised. If it's reproducible it shouldn't take long to provide a fix.

Regarding features: IMO it fills a big hole that the currently available JSON parsers leave open and does that in a neat way. I found no similar implementation when I started but [json_stream_trigger](https://github.com/bdevel/json_stream_trigger) is actually a good catch. However it doesn't seem to be maintained. I will consider supporting JSONPath in the future.

Regarding community: I know quite a few use it and the number increases steadily since it's likely the best option out there. I'm not sure though how easy it is to find out about it. Since @dgraham [decided](https://github.com/dgraham/json-stream/issues/4#issue-224052592) not to mention it as an alternative to [json-stream](https://github.com/dgraham/json-stream/) it's not really advertised.

Regarding maintenance: I'm open to move the gem to an organization if you have any suggestions. For now all you have is my word. :)

lib/kubeclient/common.rb
cben: Seems that was fixed and you can bump the dependency?
Need not block this, if still in flux/untested, I'm fine with merging this and improving later.

kubeclient.gemspec
ilackarms: update: it looks like we will *not* be able to use `json-streamer` to do what we want here. it's currently possible to yield objects one-by-one from the `items` array in a kubernetes entity list, under the assumption that `items` is the only top-level array found in the JSON structure. it's not currently possible to specify the array by key name (see discussion at https://github.com/thisismydesign/json-streamer/issues/7#issuecomment-329816147).

thus we have two options: implement our own json-streamer type of gem, or accept the workaround which would break if Kubernetes change the JSON structure of *list objects.

i would be in favor of writing our own gem, but this will be additional tech debt for us to maintain

lib/kubeclient/common.rb
ilackarms: unfortunately, we can't just ask for items specifically: https://github.com/thisismydesign/json-streamer/issues/7#issuecomment-329816147

`nesting_level:2` is the only way to return object-by-object using json-streamer. it may be prudent to write our own implementation at this point (to prevent such a regression)

thisismydesign: I think rather this one is the applicable comment: https://github.com/thisismydesign/json-streamer/issues/7#issuecomment-330229144

kubeclient.gemspec
thisismydesign: Thought of a solution since: https://github.com/thisismydesign/json-streamer/issues/7#issuecomment-330254816

lib/kubeclient/common.rb
```diff
@@ -455,6 +444,78 @@ def self.restclient_read_timeout_option
...
     private
 
+    def do_rest_request(args)
+      uri = @api_endpoint.merge("#{@api_endpoint.path}/#{@api_version}/#{args[:path]}")
+      options = {
+        url: uri.to_s,
+        method: args[:method],
+        headers: (args[:headers] || {}).merge('params' => args[:params])
+      }.merge(rest_client_options)
+      options[:block_response] = args[:block] if args[:block]
```
cben: nitpick: To avoid confusion let's call it `block_response` everywhere.

Because `Request.execute` supports both regular ruby block and a `:block_response` param, with different semantics (regular block gets RestClient::Response after redirect/error processing/decoding).
https://github.com/rest-client/rest-client/blob/v2.0.2/lib/restclient/request.rb#L718-L726

lib/kubeclient/common.rb
cben: Default redirect/error handling is implemented by `Response`, specifically in https://github.com/rest-client/rest-client/blob/v2.0.2/lib/restclient/abstract_response.rb
Perhaps we might want to use it for all non-2xx statuses, like done here:
https://github.com/rest-client/rest-client/blob/v2.0.2/lib/restclient/request.rb#L802-L803

I'm also OK with raising on any non-2xx.

BTW, another processing we're skipping is Content-encoding.  What happens when k8s will enable compression?
(Note `Request.decode` is gone in upcoming rest-client 2.1: https://github.com/rest-client/rest-client/pull/597)

kubeclient.gemspec
ilackarms: @cben appears to be working now. updated! :+1:

lib/kubeclient/common.rb
ilackarms: @cben implemented; tell me what you think of the change

lib/kubeclient/common.rb
ilackarms: good question; it looks like RestClient automatically handles decompression: https://github.com/rest-client/rest-client/blob/f450a0f086f1cd1049abbef2a2c66166a1a9ba71/spec/integration/httpbin_spec.rb#L89

lib/kubeclient/common.rb
ilackarms: fixed!

lib/kubeclient/common.rb
ilackarms: updated

ilackarms: yes; if json-streamer is updated it would be a nice improvement. however at this point it doesnt seem like much overhead, as the 2 other streamers will be 'deactivated' once they match on a single key (and for the time being, those keys are present at the  start of the json object we're parsing).

ilackarms:
can i get some help with the rubocop failures here?

```
lib/kubeclient/common.rb:477:5: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for get_entities_async is too high. [9/8]
    def get_entities_async(klass, resource_name, options = {})
    ^^^

lib/kubeclient/common.rb:477:5: C: Metrics/PerceivedComplexity: Perceived complexity for get_entities_async is too high. [9/7]
    def get_entities_async(klass, resource_name, options = {})
    ^^^
```

seem difficult to fix. i'm not using a lot of conditionals; i don't see a clean way to fix these complaints.

```
lib/kubeclient/common.rb:488:57: W: Lint/UnusedBlockArgument: Unused block argument - object. If it's necessary, use _ or _object as an argument name to indicate that it won't be used.
      conditions.yield_object = lambda do |aggregator:, object:|
                                                        ^^^^^^
```

changing the naming of or remobing this variable causes `json-streamer` to fail parsing. not sure if it's worth opening an issue and requesting changes there in order to make this rubocop failure go away

lib/kubeclient/common.rb
ilackarms: `handle_exception` tries to parse `e.response` assuming it's a string. `RestClient::NotFound.response` is the HTTP response object. Unless we change the api of that method, raising our own `RestClient::NotFound` causes the error:

```
Message: <"no implicit conversion of Net::HTTPNotFound into String">
---Backtrace---
/home/ilackarms/.rbenv/versions/2.3.0/lib/ruby/2.3.0/json/common.rb:156:in `initialize'
/home/ilackarms/.rbenv/versions/2.3.0/lib/ruby/2.3.0/json/common.rb:156:in `new'
/home/ilackarms/.rbenv/versions/2.3.0/lib/ruby/2.3.0/json/common.rb:156:in `parse'
/home/ilackarms/workspace/manageiq/kubeclient/lib/kubeclient/common.rb:114:in `rescue in handle_exception'
/home/ilackarms/workspace/manageiq/kubeclient/lib/kubeclient/common.rb:111:in `handle_exception'
/home/ilackarms/workspace/manageiq/kubeclient/lib/kubeclient/common.rb:516:in `get_entities_async'
/home/ilackarms/workspace/manageiq/kubeclient/lib/kubeclient/common.rb:259:in `get_entities'
/home/ilackarms/workspace/manageiq/kubeclient/lib/kubeclient/common.rb:180:in `block (2 levels) in define_entity_methods'
/home/ilackarms/workspace/manageiq/kubeclient/lib/kubeclient/common.rb:91:in `method_missing'
/home/ilackarms/workspace/manageiq/kubeclient/test/test_kubeclient.rb:242:in `block in test_entity_list_with_block_404'
```

enoodle:
@moolitayer @cben @abonas I see that this is held up due to the rubocop issues with the longer functions. It seems that the functions `get_entities_async` and some naming problem with `json-streamer`. are there other issues holding this ? Do we want to reopen this to split this function and  see if we  can do anything with the argument name?

moolitayer:
> Do we want to reopen this to split this function and see if we can do anything with the argument name?

That's up to you if you have a use case for this feature. Are you aware of https://github.com/abonas/kubeclient/pull/262 ?
Not the same functionality but might alleviate a similar problem.

cben:
Our use case in ManageIQ remains, to reduce peak memory usage processing large responses like /api/pods and /oapi/images.  (Our processing is item-by-item.)
However as Nick pointed out in #262, interleaving reading / parsing / processing may keep the server connection open for a long time, and it may be nicer to slurp the whole raw output into string or temp file, and then stream-parse however we want, outside of kubeclient.
I also wanted to experiment with avoiding RecursiveOpenStruct.
There was also an idea that simply popping items off the array kubeclient returns as we process them could be a significant win, as RecursiveOpenStruct starts reasonably compact and inflates as you access data inside it.
We don't have concrete numbers to compare these potential wins...

=> IMHO, kubeclient would benefit from a convenient stream parsing API, the API here is good, and I'd like to land this :+1:
But if anyone has concerns, #262 makes it easier to experiment outside kubeclient, so we could do that and come back with more data.

cben:
Just learned that k8s finally got response pagination/chunking. Opened #283.
We'll definitely want chunking (it has more benefits on server).
Do we want this too?  It'd be useful for older k8s that don't support chunking, but perhaps "use :raw and stream parse yourself" is good enough?

I'd like to understand first what API(s) we'd expose for chunking...

agrare:
That's really cool @cben !  I think chunking will cover our needs from the parser side, so we should do similar perf tests using json stream parser with chunking to see if there is any benefit.

As for the API it could be the same as it is here?

cben:
I'm going to close this for now.
We're getting chunking (#356) which is mostly better, and I feel the total code complexity isn't worth to justify doing both *in kubeclient*.
And "use `:raw` and stream parse yourself" sounds reasonable to me, until proven otherwise.

If I'm wrong, e.g. if someone really wants streaming for older k8s that don't do chunking, please speak up / reopen.

Issue #234, opened 2017-03-11
kubect like feature for delete resources.


hridyeshpant:
Kubectl delete -f <file_name> can delete all resource mentioned in input file.
i am creating deployment which have replica set,  the create API is working fine and creating all resources like replicaset with deployment.
But when i am using deleteAPI with same dataset, i am getting error and it seems i cant pass yml file to deleteAPI and can only pass name of the resource.  so in this cace i need to first delete the deployment and then replicaset,Is this correct, ? 

can we have same feature like kubectl have which take input data and delete all resource?

/Users//.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/uri/rfc3986_parser.rb:67:in `split': bad URI(is not URI?): https://api.kubetnets.us-west-2.dev.XYZcom/apis/extensions/v1beta1/namespaces/default/deployments/{:apiVersion=>"extensions/v1beta1", :kind=>"Deployment", :metadata=>{:name=>"springboot-docker-ecs-service", :version=>"73fd3ac505273c456254450330a9db6fc5c3a8f7", :namespace=>"default"}, :spec=>{:replicas=>2, :strategy=>{:type=>"RollingUpdate"}, :revisionHistoryLimit=>1, :selector=>{:matchLabels=>{:app=>"springboot-docker-ecs-service", :version=>"73fd3ac505273c456254450330a9db6fc5c3a8f7"}}, :template=>{:metadata=>{:labels=>{:app=>"springboot-docker-ecs-service", :version=>"73fd3ac505273c456254450330a9db6fc5c3a8f7"}}, :spec=>{:containers=>[{:name=>"springboot-docker-ecs-service", :image=>"docker-registry.com/library/specimen-springboot-docker-ecs-service:73fd3ac505273c456254450330a9db6fc5c3a8f7", :ports=>[{:containerPort=>8080}], :env=>[{"name"=>"APP_NAME", "value"=>"specimen-springboot-docker-ecs-service"},  {"name"=>"_DNS_ZONE", "value"=>"dev"}, {"name"=>"_ENVIRONMENT", "value"=>"dev"}, {"name"=>"_DEPLOYED_ENVIRONMENT", "value"=>"dev"}], :livenessProbe=>{:httpGet=>{:path=>"/buildInfo", :port=>8080}, :initialDelaySeconds=>45, :timeoutSeconds=>10}}], :imagePullSecrets=>[{:name=>"secreat"}]}}}} (URI::InvalidURIError)
[...]

PR #255, opened 2017-07-05
support for adding query options to a delete entities command

kamerblauwlicht:
a solution for #234

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -39,6 +39,12 @@ module Kubeclient
       'fieldSelector' => :field_selector
     }.freeze
 
+    DELETE_ARGUMENTS = {
+      'gracePeriodSeconds' => :grace_period_seconds,
+      'orphanDependents'   => :orphan_dependents,
+      'propagationPolicy'  => :propagation_policy
+    }.freeze
+
     WATCH_ARGUMENTS = { 'resourceVersion' => :resource_version }.merge!(SEARCH_ARGUMENTS).freeze
 
     attr_reader :api_endpoint
@@ -194,8 +200,9 @@ module Kubeclient
           get_entity(klass, entity.resource_name, name, namespace)
         end
 
-        define_singleton_method("delete_#{entity.method_names[0]}") do |name, ns = nil, query = {}|
-          delete_entity(entity.resource_name, name, ns, query)
+        # ns = namespace, opts = options
+        define_singleton_method("delete_#{entity.method_names[0]}") do |name, ns = nil, opts = {}|
+          delete_entity(entity.resource_name, name, ns, opts)
         end
 
         define_singleton_method("create_#{entity.method_names[0]}") do |entity_config|
@@ -297,10 +304,10 @@ module Kubeclient
     end
 
     def delete_entity(resource_name, name, namespace = nil, options = nil)
-      ns_prefix = build_namespace_prefix(namespace)
-
       params = {}
-      params.merge!(options[:query]) if options.include?(:query)
+      DELETE_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
+
+      ns_prefix = build_namespace_prefix(namespace)
 
       handle_exception do
         rest_client[ns_prefix + resource_name + "/#{name}"]
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -194,8 +194,8 @@ module Kubeclient
           get_entity(klass, entity.resource_name, name, namespace)
         end
 
-        define_singleton_method("delete_#{entity.method_names[0]}") do |name, namespace = nil|
-          delete_entity(entity.resource_name, name, namespace)
+        define_singleton_method("delete_#{entity.method_names[0]}") do |name, ns = nil, query = {}|
+          delete_entity(entity.resource_name, name, ns, query)
         end
 
         define_singleton_method("create_#{entity.method_names[0]}") do |entity_config|
@@ -296,11 +296,15 @@ module Kubeclient
       new_entity(result, klass)
     end
 
-    def delete_entity(resource_name, name, namespace = nil)
+    def delete_entity(resource_name, name, namespace = nil, options = nil)
       ns_prefix = build_namespace_prefix(namespace)
+
+      params = {}
+      params.merge!(options[:query]) if options.include?(:query)
+
       handle_exception do
         rest_client[ns_prefix + resource_name + "/#{name}"]
-          .delete(@headers)
+          .delete({ 'params' => params }.merge(@headers))
       end
     end
 
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -39,12 +39,6 @@ module Kubeclient
       'fieldSelector' => :field_selector
     }.freeze
 
-    DELETE_ARGUMENTS = {
-      'gracePeriodSeconds' => :grace_period_seconds,
-      'orphanDependents'   => :orphan_dependents,
-      'propagationPolicy'  => :propagation_policy
-    }.freeze
-
     WATCH_ARGUMENTS = { 'resourceVersion' => :resource_version }.merge!(SEARCH_ARGUMENTS).freeze
 
     attr_reader :api_endpoint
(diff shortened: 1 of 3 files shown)


lib/kubeclient/common.rb
```diff
@@ -296,11 +296,15 @@ def get_entity(klass, resource_name, name, namespace = nil)
       new_entity(result, klass)
     end
 
-    def delete_entity(resource_name, name, namespace = nil)
+    def delete_entity(resource_name, name, namespace = nil, options = nil)
       ns_prefix = build_namespace_prefix(namespace)
+
+      params = {}
+      params.merge!(options[:query]) if options.include?(:query)
```
moolitayer: We do things a little differently in get_entitites(pass only the relevant options and not a nested hash), please take a look.

why do we currently only support a limited set of url parameters, and not pass in every thing that we get?
(get a 'params' hash and pass that in)
thoughts @simon3z @abonas @cben ?

kamerblauwlicht: yes. make sense. we can create a list of allowed queries like the get_entitites and use the same format.

it would be: 
gracePeriodSeconds = `grace_period_seconds`
orphanDependents = `orphan_dependents`
and
propagationPolicy = `propagation_policy`

if agree, I can change the code.

kamerblauwlicht:
I followed the same conventions as get_entities. let me know if we are good to go. I can add documentation if the code is ok.

lib/kubeclient/common.rb
```diff
@@ -296,11 +303,15 @@ def get_entity(klass, resource_name, name, namespace = nil)
       new_entity(result, klass)
     end
 
-    def delete_entity(resource_name, name, namespace = nil)
+    def delete_entity(resource_name, name, namespace = nil, options = nil)
+      params = {}
+      DELETE_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
```
cben: :+1:, `delete_entity(..., grace_period_seconds: 30)` is nicer interface than
 `delete_entity(..., query: {grace_period_seconds: 30})`.  

I'd like to start raising errors for unexpected options.  (Specific example I've stumbled on: hash with `:namespace` works in some methods, while others expect it positional and *silently* ignore `:namespace`, giving you little clue you called it wrong...)
But this would be a wider change, not for this PR.

test/test_deployment.rb
```diff
@@ -0,0 +1,24 @@
...
+class TestService < MiniTest::Test
+  def test_delete_a_deployment_with_propagation_policy_query
+    value_propagation_policy = 'Foreground'
+    namespace = 'my_namespace'
+    deployment = 'my_deployment'
+
+    stub_request(:get, %r{/api/extensions/v1beta1$})
+      .to_return(body: open_test_file('v1beta1_api_resource_list.json'), status: 200)
+    stub_request(:delete, %r{/namespaces/#{namespace}/deployments/#{deployment}})
+      .to_return(body: open_test_file('entity_list.json'), status: 200)
```
cben: Is this what DELETE actually returns?  https://kubernetes.io/docs/api-reference/v1.5/#delete-24 (clicking "curl" on top right) says it returns:
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Success",
  "code": 200
}
```

test/json/v1beta1_api_resource_list.json
```diff
@@ -0,0 +1,11 @@
+{
+  "kind": "APIResourceList",
+  "groupVersion": "extensions/v1beta1",
```
abonas: is the api version here up to date? v1beta1?

lib/kubeclient/common.rb
```diff
@@ -39,6 +39,12 @@ module ClientMixin
       'fieldSelector' => :field_selector
     }.freeze
 
+    DELETE_ARGUMENTS = {
+      'gracePeriodSeconds' => :grace_period_seconds,
+      'orphanDependents'   => :orphan_dependents,
```
abonas: this is a deprecated param starting 1.7. propagation policy should be used. moreover, it cannot be set together with propagationPolicy so shall it be supported here, there should be a validation they are not used together. (or at least verification that the server is validating this and returning a clear error)

test/test_deployment.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'extensions/v1beta1')
+    client.delete_deployment(deployment, namespace, propagation_policy: value_propagation_policy)
```
abonas: imo propagation_policy_value or propagation_policy_type is more clear. value somewhat tends to sound numeric.

test/json/v1beta1_api_resource_list.json
kamerblauwlicht: nope, I didn't add the whole API list. Just to make the test pass. I can create the total `v1beta1` list.

lib/kubeclient/common.rb
kamerblauwlicht: true.  I suggest focusing on `1.7. Shall I remove `orphanDependents` and we don't have this issue?

test/test_deployment.rb
kamerblauwlicht: hahaha. true... grammar ;-) will change.

test/test_deployment.rb
kamerblauwlicht: nope, didn't create a valid return json. will update. thanks.

lib/kubeclient/common.rb
abonas: I suggest to focus on the non deprecated param, unless there's a big demand for prior versions.
However it also depends on when propagation policy was introduced, on other words, will using it cause errors on earlier k8s versions?
@moolitayer @simon3z wdyt?

test/json/v1beta1_api_resource_list.json
abonas: I'm unsure what do you mean by creating the total v1beta1 list. what I meant is - isn't there a newer (non beta) api version for this?

lib/kubeclient/common.rb
cben: Is it wise for kubeclient to hardcode specific options of specific kube versions?
Could we just pass all options to server (after camelizing, something like `.gsub(/_([a-z])/) { |m| m[1].upcase }`)?
If a combination is incompatible and server doesn't give an error, that's IMO a server bug...

moolitayer: Yes, [I agree](https://github.com/abonas/kubeclient/pull/255#discussion_r125842680)
- I think we should still support search arguments we used to for compatibility
- I think we should pass a `params` hash to distinguish from other types of params
@abonas WDYT?

abonas: sure

test/json/v1beta1_api_resource_list.json
kamerblauwlicht: @abonas sorry about that. I checked the 1.7 api and deployments is still in `extensions\v1beta1`. it didn't get promoted.

lib/kubeclient/common.rb
kamerblauwlicht: ok. let me change some things.

carbonin:
bump @kamerblauwlicht @abonas any progress on this? I'm hoping it will solve my concern in https://github.com/abonas/kubeclient/issues/234#issuecomment-319749375

kamerblauwlicht:
didn't had time to touch this, will look at the right solution tomorrow.

kamerblauwlicht:
@abonas I made my changes, please review when you have time
@carbonin  changed made. wait when it's merged.

you can run a delete command like this:

`client.delete_deployment(deployment, namespace, propagationPolicy: 'Foreground')`

kamerblauwlicht:
Please hold this PR. It isn't working against 1.7.1 Kubernetes. Looks like the API doesn't know how to handle the parameters anymore.

kamerblauwlicht:
REFERENCE: 
https://kubernetes.io/docs/api-reference/v1.7/#deleteoptions-v1-meta
https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

kamerblauwlicht:
close this for a new one.

PR #257, opened 2017-07-14
Pod log enhancements

Ryan-Rose-Bose-zz:
- Add ability to specify timestamps option for get_pod_log and watch_pod_log
- Add ability to specify tailLines and secondsSince for watch_pod_log


lib/kubeclient/watch_stream.rb
```diff
@@ -72,6 +72,12 @@ def build_client_options
         else
           socket_option = :socket_class
         end
+        if @http_options[:sinceSeconds] then
```
moolitayer: `then`?

moolitayer:
Missing tests & README for the new options.
Could you please also add a link to kubernetes API @Ryan-Rose-Bose?

abonas:
there are 16 commits here, for a very small code change. (plus 13 lines, minus 3 lines)
I'm wondering why are there so many commits necessary.

lib/kubeclient/watch_stream.rb
Ryan-Rose-Bose: Newbie to Ruby, forgot it was inferred.

Ryan-Rose-Bose:
@moolitayer https://kubernetes.io/docs/api-reference/v1.6/#read-log , I can make some tests and add to the README.

@abonas there shouldn't have been, other than the fact that I was trying a few different things that needed to be tested in a deployed environment, so I had to push changes that didn't stay. Also a few typing inconsistencies, etc.

Ryan-Rose-Bose:
@moolitayer @abonas I added tests in the same form of `test_get_pod_log_container` and `test_watch_pod_log`, let me know if there's a better way to test functionality.

lib/kubeclient/version.rb
```diff
@@ -1,4 +1,4 @@
 # Kubernetes REST-API Client
 module Kubeclient
-  VERSION = '2.3.0'.freeze
+  VERSION = '2.3.1'.freeze
```
abonas: This PR adds new functionality, and as such, per semantic versioning, the minor version of the gem should be bumped, not the patch. (in other words, the middle digit, not the last one)
Also, not after every PR a gem version is bumped, so in most of the cases, the version is not bumped as part of the PR.

test/test_pod_log.rb
```diff
@@ -47,4 +62,64 @@ def test_watch_pod_log
...
+    stub_request(:get, %r{/namespaces/default/pods/[a-z0-9-]+/log\?.*follow})
+      .to_return(body: open_test_file('pod_log.txt'),
+                 status: 200)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+
+    stream = client.watch_pod_log('redis-master-pod', 'default', sinceSeconds: 3600)
+    stream.to_enum.with_index do |notice, index|
+      assert_instance_of(String, notice)
+      assert_equal(expected_lines[index], notice)
```
abonas: how does this test that the parameter sinceSeconds  3600 is addressed correctly? maybe I missed something and it does test it, but the test looks the same as the one that tests tail Lines

lib/kubeclient/version.rb
Ryan-Rose-Bose: I just put it back to `2.3.0`, my mistake.

test/test_pod_log.rb
Ryan-Rose-Bose: In:
```ruby
assert_requested(:get,
                     'http://localhost:8080/api/v1/namespaces/default/pods/redis-master-pod/log?follow=true&sinceSeconds=3600',
                     times: 1)
```
which checks the parameter properly made it into the API call. It's modeled after `test_get_pod_log_container`

Ryan-Rose-Bose: @abonas let me know if I can do anything to make this better :)

simon3z:
@Ryan-Rose-Bose I think you could squash the commits in 1 (unless you think it's still premature because we're far from finalizing this).

cben:
Huh.  Turns out @holyketzer reinvented pretty much this same wheel in the recently merged #319 :-)

@Ryan-Rose-Bose-zz I'm closing this for now, but if you want to add something #319 missed (notably it didn't extend `watch_pod_log`), you're welcome to rebase & reopen this, or open a new PR.

PR #258, opened 2017-07-14
Streaming pod logs with Web-servers

holyketzer:
Investigating this issue a couple of days. All details described in readme


abonas:
is this mostly a documentation PR? I saw some code fix - is this a bug fix?

holyketzer:
Yes mostly documentation, this is a fix which required to release resources as described in readme, else it will require some additional efforts from developer. It is the easiest way to put `ensure` block in `each`.

holyketzer:
@abonas have you used watchers inside of Web-servers? How you release resources in such case?

abonas:
> @abonas have you used watchers inside of Web-servers? How you release resources in such case?

I have not, perhaps @simon3z is more familiar with that part

README.md
```diff
@@ -438,6 +439,73 @@ watcher.each do |line|
...
+        # Allow watcher thread to execute
+        watcher_thread.join(STREAMING_PING_INTERVAL)
+
+        # Check that conection is alive
+        response.stream.write("ping\n")
+      end
+    end
+  end
+
+  def handle_stream_errors(response, thread, &block)
```
simon3z: @holyketzer I see two `handle_stream_errors`, how they are used? Maybe I am missing something but I have the feeling that only one is used.

lib/kubeclient/watch_stream.rb
```diff
@@ -29,6 +29,8 @@ def each
         end
       rescue IOError
         raise unless @finished
+      ensure
+        finish
```
simon3z: @holyketzer should we now protect the `@http_client.close` call below in case it happens on an already closed connection?

simon3z:
> @abonas have you used watchers inside of Web-servers? How you release resources in such case?

@holyketzer no we haven't. Anyway you may want to look at an implementation using websockets (if you haven't yet), which I have the feeling should facilitate the implementation (they're designed to be long-standing streaming connections).

README.md
holyketzer: They are used both. It's just DRY in error handling. Special thread read logs from Kubernetes, read can be blocked, but thread can't stopped. Web-server worker thread write pings into stream. In case of web-client disconnecting `response.stream.write` will fail and we should clise stream and stop thread to release resource, it can happen in both places.

lib/kubeclient/watch_stream.rb
holyketzer: If it's possible, it's better to do it

holyketzer:
Make sense, but the issue is in the way of getting logs from Kubernetes with blocking socket which I described in readme. There is no matter how you will send logs from web-server with websockets or streaming, important thing is that you get this logs with blocking and you should release resources when web-server client is disconnected.

README.md
```diff
@@ -438,6 +439,73 @@ watcher.each do |line|
...
+      end
+    end
+
+    handle_stream_errors(response, watcher_thread) do
+      loop do
+        # Allow watcher thread to execute
+        watcher_thread.join(STREAMING_PING_INTERVAL)
+
+        # Check that conection is alive
+        response.stream.write("ping\n")
```
simon3z: @holyketzer isn't this polluting the response with `ping` strings?

holyketzer: @simon3z right, but else web-server can't detect if client is disconnected.

simon3z: @holyketzer of course servers can detect if a client disconnects, the socket gets closed.

PR #262, opened 2017-08-20
Adds raw option for client get requests

NickLaMuro:
Prevents parsing of response body into `Kubeclient` objects.  In large kubernetes cluster environments where it would saving off the raw json for later use can useful for memory tight scenarios.

Usage:
------

```ruby
client = Kubeclient::Client.new(KUBECIENT_URL, 'v1')

client.get_endpoints                                        # normal
client.get_endpoints :raw => true                           # raw

client.get_endpoints "default_endpoint"                     # normal
client.get_endpoints "default_endpoint", nil, :raw => true  # raw
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,14 @@ service = client.get_service "guestbook", 'development'
 Note - Kubernetes doesn't work with the uid, but rather with the 'name' property.
 Querying with uid causes 404.
 
+#### Getting raw responses
+By passing `as: :raw`, the response from the client is given as a string, which is the raw JSON body from openshift:
+
+```ruby
+pods = client.get_pods as: :raw
+node = client.get_node "127.0.0.1", as: :raw
+```
+
 #### Delete an entity (by name)
 
 For example: `delete_pod "pod name"` , `delete_replication_controller "rc name"`, `delete_node "node name"`, `delete_secret "secret name"`
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -190,8 +190,8 @@ module Kubeclient
         end
 
         # get a single entity of a specific type by name
-        define_singleton_method("get_#{entity.method_names[0]}") do |name, namespace = nil|
-          get_entity(klass, entity.resource_name, name, namespace)
+        define_singleton_method("get_#{entity.method_names[0]}") do |name, ns = nil, opts = {}|
+          get_entity(klass, entity.resource_name, name, ns, opts)
         end
 
         define_singleton_method("delete_#{entity.method_names[0]}") do |name, namespace = nil|
@@ -239,12 +239,12 @@ module Kubeclient
       end
     end
 
-    # Accepts the following string options:
-    #   :namespace - the namespace of the entity.
-    #   :name - the name of the entity to watch.
-    #   :label_selector - a selector to restrict the list of returned objects by their labels.
-    #   :field_selector - a selector to restrict the list of returned objects by their fields.
-    #   :resource_version - shows changes that occur after that particular version of a resource.
+    # Accepts the following options:
+    #   :namespace (string) - the namespace of the entity.
+    #   :name (string) - the name of the entity to watch.
+    #   :label_selector (string) - a selector to restrict the list of returned objects by labels.
+    #   :field_selector (string) - a selector to restrict the list of returned objects by fields.
+    #   :resource_version (string) - shows changes that occur after passed version of a resource.
     def watch_entities(resource_name, options = {})
       ns = build_namespace_prefix(options[:namespace])
 
@@ -259,10 +259,14 @@ module Kubeclient
       Kubeclient::Common::WatchStream.new(uri, http_options(uri))
     end
 
-    # Accepts the following string options:
-    #   :namespace - the namespace of the entity.
-    #   :label_selector - a selector to restrict the list of returned objects by their labels.
-    #   :field_selector - a selector to restrict the list of returned objects by their fields.
+    # Accepts the following options:
+    #   :namespace (string) - the namespace of the entity.
+    #   :label_selector (string) - a selector to restrict the list of returned objects by labels.
+    #   :field_selector (string) - a selector to restrict the list of returned objects by fields.
+    #   :as (symbol) - if :raw, return the raw response body (as a string)
+    #
+    #   Default response type will return a collection RecursiveOpenStruct
+    #   (:ros) objects, unless `:as` is passed with `:raw`.
     def get_entities(entity_type, klass, resource_name, options = {})
       params = {}
       SEARCH_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
@@ -272,6 +276,7 @@ module Kubeclient
         rest_client[ns_prefix + resource_name]
           .get({ 'params' => params }.merge(@headers))
       end
+      return response.body if options[:as] == :raw
 
       result = JSON.parse(response)
 
@@ -286,12 +291,19 @@ module Kubeclient
       Kubeclient::Common::EntityList.new(entity_type, resource_version, collection)
     end
 
-    def get_entity(klass, resource_name, name, namespace = nil)
+    # Accepts the following options:
+    #   :as (symbol) - if :raw, return the raw response body (as a string)
+    #
+    #   Default response type will return an entity as a  RecursiveOpenStruct
+    #   (:ros) object, unless `:as` is passed with `:raw`.
+    def get_entity(klass, resource_name, name, namespace = nil, options = {})
       ns_prefix = build_namespace_prefix(namespace)
       response = handle_exception do
         rest_client[ns_prefix + resource_name + "/#{name}"]
           .get(@headers)
       end
+      return response.body if options[:as] == :raw
+
       result = JSON.parse(response)
       new_entity(result, klass)
     end
@@ -350,14 +362,14 @@ module Kubeclient
       klass.new(hash)
     end
 
-    def all_entities
+    def all_entities(options = {})
       discover unless @discovered
       @entities.values.each_with_object({}) do |entity, result_hash|
         # method call for get each entities
         # build hash of entity name to array of the entities
         method_name = "get_#{entity.method_names[1]}"
         begin
-          result_hash[entity.method_names[0]] = send(method_name)
+          result_hash[entity.method_names[0]] = send(method_name, options)
         rescue Kubeclient::HttpError
           next # do not fail due to resources not supporting get
         end
--- a/test/test_kubeclient.rb
+++ b/test/test_kubeclient.rb
@@ -193,6 +193,22 @@ class KubeclientTest < MiniTest::Test
     assert_equal(404, exception.error_code)
   end
 
+  def test_nonjson_exception_raw
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
+    stub_request(:get, %r{/servic})
+      .to_return(body: open_test_file('service_illegal_json_404.json'), status: 404)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+
+    exception = assert_raises(Kubeclient::ResourceNotFoundError) do
+      client.get_services(as: :raw)
+    end
+
+    assert(exception.message.include?('Not Found'))
+    assert_equal(404, exception.error_code)
+  end
+
   def test_entity_list
     stub_request(:get, %r{/api/v1$})
       .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
@@ -210,6 +226,34 @@ class KubeclientTest < MiniTest::Test
     assert_instance_of(Kubeclient::Service, services[1])
 
     assert_requested(:get, 'http://localhost:8080/api/v1/services', times: 1)
+  end
+
+  def test_entity_list_raw
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
+    stub_request(:get, %r{/services})
+      .to_return(body: open_test_file('entity_list.json'), status: 200)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+    response = client.get_services(as: :raw)
+
+    refute_empty(response)
+    assert_equal(open_test_file('entity_list.json').read, response)
+
+    assert_requested(:get, 'http://localhost:8080/api/v1/services', times: 1)
+  end
+
+  def test_entity_list_raw_failure
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
+    stub_request(:get, %r{/services})
+      .to_return(body: open_test_file('entity_list.json'), status: 500)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+
+    exception = assert_raises(Kubeclient::HttpError) { client.get_services(as: :raw) }
+    assert_equal('500 Internal Server Error', exception.message)
+    assert_equal(500, exception.error_code)
   end
 
   def test_entities_with_label_selector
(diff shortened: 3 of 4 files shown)


moolitayer:
cc @ilackarms

ilackarms:
straightforward patch :+1: 

looks good to me. should we be checking to make sure we get `200 OK` before returning non-error? Or we expect that to be taken care of by the http client?

NickLaMuro:
Fixed cops, and added some more tests, specifically around error handling.

@ilackarms regarding your question, the `handle_exception` block that surrounds the request will create an exception in the same fashion as if `:raw` wasn't provided.  That said, I did a poor job of considering that myself when I did this originally, so good catch.  As mentioned above, added some tests to verify that this is in fact the case.

moolitayer:
Looks good, only missing a README note and should be good to go.
@simon3z @abonas PTAL

The new interface would be beneficial in watch as well (but can be implemented if/when it is needed)

NickLaMuro:
Woops, sorry for not consulting the other open issues before just shipping some code... I am a terrible opensource citizen... 😭 

> Looks good, only missing a README note and should be good to go.

Sounds good, I will add.

> The new interface would be beneficial in watch as well (but can be implemented if/when it is needed)

Don't think this would be too hard for me to add in this PR as well, if you want me to put it together.

test/test_kubeclient.rb
```diff
@@ -191,6 +191,14 @@ def test_nonjson_exception
 
     assert(exception.message.include?('Not Found'))
     assert_equal(404, exception.error_code)
+
+    # return the same with `raw: true`
```
grosser: prefer a separate test instead of making the existing more complicated

grosser:
looks reasonable ... only alternative I could think of would be `KubeClient.deserializer = -> {}` setter or option and then users can do raw / json / OpenStruct / OJ etc ...

lib/kubeclient/common.rb
```diff
@@ -259,10 +259,11 @@ def watch_entities(resource_name, options = {})
       Kubeclient::Common::WatchStream.new(uri, http_options(uri))
     end
 
-    # Accepts the following string options:
+    # Accepts the following options:
     #   :namespace - the namespace of the entity.
     #   :label_selector - a selector to restrict the list of returned objects by their labels.
     #   :field_selector - a selector to restrict the list of returned objects by their fields.
```
cben: worth clarifying the selectors are strings. examples from tests:
label_selector: "name=redis-master"
field_selector: "involvedObject.name=redis-master"

cben:
Code looks pretty good (did review test).

Let's see if we'd add couple other modes to "do less" that were on the table:
- Currently we parse whole JSON, extract `.items` array, wrap each in RecursiveOpenStruct.
- Could parse whole JSON, extract `.items` array, but skip structs.
- #254: Read whole response(?), streaming parse JSON, yield `.items` one by one, wrap structs.
  - Also useful to have streaming mode without structs.
- This PR: read whole response, return raw string.
- Can be useful to return "chunk iterator" or something like that for streaming raw read.

That's a big matrix :-)  I'm trying to think if these can be more modular / composable, and then perhaps kubeclient wouldn't need all of them.  Example: if we had "chunk iterator" mode, users could implement streaming JSON parse themselves.  Example (forgetting compatibility): if we had only streaming interfaces, getting array is simple matter of `.to_a`.

cben:
To clarify, I'm less concerned with implementation (these *can* be done one in term of other), but with clean API.
E.g. #254 follows common ruby idiom — iterate if block given, return array if not — so does not increase API "weight" IMO.  I don't actually think forcing user to `.to_a` would be a win.

lib/kubeclient/common.rb
NickLaMuro: yeah, I can do that.

test/test_kubeclient.rb
NickLaMuro: Sure.  Was just after a while, it seemed like I was not very DRY in the tests, but `¯\_(ツ)_/¯`

grosser:
yeah ... need describe blocks for that ...

NickLaMuro:
@cben I would argue that one thing with https://github.com/abonas/kubeclient/pull/254 implementation that I don't like is it always includes a JSON streamer, even if the user is just going to take the full response anyway.  This means that all requests now go through the stream parser, regardless if they take advantage of any benefits of it.

I don't really want to spend too much time discussing #254 in this PR, so I will leave it at that.  Effectively, what I am trying to include is an option that allows the user to implement their own interface to the raw data, but take advantage of the connection handling and API negotiation that Kubeclient provides.

Also, in doing all that, I was trying to only use the minimal amount of resources required to do that, and enable the user to create any one of the described matrix results you described outside of kubeclient.  I think what you are describing is maybe a nice end goal, but probably something that requires a significant refactoring and most likely a major version number bump.

test/test_kubeclient.rb
NickLaMuro: Addressed.

lib/kubeclient/common.rb
```diff
@@ -259,10 +259,11 @@ def watch_entities(resource_name, options = {})
...
     end
 
-    # Accepts the following string options:
-    #   :namespace - the namespace of the entity.
-    #   :label_selector - a selector to restrict the list of returned objects by their labels.
-    #   :field_selector - a selector to restrict the list of returned objects by their fields.
+    # Accepts the following options:
+    #   :namespace (string) - the namespace of the entity.
+    #   :label_selector (string) - a selector to restrict the list of returned objects by labels.
+    #   :field_selector (string) - a selector to restrict the list of returned objects by fields.
```
NickLaMuro: Just a note:  Had to remove the "their" from the above two lines to make rubocop happy (line length).

NickLaMuro:
@moolitayer Added the `:raw` option for the `watch_` requests as well.  Let me know if that is what you were thinking for that.

lib/kubeclient/common.rb
```diff
@@ -190,8 +190,8 @@ def define_entity_methods
         end
 
         # get a single entity of a specific type by name
-        define_singleton_method("get_#{entity.method_names[0]}") do |name, namespace = nil|
-          get_entity(klass, entity.resource_name, name, namespace)
+        define_singleton_method("get_#{entity.method_names[0]}") do |name, ns = nil, opts = {}|
```
abonas: why is `namespace` renamed to `ns` as part of this patch (and at all?)
through several code parts below it is called namespace, I don't see any reason to rename it in one place, and ns is not very readable imo anyway for newcomers to code.

lib/kubeclient/common.rb
```diff
@@ -253,16 +254,18 @@ def watch_entities(resource_name, options = {})
       uri = @api_endpoint.merge("#{@api_endpoint.path}/#{@api_version}/#{path}")
 
       params = {}
+      format = options.delete(:raw) ? :text : :json
```
abonas: please elaborate a bit on this part. deleting a passed option without visibility to whoever called the code?

moolitayer:
@NickLaMuro following @cben concerns on the API change we took the time to think on the future outputs we want to support and tried to come up with a plan in https://github.com/abonas/kubeclient/issues/263

We think it is a good way forward, take a look and let use know.

lib/kubeclient/common.rb
NickLaMuro: @abonas `rubocop` was yelling because the line was too long.  Open to a better way of handling this, but this was the best that I came up with.

lib/kubeclient/common.rb
NickLaMuro: I guess a delete isn't necessary.  Misread the code into thinking that everything passed into `options` would be passed as option to `WatchStream`, but it looks like there is a filtered list of options that are allowed.

I can change this from a delete to just a `[]` call if you would prefer, just didn't want `:raw` to pollute things in method calls farther down the stack (again, doesn't look like this would have been the case).

moolitayer:
@NickLaMuro can you elaborate a little about how you plan to use this?

moolitayer:
@NickLaMuro This PR is valuable as is. I ask that input parameter will be changed to accommodate additional values in the future. I think `:as => :raw` is the best name we have right now, you might have a better idea.

NickLaMuro:
> I think :as => :raw is the best name we have right now, you might have a better idea.

@moolitayer Nope, that works for me.  I can make the change.

moolitayer:
@cben @grosser any chance for an explicit ack once we have a new version?

grosser:
can do! using `as: :raw` also leaves the door open for more de-serializer choices later on :)

lib/kubeclient/common.rb
NickLaMuro: I changed this to not use a delete when addressing other changes:  https://github.com/abonas/kubeclient/pull/262#issuecomment-324698152

lib/kubeclient/common.rb
```diff
@@ -253,16 +254,18 @@ def watch_entities(resource_name, options = {})
       uri = @api_endpoint.merge("#{@api_endpoint.path}/#{@api_version}/#{path}")
 
       params = {}
+      format = options[:as] == :raw ? :text : :json
       WATCH_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
       uri.query = URI.encode_www_form(params) if params.any?
 
-      Kubeclient::Common::WatchStream.new(uri, http_options(uri))
+      Kubeclient::Common::WatchStream.new(uri, http_options(uri), format: format)
```
cben: perhaps this should be named `as:` too?

cben: Does passing this option have any effect?  I don't see you modified WatchStream.

lib/kubeclient/common.rb
```diff
@@ -253,16 +254,18 @@ def watch_entities(resource_name, options = {})
       uri = @api_endpoint.merge("#{@api_endpoint.path}/#{@api_version}/#{path}")
 
       params = {}
+      format = options[:as] == :raw ? :text : :json
```
cben: we've been throwing "json" as name for "native hashes/arrays, without recursiveopenstruct wrapping" but it's not a very good name.  it could well be understood to mean "string in JSON format" which is exactly what "raw" is...

moolitayer: I like the `as` suggestion. I think the value should be 'ros' since that's what we provide now as default and `json` is a value we could provide in the future. BTW Format is also something we could find ourselves adding in the feature e.g someone might want a yaml response from k8s

lib/kubeclient/common.rb
```diff
@@ -253,16 +254,18 @@ def watch_entities(resource_name, options = {})
...
 
-    # Accepts the following string options:
-    #   :namespace - the namespace of the entity.
-    #   :label_selector - a selector to restrict the list of returned objects by their labels.
-    #   :field_selector - a selector to restrict the list of returned objects by their fields.
+    # Accepts the following options:
+    #   :namespace (string) - the namespace of the entity.
+    #   :label_selector (string) - a selector to restrict the list of returned objects by labels.
+    #   :field_selector (string) - a selector to restrict the list of returned objects by fields.
+    #   :raw (boolean) - if true, return the raw response body (as a string)
```
moolitayer: this one has to be updated

lib/kubeclient/common.rb
moolitayer: Can we please do the watch changes in another PR @NickLaMuro?
That way we could get this one in faster.

lib/kubeclient/common.rb
```diff
@@ -239,12 +239,13 @@ def rest_client
...
-    #   :label_selector - a selector to restrict the list of returned objects by their labels.
-    #   :field_selector - a selector to restrict the list of returned objects by their fields.
-    #   :resource_version - shows changes that occur after that particular version of a resource.
+    # Accepts the following options:
+    #   :namespace (string) - the namespace of the entity.
+    #   :name (string) - the name of the entity to watch.
+    #   :label_selector (string) - a selector to restrict the list of returned objects by labels.
+    #   :field_selector (string) - a selector to restrict the list of returned objects by fields.
+    #   :resource_version (string) - shows changes that occur after passed version of a resource.
+    #   :raw (boolean) - if true, return the raw response body (as a string)
```
moolitayer: needs update

lib/kubeclient/common.rb
NickLaMuro: @moolitayer No problem for me.  Sorry I jumped the gun on that one, but it definitely can wait.  Will remove shortly.

lib/kubeclient/common.rb
NickLaMuro: Good catch, will update.

lib/kubeclient/common.rb
NickLaMuro: and this one too... 😞

test/test_kubeclient.rb
```diff
@@ -341,6 +385,74 @@ def test_get_all
     assert_instance_of(Kubeclient::ServiceAccount, result['service_account'][0])
   end
 
+  def test_get_all_raw
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list.json'), status: 200)
+
+    stub_request(:get, %r{/bindings})
+      .to_return(body: open_test_file('bindings_list.json'), status: 404)
```
cben: why 404?

NickLaMuro: ¯\\(°_o)/¯

I basically did a copy pasta from the test above, and the stubs are what was done there.  The 404 was something I questioned myself, but was unsure why it was there.  Just left it.

test/test_kubeclient.rb
```diff
@@ -341,6 +385,74 @@ def test_get_all
...
+      .to_return(body: open_test_file('service_account_list.json'), status: 200)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+    result = client.all_entities(as: :raw)
+    assert_equal(16, result.keys.size)
+
+    %w[
+      component_status config_map endpoint event limit_range namespace node
+      persistent_volume persistent_volume_claim pod replication_controller
+      resource_quota secret service service_account
```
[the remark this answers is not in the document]
NickLaMuro: Not opposed to the idea, though, as [mentioned above](https://github.com/abonas/kubeclient/pull/262/files#r135650388), this was basically a copy-pasta of the previous test, but using the `as: :raw` option.

It was one of those times where I wondered:

> do I refactor these tests as part of the PR, muddling the diff from what was changed and what was refactored, or do I copy for now and leave the refactoring to a separate effort?

I obviously choose the latter option, but if you would prefer, I can add a commit in here where I make a separate pass to clean up some of these "not-DRY" tests.

test/test_kubeclient.rb
cben: and why didn't it fail the test?  bindings is not in tested array below, so either drop this stub or add it?

test/test_kubeclient.rb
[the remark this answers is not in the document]
NickLaMuro: To follow up on that last answer, regarding your question specifically:

> can you use same array to DRY above `stub_request`s?

The part where I used the array was part that I changed from the above test, since the assertions there were invalid for this test.  The `stub_request` portion was a copy-paste, and didn't need to be adjusted, so I wanted to leave that part unchanged.

test/test_kubeclient.rb
```diff
@@ -403,6 +515,27 @@ def test_api_bearer_token_failure
     assert_equal(response, exception.response)
   end
 
+  def test_api_bearer_token_failure_raw
+    error_message =
+      '"/api/v1" is forbidden because ' \
+      'system:anonymous cannot list on pods in'
+    response = OpenStruct.new(code: 401, message: error_message)
```
cben: [unrelated to this PR]
looks like `code:` here is either unnecessary or should be 403.
I'm not quite sure why this test asks stub `.to_raise(Kubeclient::HttpError...)` instead of `.to_return(..., status: 401)`.
Similar in `test_api_bearer_token_failure` from which this was copied, and several other tests.

test/test_kubeclient.rb
NickLaMuro: This is where I disagree, but to do so, I have to take some guesses into what the original `test_get_all` test above this one was attempting to accomplish originally (again, I didn't write this, just copied, so I am definitely taking a guess here).

To start, `test_get_all` also doesn't test against the `bindings` result, and I think that is in part to why the `404` was added in the first place.  There is a portion of that method that actually skips over any errors in that might have occurred when fetching everything:

https://github.com/abonas/kubeclient/blob/30b8285/lib/kubeclient/common.rb#L362

That means that a `404` (or I assume even a `500`), could occur, but it would allow the rest of the requests to be fetched and the results that were successful returned normally.  Whether or not this is the _**#right™**_ implementation is another story, but discussing that is probably out of scope for this PR.

I wanted to make sure that when testing any of these methods where the only change was the addition of the `as: :raw`, that the functionality mirrored what was being tested previously, and the only change was the output format.  Hence why most of these tests are carbon copies of tests that came prior to them, but have different assertions because the output differs.

Regarding making this DRY, as elaborated in my [other answer](https://github.com/abonas/kubeclient/pull/262#discussion_r135651199), I am fine doing this if that is what you would prefer, but didn't do so to start.  Waiting on a green light though before doing that.

test/test_kubeclient.rb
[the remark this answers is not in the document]
moolitayer: :+1: for suggesting a fix for this in another PR

simon3z:
@cben can you officially approve if you don't need any other changes? @ilackarms please review.

README.md
```diff
@@ -296,6 +296,14 @@ service = client.get_service "guestbook", 'development'
 Note - Kubernetes doesn't work with the uid, but rather with the 'name' property.
 Querying with uid causes 404.
 
+#### Get the raw response for an entity or entities
```
grosser: could simplify to `Getting raw responses`
and maybe add "To speed up processing or reduce memory consumption return raw JSON."

README.md
```diff
@@ -296,6 +296,14 @@ service = client.get_service "guestbook", 'development'
...
+By passing the `:as` option with type `:raw`, the response from the client is given as a string, which is the raw JSON body from openshift:
```
grosser: simplify to "by passing `as: :raw`"

NickLaMuro:
@grosser Good suggestions, and thanks for looking over the README additions.  Wasn't sure how to put things, so I like your changes.

test/test_kubeclient.rb
cben: agree to not refactor here, LGTM.
thanks for the detailed explanation, I didn't realize `all_entities` ignores errors, it makes sense now.
sorry for side-tracking this.

NickLaMuro: > sorry for side-tracking this.

No worries.  As I said, I had the same thought as you regarding the refactoring initially, but came to this conclusion for the time being.  I will probably open a PR just for some test refactoring regarding some of the duplicate code I generated after this gets merged 😄

lib/kubeclient/common.rb
```diff
@@ -259,10 +259,14 @@ def watch_entities(resource_name, options = {})
...
-    #   :label_selector - a selector to restrict the list of returned objects by their labels.
-    #   :field_selector - a selector to restrict the list of returned objects by their fields.
+    # Accepts the following options:
+    #   :namespace (string) - the namespace of the entity.
+    #   :label_selector (string) - a selector to restrict the list of returned objects by labels.
+    #   :field_selector (string) - a selector to restrict the list of returned objects by fields.
+    #   :as (symbol) - if :raw, return the raw response body (as a string)
+    #
+    #   Default response type will return a collection RecursiveOpenStruct
+    #   (:ros) objects, unless `:as` is passed with `:raw`.
```
moolitayer: in the future when/if we will have more values we will probably want to have a list here
```
:as (symbol) -
  :ros [default] ...
  :raw ...
```

This is good for this stage.

moolitayer:
LGTM :+1: 
cc @simon3z @ilackarms

moolitayer:
@NickLaMuro can you restart Travis

NickLaMuro:
@moolitayer Sorry for the delay.  Am on PTO today, but made a quick commit message update and re-pushed the last commit.  Hopefully that causes the tests to pass.

moolitayer:
:tada: :cake:

stefanmb:
@simon3z @abonas Is it possible to release a new version of the gem with this patch? It would be super useful for me, thank you!

simon3z:
> @simon3z @abonas Is it possible to release a new version of the gem with this patch? It would be super useful for me, thank you!

@stefanmb let's continue on https://github.com/abonas/kubeclient/pull/271

Issue #266, opened 2017-09-11
Passing extra query params when deleting an entity? (i.e. a deployment)


zeroDivisible:
Hi folks,

I had recently started using `kubeclient` as a library for abstracting communication with `Kubernetes` and I'm wondering - does this gem allow specifying extra query params when deleting an entity using Kube API?

The use case which I have is that I want to delete a deployment (`extensions/v1beta` API), but I want to pass the `propagationPolicy` param, otherwise, k8s leaves the ReplicaSet and the pods after deleting the Deployment object (basically, I want a cascade delete and need to instruct k8s to do so).

This is documented in https://kubernetes.io/docs/api-reference/v1.7/#delete-603 (`queryParameters` section).

Is this something which is supported / planned to be supported by this gem? If it's not supported yet, would incoming patches get accepted here?

simon3z:
cc @moolitayer @cben

cben:
Also discussed on https://github.com/abonas/kubeclient/issues/234#issuecomment-313212608 (but I think not a dup, that issue had wider goal to delete all resources from a file?).

=> #255 is open PR to support delete with query params.  Last update there was "Please hold this PR. It isn't working against 1.7.1 Kubernetes. Looks like the API doesn't know how to handle the parameters anymore."
@zeroDivisible could you see if #255 works for you / how to do it on most recent kubernetes?

zeroDivisible:
Thanks for the updates @cben I'll give those a read and check the work carried there.

kamerblauwlicht:
@cben @zeroDivisible @simon3z @moolitayer 

I'm going to work on the PR to make this work. Probably tomorrow there will be a fix using a body payload instead of query parameters because of the depreciation in `1.7.x`

kamerblauwlicht:
REFERENCE: https://kubernetes.io/docs/api-reference/v1.7/#deleteoptions-v1-meta

kamerblauwlicht:
implementation done. check my PR. 

@cben @zeroDivisible @simon3z @moolitayer

zeroDivisible:
A bit late to the party and I didn't do a full verification, but following the changes in the README, I was able to successfully verify that the code from @kamerblauwlicht 's  PR does indeed trigger the recursive delete (at least in my scenario). 👍 from me.

kamerblauwlicht:
@zeroDivisible good to hear. we have my code running in production ;-) and all good. hope we can merge the code soon.

abonas:
[...]

PR #267, opened 2017-09-13
cascade delete entities with a delete_option parameter.

kamerblauwlicht:
fix for #266

@cben @zeroDivisible @simon3z @moolitayer

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,20 @@ Input parameter - name (string) specifying service name, pod name, replication c
 client.delete_service("redis-service")
 ```
 
+If you want to cascade delete, for example a deployment, you can use the `delete_options` parameter.
+
+```ruby
+deployment_name = 'redis-deployment'
+namespace = 'default'
+delete_options = Kubeclient::Resource.new(
+    apiVersion: 'meta/v1',
+    gracePeriodSeconds: 0,
+    kind: 'DeleteOptions',
+    propagationPolicy: 'Foreground' # Orphan, Foreground, or Background
+)
+client.delete_deployment(deployment_name, namespace, delete_options: delete_options)
+```
+
 #### Create an entity
 For example: `create_pod pod_object`, `create_replication_controller rc_obj`, `create_secret secret_object`, `create_resource_quota resource_quota_object`, `create_limit_range limit_range_object`, `create_persistent_volume persistent_volume_object`, `create_persistent_volume_claim persistent_volume_claim_object`, `create_service_account service_account_object`
 
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -190,12 +190,14 @@ module Kubeclient
         end
 
         # get a single entity of a specific type by name
-        define_singleton_method("get_#{entity.method_names[0]}") do |name, ns = nil, opts = {}|
-          get_entity(klass, entity.resource_name, name, ns, opts)
+        define_singleton_method("get_#{entity.method_names[0]}") \
+        do |name, namespace = nil, opts = {}|
+          get_entity(klass, entity.resource_name, name, namespace, opts)
         end
 
-        define_singleton_method("delete_#{entity.method_names[0]}") do |name, namespace = nil|
-          delete_entity(entity.resource_name, name, namespace)
+        define_singleton_method("delete_#{entity.method_names[0]}") \
+        do |name, namespace = nil, opts = {}|
+          delete_entity(entity.resource_name, name, namespace, opts)
         end
 
         define_singleton_method("create_#{entity.method_names[0]}") do |entity_config|
@@ -308,11 +310,20 @@ module Kubeclient
       new_entity(result, klass)
     end
 
-    def delete_entity(resource_name, name, namespace = nil)
+    def delete_entity(resource_name, name, namespace = nil, delete_options: {})
+      delete_options_hash = delete_options.to_hash
       ns_prefix = build_namespace_prefix(namespace)
-      handle_exception do
-        rest_client[ns_prefix + resource_name + "/#{name}"]
-          .delete(@headers)
+      payload = delete_options_hash.to_json unless delete_options_hash.empty?
+      _response = handle_exception do
+        rs = rest_client[ns_prefix + resource_name + "/#{name}"]
+        RestClient::Request.execute(
+          rs.options.merge(
+            method: :delete,
+            url: rs.url,
+            headers: { 'Content-Type' => 'application/json' }.merge(@headers),
+            payload: payload
+          )
+        )
       end
     end
 
--- a/test/test_replication_controller.rb
+++ b/test/test_replication_controller.rb
@@ -24,4 +24,29 @@ class TestReplicationController < MiniTest::Test
                      'http://localhost:8080/api/v1/namespaces/default/replicationcontrollers/frontendController',
                      times: 1)
   end
+
+  def test_delete_replicaset_cascade
+    stub_request(:get, %r{/api/v1$})
+      .to_return(body: open_test_file('core_api_resource_list.json'),
+                 status: 200)
+
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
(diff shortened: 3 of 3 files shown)


lib/kubeclient/common.rb
```diff
@@ -308,11 +308,20 @@ def get_entity(klass, resource_name, name, namespace = nil, options = {})
...
+    def delete_entity(resource_name, name, namespace = nil, delete_options = {})
+      delete_options_hash = delete_options.to_hash
       ns_prefix = build_namespace_prefix(namespace)
-      handle_exception do
-        rest_client[ns_prefix + resource_name + "/#{name}"]
-          .delete(@headers)
+      payload = delete_options_hash.to_json unless delete_options_hash.empty?
+      _response = handle_exception do
+        rs = rest_client[ns_prefix + resource_name + "/#{name}"]
+        RestClient::Request.execute(
```
cben: @ilackarms Please review — does Request.execute require any extra error handling (check status code)?
Or was that only a concern in your PR when used with `block_response`?

README.md
```diff
@@ -314,6 +314,20 @@ Input parameter - name (string) specifying service name, pod name, replication c
...
+```ruby
+deployment_name = 'redis-deployment'
+namespace = 'default'
+delete_options = Kubeclient::Resource.new(
+    apiVersion: 'meta/v1',
+    gracePeriodSeconds: 0,
+    kind: 'DeleteOptions',
+    propagationPolicy: 'Foreground' # Orphan, Foreground, or Background
+)
+client.delete_deployment(deployment_name, namespace, delete_options)
```
cben: It's a bit unfortunate `namespace` is a positional parameter in several methods, now adding more params sometimes requires callers to `foo(name, nil, ...)`.
IMHO it's better to add any new params as named params, so this would be:
```
client.delete_deployment(deployment_name, namespace, delete_options: delete_options)
```

[out of scope: it'd nice to eventually also support keyword `namespace:` everywhere.  Would require some careful args parsing so 2nd arg can be either namespace string/nil or an options hash, but would make API clenear.]

kamerblauwlicht: total agree @cben 
I wanted to change it but a lot of things broke  ;-(

kamerblauwlicht:
@cben changed to named params

lib/kubeclient/common.rb
ilackarms: @cben i don't think it has anything to do with `block_response`; `Request.execute` handles errors differently than `Resource.get`. I assume that's because RestClient makes assumptions about what status codes mean for a `Resource`, but not for a generic `Request` (e.g.: that a 404 should produce an error for a `GET` on a Resource, but a 404 might be a non-error status for a different type of request). `Request.execute` will simply return a response object without raising anything. 

Again I haven't tested it the other way, but my gut feeling is that `Request.execute` is similar to `curl`; as long as the remote server returns an HTTP response of some kind, no error will be raised.

kamerblauwlicht: @ilackarms  the method will raise an error when the resource is not found.

simon3z:
@cben @moolitayer @ilackarms PTAL

README.md
```diff
@@ -314,6 +314,20 @@ Input parameter - name (string) specifying service name, pod name, replication c
...
+If you want to cascade delete, for example a deployment, you can use the `delete_options` parameter.
+
+```ruby
+deployment_name = 'redis-deployment'
+namespace = 'default'
+delete_options = Kubeclient::Resource.new(
+    apiVersion: 'meta/v1',
+    gracePeriodSeconds: 0,
+    kind: 'DeleteOptions',
+    propagationPolicy: 'Foreground' # Orphan, Foreground, or Background
```
ilackarms: can we encapsulate these values with an enum/const type? e.g. `Kubeclient::PropagationPolicy::Foreground`?

cben: I want to differ, simple string LGTM here.
What if k8s later adds more values?  I don't see a benefit to kubeclient maintaining this set in sync with the k8s reality (especially if reality can differ by server version).

Also, if the values these consts will have are still the strings 'Foreground' etc, we can always add such consts later, need not block this PR.

lib/kubeclient/common.rb
cben: Confirmed experimentally.  Normal `Request.execute` follows redirects and raises on errors.  The high-level `get` and `Resource#get` are thin wrappers around it, no extra handling logic:
https://github.com/rest-client/rest-client/blob/v2.0.2/lib/restclient.rb#L66-L68
https://github.com/rest-client/rest-client/blob/v2.0.2/lib/restclient/resource.rb#L49-L55

It's only with `block_response` that [bypasses normal processing](https://github.com/rest-client/rest-client/blob/v2.0.2/lib/restclient/request.rb#L718-L726).

README.md
kamerblauwlicht: agree @cben

lib/kubeclient/common.rb
```diff
@@ -308,11 +308,20 @@ def get_entity(klass, resource_name, name, namespace = nil, options = {})
       new_entity(result, klass)
     end
 
-    def delete_entity(resource_name, name, namespace = nil)
+    def delete_entity(resource_name, name, namespace = nil, delete_options: {})
```
moolitayer: There is nothing specific here for delete options right? we should rename in that case

moolitayer:
I don't get why the example is passing a Kubeclient::Resource?
Can we use a hash in the example? 

I think the client and not the user should add `apiVersion` and ` kind: 'DeleteOptions'` to the options.
We set `apiVersion` in several other places.

:+1: for always using payload instead of query parameters for a DELETE request

README.md
ilackarms: OK that's fair

kamerblauwlicht:
@moolitayer I get your point. The example make clear it consume a resource and you (as the user) can change it. If we are going to keep that away from the user and set `apiVersion` and so on in the client we need to maintain this. IMO this is the way to keep up with the fast-moving changes in Kubernetes.

@abonas What is needed more to merge this puppy?

lib/kubeclient/common.rb
```diff
@@ -194,8 +194,8 @@ def define_entity_methods
           get_entity(klass, entity.resource_name, name, ns, opts)
         end
 
-        define_singleton_method("delete_#{entity.method_names[0]}") do |name, namespace = nil|
-          delete_entity(entity.resource_name, name, namespace)
+        define_singleton_method("delete_#{entity.method_names[0]}") do |name, ns = nil, opts = {}|
```
simon3z: @kamerblauwlicht we'd like to keep `namespace` for consistency. You can split the line to make rubocop happy. cc @NickLaMuro

kamerblauwlicht: understood, but I can't split the line when using a block. can you show me an example to split @simon3z

simon3z: @kamerblauwlicht check this https://github.com/abonas/kubeclient/pull/271#pullrequestreview-68545556
Since you're at it could you change also the other `ns` instance that was recently introduced in the code? :blush:

kamerblauwlicht: hahah I will @simon3z  ;-)

kamerblauwlicht:
@simon3z done

simon3z:
> @simon3z done

@kamerblauwlicht thanks! I was about to merge and then I noticed that it has some redundant commits. Do you think you can squash all in a single commit? (After all it's just one feature addition)

abonas:
please squash redundant commits so this can be merged. thanks!

mleklund:
I would _love_ to see this feature merged.

kamerblauwlicht:
sorry for the late response (on holiday). will squash the commits today ;-)

kamerblauwlicht:
@abonas squashed the commits. please merge.

README.md
grosser: 2-space indent ?

grosser:
great to see this getting fixed ... atm I'm manually deleting the pods ... which is a bit annoying :D

PR #269, opened 2017-09-18
Require rest-client 2.x, drop 1.x support code

### LINKED ISSUE #264: Specify rest-client version
While investigating some certificate verification errors, I discovered that we were inadvertently using rest-client v1.6.9, which doesn't support specifying custom certificate stores (this was added in [1.7.0](https://github.com/rest-client/rest-client/commit/f378ad154a6cc3a15c162552f6b1b08822885099)). Currently kubeclient [doesn't specify a rest-client version](https://github.com/abonas/kubeclient/blob/30b8285411ecfb485f7cfd816684936e32d46ba6/kubeclient.gemspec#L30), it would be helpful if it did for things like this. At least v1.7.0 is required to solve this issue, but perhaps there are other considerations which need an even newer version?

Ref: Shopify/kubernetes-deploy#157

cben:
Closes #264.

Made it `~> 2.0` — will allow any 2.y.z; I don't want to pin to 2.0.z without specific reasons as rest-client is widely used gem (and my impression is they're careful about compatibility so we can trust their semver)...
Tests pass locally against:

- [x] 2.0.0
- [x] 2.0.2 (current)
- [x] 2.1.0rc1


simon3z:
+1 on the effort, haven't reviewed in detail.
@moolitayer PTAL

lib/kubeclient/common.rb
```diff
@@ -441,18 +441,6 @@ def api
       JSON.parse(response)
     end
 
-    def self.restclient_read_timeout_option
```
moolitayer: :+1: :tada:

kubeclient.gemspec
```diff
@@ -27,7 +27,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'webmock', '~> 3.0.1'
   spec.add_development_dependency 'vcr'
   spec.add_development_dependency 'rubocop', '= 0.49.1'
-  spec.add_dependency 'rest-client'
+  spec.add_dependency 'rest-client', '>= 2.0.0'
```
moolitayer: As discussed I'm fine with both `~> 2.0.0` or `~> 2.0` here.

cben: `>= 2.0.0` was wrong as it would also accept 3.0.
Made it `~> 2.0` — will allow any 2.y.z; I don't want to pin to 2.0.z without specific reasons as rest-client is widely used gem...
Tests pass locally against:
- [x] 2.0.0
- [x] 2.0.2 (current)
- [x] 2.1.0rc1

and rest-client is generally careful about backward compat so I think we can trust their semver.

I can make travis test against rest-client master in parallel to 2.0, if anyone wants.

PR #271, opened 2017-10-10
[v2.x] Adds raw option for client get requests

simon3z:
@NickLaMuro @stefanmb @cben @moolitayer can you review this backport of the raw option to the v2.x branch?

It seemed straight-forward to me and backward compatible (but please double-check that). Let's see if there are any travis failures.

Ref: https://github.com/abonas/kubeclient/pull/262

Thanks.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -189,8 +189,8 @@ module Kubeclient
         end
 
         # get a single entity of a specific type by name
-        define_singleton_method("get_#{entity.method_names[0]}") do |name, namespace = nil|
-          get_entity(klass, entity.resource_name, name, namespace)
+        define_singleton_method("get_#{entity.method_names[0]}") do |name, ns = nil, opts = {}|
+          get_entity(klass, entity.resource_name, name, ns, opts)
         end
 
         define_singleton_method("delete_#{entity.method_names[0]}") do |name, namespace = nil|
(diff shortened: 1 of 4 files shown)


lib/kubeclient/common.rb
```diff
@@ -189,8 +189,8 @@ def define_entity_methods
         end
 
         # get a single entity of a specific type by name
-        define_singleton_method("get_#{entity.method_names[0]}") do |name, namespace = nil|
-          get_entity(klass, entity.resource_name, name, namespace)
+        define_singleton_method("get_#{entity.method_names[0]}") do |name, ns = nil, opts = {}|
+          get_entity(klass, entity.resource_name, name, ns, opts)
```
simon3z: @NickLaMuro the `namespace` => `ns` renaming is backward-compatible?

stefanmb: For consistency with the existing functions, this could be kept as `namespace` (instead of `ns`).

NickLaMuro: @simon3z @stefanmb This was done because of line length issues for rubocop.  Wasn't sure of a better way to do this, and I don't know how concerned you are with rubocop failures (but that did cause the travis run to fail without me doing this).

test/test_kubeclient.rb
```diff
@@ -483,6 +616,26 @@ def test_api_basic_auth_failure
...
+      'http://localhost:8080/api/',
+      auth_options: { username: 'username', password: 'password' }
+    )
+
+    exception = assert_raises(KubeException) { client.get_pods(as: :raw) }
+    assert_equal(401, exception.error_code)
+    assert_equal(error_message, exception.message)
+    assert_equal(response, exception.response)
+
+    assert_requested(:get, 'http://localhost:8080/api/v1', times: 1)
```
NickLaMuro: This test needs to check against the new URL you changed it to in line `623` (will fix tests):

```ruby
assert_requested(:get, 'http://username:password@localhost:8080/api/v1', times: 1)
```

lib/kubeclient/common.rb
```diff
@@ -189,8 +189,9 @@ def define_entity_methods
         end
 
         # get a single entity of a specific type by name
-        define_singleton_method("get_#{entity.method_names[0]}") do |name, namespace = nil|
-          get_entity(klass, entity.resource_name, name, namespace)
+        define_singleton_method("get_#{entity.method_names[0]}") \
+            do |name, namespace = nil, opts = {}|
```
simon3z: @NickLaMuro could you forward-port this so that master won't have to change to use `ns` instead of `namespace`?

test/test_kubeclient.rb
```diff
@@ -457,11 +457,11 @@ def test_get_all_raw
     result = client.all_entities(as: :raw)
     assert_equal(16, result.keys.size)
 
-    %w[
+    %w(
```
simon3z: @NickLaMuro would we want to forward-port this as well?

simon3z:
Waiting for @cben and @moolitayer ACK on backward-compatibility.

cben:
ack, LGTM.

PR #279, opened 2017-12-15
Recurse over arrays for watch notices

jhernand:
Currently the elements of arrays inside watch notices aren't converted
to intances of `RecursiveOpenStruct`, they are converted to instances of `Hash`
instead. This mean that accessing certain fields works differently for
objects retrieved directly than for objects received in a notice. For
example, to extract the internal IP address of a node retrieved directly
the code is like this:

```ruby
ip = node.status.addresses.detect { |x| x.type == 'InternalIP' }.address
```

But for a notice it is like this:

```ruby
ip = node.status.addresses.detect { |x| x['type'] == 'InternalIP' }['address']
```

This complicates code that needs to treat objects in the same way,
regardless of how they have been retrieved.

To avoid that this patch changes the `WatchNotice` class so that the
constructor adds the _RecursiveOpenStruct_ `:recurse_over_arrays` option.

### DIFF
--- a/test/test_watch_notice.rb
+++ b/test/test_watch_notice.rb
@@ -1,0 +1,37 @@
+require 'test_helper'
+
+class TestWatchNotice < MiniTest::Test
+  #
+  # Checks that elements of arrays are converted to instances of `RecursiveOpenStruct`, and that the
+  # items can be accessed using the dot notation.
+  #
+  def test_recurse_over_arrays
+    json = JSON.parse(open_test_file('node_notice.json').read)
+    notice = Kubeclient::Common::WatchNotice.new(json)
+    assert_kind_of(Array, notice.object.status.addresses)
+    notice.object.status.addresses.each do |address|
+      assert_kind_of(RecursiveOpenStruct, address)
+    end
+    assert_equal('InternalIP', notice.object.status.addresses[0].type)
+    assert_equal('192.168.122.40', notice.object.status.addresses[0].address)
+    assert_equal('Hostname', notice.object.status.addresses[1].type)
+    assert_equal('openshift.local', notice.object.status.addresses[1].address)
+  end
+
+  #
+  # Checks that even when arrays are converted to instances of `RecursiveOpenStruct` the items can
+  # be accessed using the hash notation.
+  #
+  def test_access_array_items_as_hash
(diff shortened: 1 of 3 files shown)


jhernand:
@abonas @simon3z @masayag please review.

simon3z:
@moolitayer @cben can you review?

cben:
:+1: good catch, recurse_over_arrays is desired and consistent with rest of kubeclient.

Is there any backward-compatibility concern?
I think not as a RecursiveOpenStruct is nearly drop-in compatible with a hash.

test/test_watch_notice.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_helper'
+
+class TestWatchNotice < MiniTest::Test
+  def test_recurse_over_arrays
+    json = JSON.parse(open_test_file('node_notice.json').read)
+    notice = Kubeclient::Common::WatchNotice.new(json)
+    assert_kind_of(Array, notice.object.status.addresses)
+    notice.object.status.addresses.each do |address|
+      assert_kind_of(OpenStruct, address)
```
moolitayer: s/OpenStruct/RecursiveOpenStruct/ 
Will be helpful to do the same in the commit message

moolitayer:
Look good, one nit. Agree with @cben this can be considered backward compatible

moolitayer:
Maybe we cab add a simple test for the array access compat?
Silly really since we all know it works but I would add one anyway

jhernand:
Thanks for your review @cben and @moolitayer. I have added the backwards compatibility test as you suggested.

jhernand:
@simon3z thanks for merging this. One can we expect a new release of the gem containing the fix?

simon3z:
> @simon3z thanks for merging this. When can we expect a new release of the gem containing the fix?

@jhernand you're welcome, the PR was reviewed and approved by @cben and @moolitayer so merging was just bookkeeping.

@abonas @bazulay any plans around the gem release?

abonas:
> @abonas @bazulay any plans around the gem release?

no plans from my side, I am not focusing at this gem development at the moment

jhernand:
@abonas who is currently responsible for doing releases of the gem?

abonas:
@jhernand I suggest asking @simon3z and @bazulay

Issue #282, opened 2018-01-09
support as: :raw for watch


grosser:
would accept a PR ?
same as https://github.com/abonas/kubeclient/pull/271
/cc @simon3z @stefanmb @cben

cben:
Sounds good to me.

moolitayer:
was included in https://github.com/abonas/kubeclient/issues/263
Related to https://github.com/abonas/kubeclient/pull/254

Issue #263, opened 2017-08-22
Extend get_entities interface with result_types and streaming VS async resonses


moolitayer:
Since we have a few[1][2] efforts going in the same direction that have implication on our external API
@cben @ilackarms and myself brainstormed to come up with
- minimal interface to allow most options
- implementation for the actual get_entities methods to accommodate most options
- what is the minimum we have to do now, what can we do later

### What do we want to allow?
Currently `get_entities` is synchronous and returns ROS wrapping JSON objects for a complete response after a request response cycle complete. These are all viable options:
1 synchronous get whole raw response ([2])
2 get each raw chunk when available
3 synchronous get whole JSON response
4 get each JSON object when available
5 synchronous get whole ROS response (current default)
6 get each ROS object when available ([1])

### How should our new interface look like?
1 `get_pods(:result_type[3] => :raw)`
2 `get_pods(:result_type => :raw) do |chunk|
...
end
`
3 `get_pods(:result_type => :json)`
4 `get_pods(:result_type => :json) do |json_pod|
...
end
`
5 `get_pods()` # :result_type => ros is our current default
6 `get_pods() do |ROS_pod|
...
end` # :result_type => ros is our current default

### What is the minimum we can do now?
- [1] internal get_entities always serves raw chunks since that is the lowest denominator, need to also accommodate current behavior - option 5 and option 6 if block is given
- [2] implement 1 and 2
- touch only get, not watch 

cc @ilackarms @cben 
@simon3z @abonas @NickLaMuro @grosser if you have no objection we like to proceed this way ASAP

[1] https://github.com/abonas/kubeclient/pull/254
[2] https://github.com/abonas/kubeclient/pull/262
[3] the name `:result_type` is ephemeral, might change in Code review

cben:
> internal get_entities always serves raw chunks since that is the lowest
denominator

Are `get_entity/ies`, `update_entity` considered private or part of public
API?
They're not in README, but also aren't `private` methods.
I actually thought of them as the "core API" and `get_pods` etc as
convenience additions.

moolitayer:
Interesting @cben I never thought about it that way but it is valid - I guess for sure we need to maintain api in both the dynamically created get_pods and static get_entities. I don't remember seeing much usage of get_entities in example code snippets we got from users though.

grosser:
sounds good to me ... result_type sounds a bit clunky to me ... maybe `as:` or `format:` would read nicer ... but I guess they could be confused with other things ...

[...]

PR #285, opened 2018-01-12
implement as: :raw for watch

grosser:
fixes https://github.com/abonas/kubeclient/issues/282
part of https://github.com/abonas/kubeclient/issues/263

@cben @moolitayer

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -247,6 +247,8 @@ module Kubeclient
     #   :label_selector (string) - a selector to restrict the list of returned objects by labels.
     #   :field_selector (string) - a selector to restrict the list of returned objects by fields.
     #   :resource_version (string) - shows changes that occur after passed version of a resource.
+    #   Default response type will return an entity as a  RecursiveOpenStruct
+    #   (:ros) object, unless `:as` is passed with `:raw`.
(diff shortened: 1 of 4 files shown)


lib/kubeclient/common.rb
```diff
@@ -247,6 +247,7 @@ def rest_client
     #   :label_selector (string) - a selector to restrict the list of returned objects by labels.
     #   :field_selector (string) - a selector to restrict the list of returned objects by fields.
     #   :resource_version (string) - shows changes that occur after passed version of a resource.
+    #   :as (symbol) - if :raw, return the raw response body (as a string)
```
moolitayer: :+1: for taking doc from get_entit* please also include the default clause
```
#   Default response type will return an entity as a  RecursiveOpenStruct
#   (:ros) object, unless `:as` is passed with `:raw`.
```

moolitayer:
Looks good. One nit comment, also please add a README note.

It could be in "Getting raw responses" section that  mentions raw for get_x or in/after "Receive entity updates" that talks about watch. I prefer in the watch section since that is probably what a developer writing watch code would go to.

grosser:
updated, looks good ?

moolitayer:
Labeled v2.x/yes

CC @cben @jhernand

PR #288, opened 2018-01-22
Fix README comment on Kubeclient::HttpError

moolitayer:
https://github.com/abonas/kubeclient/pull/195/files#diff-04c6e90faac2675aa89e2176d2eec7d8R425 introduced the new exception and @cben noticed the README note matches a name that was changed during review.

This change is not currently(2018-01-25) included in a release and it is planned for upcoming `v2.6.0` and `v3.0.0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,9 +476,9 @@ client.process_template template
 
 ## Upgrading
 
-#### past version 2.0
+#### past version 2.6
 
-Replace `KubeException` with `Kubeclient::HttpException`
+The gem raises Kubeclient::HttpError or subclasses now. Catching KubeException still works but is deprecated.
 
 #### past version 1.2.0
 Replace Specific Entity class references:


moolitayer:
@yaacov @zeari please take a look

yaacov:
LGTM, but don't you want to do actually release 5.6 before documenting the changes ?

moolitayer:
I provided some context. commit message updated.

When the new version comes out we want the comment already there. If we wanted to release a version < `2.6` we should have kept this out of master.

moolitayer:
@jhernand please review

cben:
This is an improvement :+1:, but I find the "Upgrading past 2.6" headers ambiguous.  It could be read as either:
- You MAY use the new API since 2.6
- You MUST use the new API since 2.6 (false, may happen in 3.0)

How about less prescriptive more informative wording, eg. "The gem raises Kubeclient::HttpError or subclasses now; catching KubeException still works but is deprecated."

We need an actual changelog.  Will it become an expansion of this section, or separate file and we'll also keep a section here with only important deprecations/breaking changes?

moolitayer:
@cben updated the wording please look.

I think for the change log should be in it's own file(`CHANGELOG.md`?). I would generate one and merge a pull request updating it prior to the release. Another question: Do we dump the all the non merge commits in there? for example ManageIQ's  changelog includes enhancements or bugfixes bug but not developer/test/cleanups.

jhernand:
I think that the change log should be useful for humans, mostly for humans that using the gem in their own projects. In these regards enhancements and bug fixes are relevant, clean-ups, refactorings, etc, are not.

README.md
```diff
@@ -476,9 +476,9 @@ client.process_template template
 
 ## Upgrading
 
-#### past version 2.0
+#### past version 2.6
 
-Replace `KubeException` with `Kubeclient::HttpException`
+The gem raises Kubeclient::HttpError or subclasses now. Catching KubeException still works but is deprecated
```
cben: [nit] period at the end please, and I'll merge.

PR #290, opened 2018-01-22
update delete and patch should return ROS

### LINKED ISSUE #196: update and delete should return proper objects
create and get return RecursiveObjectStruct ... update and delete return RestClient::Response ... not good ...

moolitayer:
Fix #196 

Public methods update_x delete_x and patch_x should all return RecursiveOpenStruct like get_x and watches.

I did not add the `:as => :raw` options in these cases because the response is small and limited so I do not expect to hit performance problems there. Anyway since the `:as` is an additional option it can be added in a minor version.

### DIFF
--- a/test/test_replication_controller.rb
+++ b/test/test_replication_controller.rb
@@ -41,9 +41,9 @@ class TestReplicationController < MiniTest::Test
     stub_request(:delete,
                  'http://localhost:8080/api/v1/namespaces/default/replicationcontrollers/frontendController')
       .with(body: opts.to_hash.to_json)
-      .to_return(status: 200, body: '', headers: {})
-
-    client.delete_replication_controller('frontendController', 'default', delete_options: opts)
+      .to_return(status: 200, body: open_test_file('replication_controller.json'), headers: {})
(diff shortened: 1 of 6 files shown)


abonas:
Can u please clarify in the commit message what does format mean here

moolitayer:
@abonas updated please take a look

moolitayer:
@grosser @kirs please @cben please review

abonas:
can you please explain why a delete method should return ROS?

moolitayer:
Why return anything?

Because the API server gives you the last state of the object (it will be an object in deletion) and we are throwing that away.

Why ROS?

Compatibility with other methods mostly.
I mentioned in the first comment why I did not add the :as option

test/test_replication_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 # Replication Controller entity tests
 class TestReplicationController < MiniTest::Test
-  def test_get_from_json_v1
+  def tes_get_from_json_v1
```
cben: typo (test -> tes)

cben:
@abonas any objections?

Issue #199, opened 2016-10-05
Kubeclient next: tracker of non backward compatible issues


moolitayer:
- https://github.com/abonas/kubeclient/issues/196
- https://github.com/abonas/kubeclient/issues/198
- https://github.com/abonas/kubeclient/issues/187#issuecomment-247860068
  (basically a Readme update getting rid of all Kubeclient::X classes)

grosser:
- remove KubeException https://github.com/abonas/kubeclient/pull/195

cben:
#240 probably (start or stop always wrapping SSL errors — either one may break code expecting current behavior, though current behavior is inconsistent.)

abonas:
@moolitayer have you considered tracking backward incompatible issues by labeling them directly as such? tracking them here is nice, but manual tracking is fragile and some things can slip away

moolitayer:
@abonas sounds great. Note I cannot label stuff ATM

abonas:
> @abonas sounds great. Note I cannot label stuff ATM

@moolitayer do you mean you cannot create new labels or you can't apply existing labels? I think you should be able to label things. Please check. I can create a new label, any preferred name?

moolitayer:
As a non administrator (is that how it is called) I cannot assign labels. I think that why we label things using the bot in miq?

![kubeclient next tracker of non backward compatible issues issue 199 abonas kubeclient](https://cloud.githubusercontent.com/assets/3010449/24454358/3642f686-1494-11e7-95ce-f37c5d6c19c3.png)

abonas:
@moolitayer I believe that you're right. I can create a new label and assign to the relevant issues. How would you like this label to be called?

moolitayer:
api change ?

abonas:
> api change ?

But backwards incompatible changes can be also non related to api change, so perhaps something more generic would be better

moolitayer:
I sent PRs for the items mentioned here:
- https://github.com/abonas/kubeclient/pull/292
- https://github.com/abonas/kubeclient/pull/293
- https://github.com/abonas/kubeclient/pull/290

And some fixes:
- https://github.com/abonas/kubeclient/pull/288
- https://github.com/abonas/kubeclient/pull/291
- https://github.com/abonas/kubeclient/pull/294

I did not tackle #240 at this time. We need something to do for the next release ;)

cben:
Out of date. Most already done.  We've had a couple major releases since.
[...]

Issue #235, opened 2017-03-13
Automatically detect kind and apiVersion in Resources


danajp:
So I've found myself writing a lot of code like this lately:

```ruby
kube_url = 'https://k8s.example.com'
core_client = Kubeclient::Client.new(kube_url, 'v1', :ssl_options => some_ssl_options)
ext_client = Kubeclient::Client.new(kube_url, 'apis/extensions/v1beta1', :ssl_options => some_ssl_options)
batch_client = Kubeclient::Client.new(kube_url, 'apis/batch/v1', :ssl_options => some_ssl_options)
```

And then I have to decide which client instance to use depending on the kind of entity I'm working with. The `core` client for services, `ext` client for deployments, `batch` client for jobs, etc. It'd be nice if this worked like `kubectl apply -f some-entity.yml` where I could specify the `kind` and `apiVersion` fields in my `Kubeclient::Resource` object and have the client instance figure out which api endpoint to use:

```ruby
deployment = Kubeclient::Resource.new(
  :kind => 'Deployment',
  :apiVersion => 'extensions/v1beta1',
  # ...
)
service = Kubeclient::Resource.new(
  :kind => 'Service',
  :apiVersion => 'v1',
  # ...
)

client.create_entity(deployment)
client.create_entity(service)
```

I'm considering writing a wrapper to implement this, but I'd much rather modify kubeclient and contribute those changes upstream instead. Is this something the project would be interested in?

simon3z:
@danajp is this a duplicate of #208 ?

danajp:
Yep, my issue search-fu is weak. I'll close this. Thanks!

PR #292, opened 2018-01-23
Drop entity classes

### LINKED ISSUE #198: Only one resource class should exists: Kubeclient::Resource
Kubeclient::Resource should be returned for all entites

moolitayer:
Fix #198 

Drop entity classes (e.g Kubeclient::Service).
Using them required dynamically creating classes under Kubeclient namespace during discovery.
We have had several hairy bugs around this area, and the code does not have  real benefit for users.

Part of: #199


moolitayer:
@cben @grosser @kirs @jimmidyson Please take a look

moolitayer:
@jhernand please take a look

jhernand:
@moolitayer will this be backported to 2.x, or is it only for 3.x?

README.md
```diff
@@ -476,10 +476,21 @@ client.process_template template
...
 ## Upgrading
 
+#### past version 3.0
+
+Specific entity classes mentioned in (past version 1.2.0)[past_version_1.2.0] have been dropped.
+Return values and expected classes are always Kubeclient::Resource.
+Checking the type of a resource can be done using:
+```
+> pod.kind
+Pod
```
cben: I think this is a string — quotes missing?

test/test_guestbook_go.rb
```diff
@@ -113,7 +113,7 @@ def delete_replication_controllers(client, namespace, replication_controllers)
   private
 
   def construct_base_rc(namespace)
-    rc = Kubeclient::ReplicationController.new
+    rc = Kubeclient::Resource.new
```
cben: Shouldn't one set `rc.kind` now that it's not implied by class?
IIUC, it works because we use these objects via `create_replication_controller`.
It's not ideal end state (though maybe fine for now) if some Resource objects around will have `.kind` and some won't, and what works depends on how you use it...
cf #235, #241.

lib/kubeclient/common.rb
```diff
@@ -271,7 +252,7 @@ def watch_entities(resource_name, options = {})
     #
     #   Default response type will return a collection RecursiveOpenStruct
     #   (:ros) objects, unless `:as` is passed with `:raw`.
-    def get_entities(entity_type, klass, resource_name, options = {})
```
cben: Are `get_entities`, `get_entity` public APIs?  :+1: on this change, but if yes it should be documented.

test/test_guestbook_go.rb
moolitayer: It is correct though it did not change in this PR.

lib/kubeclient/common.rb
moolitayer: They are not, we agreed to add a comment separately in the README.md saying any usage 
not documented in the README.md is not supported.

PR #293, opened 2018-01-23
[WIP] Remove deprecated KubeException

moolitayer:
See discussion in https://github.com/abonas/kubeclient/pull/195

Part of: https://github.com/abonas/kubeclient/issues/199

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,6 +476,10 @@ client.process_template template
 
 ## Upgrading
 
+### past version 3.0
+
+Deprecated KubeException has been dropped, use `Kubeclient::HttpError` instead.
+
 #### past version 2.0
 
 Replace `KubeException` with `Kubeclient::HttpException`
--- a/lib/kubeclient/http_error.rb
+++ b/lib/kubeclient/http_error.rb
@@ -1,25 +1,20 @@
-# TODO: remove this on next major version bump
-# Deprected http exception
-class KubeException < StandardError
-  attr_reader :error_code, :message, :response
-
-  def initialize(error_code, message, response)
-    @error_code = error_code
-    @message = message
-    @response = response
-  end
-
-  def to_s
-    string = "HTTP status code #{@error_code}, #{@message}"
-    if @response.is_a?(RestClient::Response) && @response.request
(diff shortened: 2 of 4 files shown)


moolitayer:
@cben @grosser @kirs please review

moolitayer:
@jhernand please take a look

README.md
```diff
@@ -476,6 +476,10 @@ client.process_template template
 
 ## Upgrading
 
+### past version 3.0
+
+Deprecated KubeException has been dropped, use `Kubeclient::HttpException` instead.
+
```
jhernand: Should be `Kubeclient::HttpError`.

jhernand:
@pkliczewski, @masayag the KubeVirt provider uses the deprecated exception:

https://github.com/ManageIQ/manageiq-providers-kubevirt/blob/d46aca33eb6330b4e79a16100c6379e53a912ab5/app/models/manageiq/providers/kubevirt/infra_manager/refresh_worker/runner.rb#L192

Makes sure to change that to `Kubeclient::HttpError` before upgrading to the upcoming version 3.0 of `kubeclient`. You can change it now, if you want, as version 2.4 already supports it.

README.md
moolitayer: see https://github.com/abonas/kubeclient/pull/288

moolitayer:
@cben please review

moolitayer:
Discussed with @cben 
Planning to release maser, as we haven't yet released a version where this is deprecated we should wait with this one.

cben:
status: I've been postponing this as I see an increasing needs for errors where HttpError would make no sense, specifically around `Config`.  Plus if we'll get auth auto-renewal (#393), we might start seeing Config-related errors from `Client` methods (get_pods etc) that currently only raise HttpError.

All this leads to me to think kubeclient still needs a top class/module covering all its errors.
For now that's  `KubeException` but we might want to transition to say `Kubeclient::Error`.
As `Config` errors are currently various RuntimeError, KeyError etc, I want to figure out if it's reasonable to substitute a different class...
=> So, still stalling on this... :construction:

PR #294, opened 2018-01-23
Comment on internal interfaces

moolitayer:


abonas:
Why is there a need to place those comments? Wont it be hard to maintain them?

lib/kubeclient/http_error.rb
```diff
@@ -18,6 +19,7 @@ def to_s
   end
 end
 
+# Note: this IS NOT(!) a supported interface, and might change without a version bump
```
cben: why?  what are people supposed to do?  they need to know what exceptions to expect from us.

lib/kubeclient/watch_stream.rb
```diff
@@ -3,6 +3,7 @@
 module Kubeclient
   module Common
     # HTTP Stream used to watch changes on entities
+    # Note: this IS NOT(!) a supported interface, and might change without a version bump
```
cben: I think only `WatchStream.new` is internal here.  This *is* an object users get back from `.watch_pods` etc., and `.watch` and `.finish` are methods users interact with, both mentioned in README I think.

moolitayer:
`WatchStream.initialize` really seems like the only one and it seems users would have to be doing something out of the ordinary (and undocumented) to be using it.

I'd prefer to close this one

cben:
Thanks for the analysis.  Agreed.

PR #299, opened 2018-02-03
allow setting default :as for the whole client

grosser:
also passes all replies through format_response which will allow for easy addition of a new
:as type, I'm plaing to use `JSON.parse(value, symbolize_names: true)`


cben:
Didn't review in detail but looks pretty good!
Some ideas, not sure if helping or overkill:

Consider making WatchStream too ask the client to format notices, so overriding `format_response` would affect it too.  (May becomes better to split it into 3 methods rather than one returning 3 types depending on args.)

I've thought about letting :as param take an object, that's expected to have a method (or 3) implementing conversion of responses.
There would be a helper mapping :raw and :ros symbols into 2 such builtin objects; but any other object is possible.

grosser:
updated ... trying to keep this change as small as possible :D
looks good ?

cben:
LGTM :+1:  @enoodle can you review as well?

@grosser can you document this in README?

cben:
Merging; @grosser please either send followup to document this in README & CHANGELOG, or nag me to do it.

PR #305, opened 2018-02-16
document as being read from initializer

grosser:
@cben 

for https://github.com/abonas/kubeclient/pull/299


cben:
Should still include some explanation what it does (that you get a JSON string)

grosser:
better ?

cben:
:+1:

PR #306, opened 2018-02-16
support parsed json and parsed json with symbolized keys

grosser:
@cben

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,11 @@ pods = client.get_pods as: :raw
 node = client.get_node "127.0.0.1", as: :raw
 ```
 
+Other formats are:
+ - `:ros` (default) for `RecursiveOpenStruct`  
+ - `:parsed` for `JSON.parse` 
+ - `:parsed_symbolized` for `JSON.parse(..., symbolize_names: true)`
+
 #### Delete an entity (by name)
 
 For example: `delete_pod "pod name"` , `delete_replication_controller "rc name"`, `delete_node "node name"`, `delete_secret "secret name"`
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -427,22 +427,31 @@ module Kubeclient
     private
 
     def format_response(as, body, list_type = nil)
-      return body if as == :raw
+      case as
+      when :raw
+        body
+      when :parsed
+        JSON.parse(body)
+      when :parsed_symbolized
+        JSON.parse(body, symbolize_names: true)
+      when :ros
+        result = JSON.parse(body)
 
-      result = JSON.parse(body)
+        if list_type
+          resource_version =
+            result.fetch('resourceVersion') do
+              result.fetch('metadata', {}).fetch('resourceVersion', nil)
+            end
 
-      if list_type
-        resource_version =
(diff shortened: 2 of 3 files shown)


lib/kubeclient/common.rb
```diff
@@ -427,22 +427,35 @@ def api
...
-        # result['items'] might be nil due to https://github.com/kubernetes/kubernetes/issues/13096
-        collection = result['items'].to_a.map { |item| Kubeclient::Resource.new(item) }
-
-        Kubeclient::Common::EntityList.new(list_type, resource_version, collection)
+      case as
+      when :raw
+        body
+      when :parsed
+        result = JSON.parse(body)
+        result = result.fetch('items') if list_type
```
cben: I'm reluctant to drop information, especially collection resource_version.  "Use this format, unless you need that one info it doesn't support"...
Simplest alternative is just return top-level data, but I agree requiring user to access `[:items]` for common case is cumbersome (and inconsistent with :ros behavior).
How about returning an array that also answers .kind and .resource_version? One way to implement:
```
module EntityListMixin
  attr_accessor :kind, :resource_version
end  

result = JSON.parse(body)
if list_type
  result.extend(EntityListMixin)
  result.kind = list_type
  result.resource_version = resource_version
end
```

P.S. opened #307 to discuss some funkiness in how we set collection `.kind`.  That is orthogonal, but there is one aspect that might be relevant here:
- I'd love `get_foos` in the future to populate .kind on each entity in the array, not just on the collection.  The k8s API is IMHO unfriendly here, it only gives you `"kind": "NodeList"` and assumes you're a strongly-typed client that knows `NodeList.items` is an array of `Node`.

grosser: why would anyone need the kind/version of something they just queried ?

cben: kind is quite redundant but collection resource_version is useful to start a watch from.

Anyway, this can be added later.

cben:
LGTM.
I'd like more eyes on a new interface...  @jhernand @masayag @enoodle (anyone else welcome of course)

README.md
```diff
@@ -304,6 +304,8 @@ pods = client.get_pods as: :raw
 node = client.get_node "127.0.0.1", as: :raw
 ```
 
+Other formats are: `:parsed` for `JSON.parse` or `:parsed_symbolized` for `JSON.parse(..., symbolize_names: true)`
+
```
masayag: Could you include ```:ros``` also in the supported formats ?

Before this PR, any type other than ```:raw``` was considered as ```:ros```(default response format), or even if miss-typed (couldn't find any validation for the expected ```:as``` value). 

However, with this PR, it is mandatory now for clients to specify exactly ```:ros``` or not to pass a format in order to get response it in that format.

grosser: added ros

lib/kubeclient/common.rb
grosser: - `Continue` is never returned from `ros` either ?
 - `Kind` is always what the user requested anyway, so idk if there are a lot of usecases that need that

... changed it back since fetching `items` is easy and means we have less code/feature-requests ...

PR #308, opened 2018-02-28
Read and store namespace in Config object

hammady:
If a namespace exists in a `KUBECONFIG` file, it will be read and stored
in the `Kubeclient::Config::Context` object. This is what `kubectl` implements
when reading from config files and automatically applies on all
commands.

Example usage:

    puts config.context.namespace

Tests modified to add an assertion in `test_config.rb:check_context`. All pass.


hammady:
Welcome 😊

PR #313, opened 2018-03-23
Update recursive-open-struct dependency

pravi:
They have not introduced a breaking change in the past in minor releases https://github.com/aetherknight/recursive-open-struct/blob/master/CHANGELOG.md

Asking them to be sure https://github.com/aetherknight/recursive-open-struct/issues/57 

Debian already has 1.1.0 and it would be good to sync versions


pravi:
@cben how about now?

cben:
LGTM.  @moolitayer what do think?

PR #319, opened 2018-04-27
Add timestamps and since_time to get_pod_log

holyketzer:
Hi guys! We like your client and need this small feature )

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -360,10 +360,13 @@ module Kubeclient
       end
     end
 
-    def get_pod_log(pod_name, namespace, container: nil, previous: false)
+    def get_pod_log(pod_name, namespace,
+                    container: nil, previous: false, timestamps: false, since_time: nil)
       params = {}
       params[:previous] = true if previous
       params[:container] = container if container
+      params[:timestamps] = timestamps if timestamps
+      params[:sinceTime] = format_datetime(since_time) if since_time
 
       ns = build_namespace_prefix(namespace)
       handle_exception do
@@ -426,6 +429,18 @@ module Kubeclient
     end
 
     private
+
+    # Format ditetime according to RFC3339
+    def format_datetime(value)
+      case value
+      when DateTime, Time
+        value.strftime('%FT%T.%9N%:z')
+      when String
+        value
+      else
+        raise ArgumentError, "unsupported type '#{value.class}' of time value '#{value}'"
+      end
+    end
 
     def format_response(as, body, list_type = nil)
       case as
--- a/test/test_common.rb
+++ b/test/test_common.rb
@@ -2,6 +2,14 @@ require_relative 'test_helper'
 
 # Unit tests for the common module
 class CommonTest < MiniTest::Test
+  class ClientStub
+    include Kubeclient::ClientMixin
+  end
+
+  def client
+    @client ||= ClientStub.new
+  end
+
   def test_underscore_entity
     %w[
       Pod pod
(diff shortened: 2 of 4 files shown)


cben:
Nice!
For since_time, can you document the exact the allowed format(s)?
Or perhaps avoid the issue by having user supply `datetime` object and kubeclient format it.
Or support both string and datetime?

cben:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#-strong-proxy-operations-strong-

hmm, sinceTime is missing from the doc (although mentioned in other params)

holyketzer:
> hmm, sinceTime is missing from the doc (although mentioned in other params)

it presents in `kubectl` help and this branch works with Kubernetes v1.9. 

```
kubectl logs -h
Print the logs for a container in a pod or specified resource. If the pod has only one container, the container name is
optional.

Aliases:
logs, log

Examples:
  # Return snapshot logs from pod nginx with only one container
  kubectl logs nginx

  # Return snapshot logs for the pods defined by label app=nginx
  kubectl logs -lapp=nginx

  # Return snapshot of previous terminated ruby container logs from pod web-1
  kubectl logs -p -c ruby web-1

  # Begin streaming the logs of the ruby container in pod web-1
  kubectl logs -f -c ruby web-1

  # Display only the most recent 20 lines of output in pod nginx
  kubectl logs --tail=20 nginx

  # Show all logs from pod nginx written in the last hour
  kubectl logs --since=1h nginx

  # Return snapshot logs from first container of a job named hello
  kubectl logs job/hello

  # Return snapshot logs from container nginx-1 of a deployment named nginx
  kubectl logs deployment/nginx -c nginx-1

Options:
  -c, --container='': Print the logs of this container
  -f, --follow=false: Specify if the logs should be streamed.
      --include-extended-apis=true: If true, include definitions of new APIs via calls to the API server. [default true]
      --interactive=false: If true, prompt the user for input when required.
      --limit-bytes=0: Maximum bytes of logs to return. Defaults to no limit.
      --pod-running-timeout=20s: The length of time (like 5s, 2m, or 3h, higher than zero) to wait until at least one
pod is running
  -p, --previous=false: If true, print the logs for the previous instance of the container in a pod if it exists.
  -l, --selector='': Selector (label query) to filter on.
      --since=0s: Only return logs newer than a relative duration like 5s, 2m, or 3h. Defaults to all logs. Only one of
since-time / since may be used.
      --since-time='': Only return logs after a specific date (RFC3339). Defaults to all logs. Only one of since-time /
since may be used.
      --tail=-1: Lines of recent log file to display. Defaults to -1 with no selector, showing all log lines otherwise
10, if a selector is provided.
      --timestamps=false: Include timestamps on each line in the log output

Usage:
  kubectl logs [-f] [-p] (POD | TYPE/NAME) [-c CONTAINER] [options]
```

holyketzer:
> Nice!
> For since_time, can you document the exact the allowed format(s)?
> Or perhaps avoid the issue by having user supply datetime object and kubeclient format it.
> Or support both string and datetime?

Ok, I supported both and documented it.

lib/kubeclient/common.rb
```diff
@@ -427,6 +430,18 @@ def api
 
     private
 
+    # Format ditetime according to RFC3339
+    def format_datetime(value)
+      case value
+      when DateTime, Time
+        value.rfc3339(9)
```
cben: are you sure this works with Time?
```
[2] pry(main)> Time.now.rfc3339(9)
NoMethodError: undefined method `rfc3339' for 2018-04-30 21:03:57 +0300:Time
[3] pry(main)> DateTime.now.rfc3339(9)
=> "2018-04-30T21:04:20.133891062+03:00"
```

holyketzer: Hmm, I checked it in Rails console

```
2.4.3 :116 >   Time.now.rfc3339(9)
 => "2018-04-30T21:28:46.315155000+03:00"
```

but... yeah, in pure Ruby it doesn't work

holyketzer: Done

test/test_common.rb
```diff
@@ -29,4 +37,16 @@ def test_underscore_entity
...
     end
   end
+
+  def test_format_datetime_with_string
+    value = '2018-04-27T18:30:17.480321984Z'
+    formatted = client.send(:format_datetime, value)
+    assert_equal(formatted, value)
+  end
+
+  def test_format_datetime_with_time
```
cben: last thing: can you also test with a Time instance

holyketzer: Done

Issue #318, opened 2018-04-26
::Kubeclient::Client::handle_uri assumes k8s api mounted at / for paths longer than 2 components


dirkd:
Hi,

when trying to use kubeclient (v3.0.0 via Gitlab's Kubernetes integration) with a Rancher k8s cluster (both in v1.6 and v2) there's a problem regarding kubeclients assumptions (rancher/rancher#13058).

The following code assumes that as soon as there's more than two path components the second one (third entry in components after splitting) has to be the k8s api_group the user wants set:

https://github.com/abonas/kubeclient/blob/e37a8029275a6032a6f9b641b3800601ff1a2a0b/lib/kubeclient/common.rb#L149-L150

This works as long as the k8s apiserver is mounted at the `/`-route on the host providing the apiserver. As Rancher provides a common endpoint for multi-cluster environments it mounts the k8s apiserver at a deeper point of the path (to differentiate the environment) which breaks kubeclients assumptions on URL structure.

cben:
Great find!
I think this is strongly related to  #284 and we could to solve both at once.
Currently user gives kubeclient 2 parts e.g.: `'https://host:port/apis/batch', 'v1'` (to take one API group as a running example)
Unfortunately that's neither a convenient way for user to split it (this is what #284 argues) — **nor for kubeclient**!

Looking at the code, what kubeclient needs in addition to the combined path `https://host:port/apis/batch/v1` is knowing the `batch/v1` part because it also has (had?) to be sent in `apiVersion` due to https://github.com/kubernetes/kubernetes/issues/6439.
What you see here is the silliness of kubeclient trying to extract the `batch` part from url, to recombine with `/v1`! https://github.com/abonas/kubeclient/blob/e37a8029275a6032a6f9b641b3800601ff1a2a0b/lib/kubeclient/common.rb#L324

## Proposal

```
Kubeclient::Client.new('http://localhost:8080', path: 'apis', api_version: 'batch/v1')
```

- Naming is hard, improvements welcome.

- I do NOT want to overload existing positional args with new meaning, would require too much guessing.  Named hash arguments are good way to distinguish new interface from old.

- Why explicit param for `apis/` portion?  It usually could be guessed — `api/v1`, `apis/...` for everything else (though at some point `api/v2` will come along?).  But we already know one other case: Openshift extensions are under `oapi/v1`!  So let's keep it explicit.

[...]

Issue #284, opened 2018-01-10
dealing with api/v1 vs apis/extensions/v1beta1 is annoying since they are split between url and version


grosser:
url should be everything without the /api ... and version should be either `api/v1` or maybe just `v1 / extensions/v1beta1` and then smartly detect if it should be `api` or `apis`

cben:
Right!  Here's our similar code, also spliting on `/`:
https://github.com/ManageIQ/manageiq-providers-openshift/blob/93009fcb167/app/models/manageiq/providers/openshift/container_manager_mixin.rb#L40-L52

1. Any idea if this can be backward compatible?
2. Does this interact with #208?  I this would be a first step to make multi-version use by multiple Client instances a tiny bit less annoying, and doesn't affect/depend on ideas for multi-version with one instance...

grosser:
I'd love to have multi-version in a single client too, so you can pass the
version as an option, but that can be step 2 ...
to be backwards compatible it could just blow up if the url has a path or
store the url path as `version_base` and then not override it when present

cben:
For anyone tackling this: the tests added in #457 are a good starting point.
We should still support everything these tests cover, for compatibility.

However, if it makes things easier, can have new keyword params for new behavior, supporting only one scheme.
Something like
```
Kubeclient::Client.new(
  # Exactly what you get from `config.context(...).api_endpoint`.
  # No /api, /apis, /oapi allowed.
  endpoint: "http://example.com/k8s/clusters/c-somerancherID",  

  # Exactly the `apiVersion` in the JSON representation for resources you want to access.
  # (One should NOT pass here `config.context.api_version` which is version of kubeconfg file, 
  #  see commit c593c9eef58e56f00c396f4715437eed673ca98e).
  # Would `group:` can be a clearer name then `api_version:`?
  api_version: "autoscaling/v1",

  ssl_options: config.context.ssl_options,
  auth_options: config.context.auth_options
)
```

PR #58, opened 2015-04-01
Added automatic inclusion of object kind and api version for create requ...

stephenjlevy:
Creating entities was failing because the spec requires the entity kind and apiversion.  Added code to automatically include these fields.

PR #333, opened 2018-05-31
[WIP] [EXPERIMENT] Don't send kind, apiVersion in create POST

cben:
Can we revert #58?

Seems https://github.com/kubernetes/kubernetes/issues/6439 fixed - from which kubernetes version?
https://github.com/kubernetes/kubernetes/pull/10200 landed in v1.1.0.

Works for me on OpenShift 3.10-ish:

```
[8] pry(main)> kclient.create_service(Kubeclient::Resource.new(metadata: {namespace: 'default', name: 'foo'}, spec: {ports: [{port: 80}]}))
RestClient.post "https://internal-api.bpaskinc.origin-gce.dev.openshift.com:8443/api/v1/namespaces/default/services", "{\"metadata\":{\"namespace\":\"default\",\"name\":\"foo\"},\"spec\":{\"ports\":[{\"port\":80}]}}", "Accept"=>"*/*", "Accept-Encoding"=>"gzip, deflate", "Content-Length"=>"80", "Content-Type"=>"application/json", "User-Agent"=>"rest-client/2.0.2 (linux-gnu x86_64) ruby/2.4.1p111"
=> #<Kubeclient::Resource kind="Service", apiVersion="v1", metadata={:name=>"foo", :namespace=>"default", :selfLink=>"/api/v1/namespaces/default/services/foo", :uid=>"b2b41793-64e1-11e8-9540-42010a8e0002", :resourceVersion=>"45722", :creationTimestamp=>"2018-05-31T14:48:36Z"}, spec={:ports=>[{:protocol=>"TCP", :port=>80, :targetPort=>80}], :clusterIP=>"172.30.186.136", :type=>"ClusterIP", :sessionAffinity=>"None"}, status={:loadBalancer=>{}}>
```
(note POST body didn't include kind nor apiVersion)

Motivation: if we don't need to know apiVersion, this may (?) open up #284, #318, and other issues to simpler interface.
- I'm not sure yet if we should commit to "we don't need to know apiVersion"
- OTOH, I suspect we could infer apiVersion from discovery anyway?

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -306,17 +306,11 @@ module Kubeclient
     def create_entity(entity_type, resource_name, entity_config)
       # Duplicate the entity_config to a hash so that when we assign
       # kind and apiVersion, this does not mutate original entity_config obj.
+      # TODO: skip?
       hash = entity_config.to_hash
 
       ns_prefix = build_namespace_prefix(hash[:metadata][:namespace])
 
-      # TODO: temporary solution to add "kind" and apiVersion to request
-      # until this issue is solved
-      # https://github.com/GoogleCloudPlatform/kubernetes/issues/6439
-      # TODO: #2 solution for
-      # https://github.com/kubernetes/kubernetes/issues/8115
-      hash[:kind] = (entity_type.eql?('Endpoint') ? 'Endpoints' : entity_type)
-      hash[:apiVersion] = @api_group + @api_version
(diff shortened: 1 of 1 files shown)


chasebolt:
i feel like k8s 1.1.0 is old enough that removing `apiVersion` and `kind` is sufficient. bump the gem version as a breaking change. can always re-eval inferring `apiVersion` from a discovery method later if needed.

PR #340, opened 2018-06-26
remove broken badge

grosser:
@cben 

this looks bad :D 
screen shot 2018-06-26 at 4 01 39 pm

... ideally get rid of code-climate too and replace with https://github.com/grosser/single_cov :D


cben:
Thanks!  Looks good (literally :-) to me.

Issue #322, opened 2018-05-10
proxy_url generating invalid URLs for k8s 1.10


lambertjosh:
It looks like kubeclient proxy_url may be generating an incorrect or out of date proxy URL format. 

Right now for services, it outputs as [documented in the README](https://github.com/abonas/kubeclient#get-a-proxy-url) this format: `/api/v1/proxy/namespaces/ns/services/srvname:srvportname`

However the [Kubernetes documentation](https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#manually-constructing-apiserver-proxy-urls) specifically states the following format: `/api/v1/namespaces/ns/services/srvname:srvportname/proxy`

We've found that the current string format, works well in GKE and other places. However recently on minikube, it has started to fail. The officially documented format however still works.

Perhaps the gem should switch over to generating the officially documented version?

lambertjosh:
I was peeking at the code, and it seems like not only does services support the new scheme it may no longer support the old scheme? 

https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/common.rb#L402

lambertjosh:
Looks like it was removed in 1.10 here: https://github.com/kubernetes/kubernetes/pull/59884

PR #321, opened 2018-05-05
update http gem to 3.0+

pravi:
Only breaking change is drop of support for ruby 2.0 and 2.1 which we don't care. Fixes #320 
twitter gem already moved to http 3.0 and this will help in debian packaging. We prefer to maintain only a single version of http.

PR #344, opened 2018-07-23
Changelog & bump for 4.0.0 release

cben:
Includes #323 fixing #322 (finally! sorry :flushed:) and #321.
Bumping major because we're knowingly breaking a k8s version that previously was believed to work (even if so old it's hard to test).

@lambertjosh @grosser @pravi please review.


cben:
Released https://rubygems.org/gems/kubeclient/versions/4.0.0

PR #355, opened 2018-10-08
make api_service method work

eatwithforks:
@cben  @moolitayer

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -240,8 +240,12 @@ module Kubeclient
       end
     end
 
+    # Inspired by https://stackoverflow.com/questions/1509915/converting-camel-case-to-underscore-case-in-ruby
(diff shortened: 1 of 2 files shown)


grosser:
I'd say either major or we make both available and call it a patch.
Is there a planned major bump coming up that we can piggy back on?

cben:
I'm working on making both old and new work...

cben:
Rebased, still blocked on compatibility, working on it.
I'm going to first rewrite `parse_definition` and add more tests in separate PR and then return to this...

cben:
reminder to self: there will never be a perfect time but fixing this sooner is better.
I should merge this and bump major version sometime soon...

cben:
Oh, wow.  Sorry for holding this up so long.
I'm planning a major version bump soon, see #435

Issue #283, opened 2018-01-10
Support API chunking


cben:
k8s recently finally got response chunking (kinda pagination, of a consistent snapshot at single resourceVersion!):
https://github.com/kubernetes/community/pull/896 => https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/api-chunking.md
https://github.com/kubernetes/features/issues/365
https://kubernetes.io/docs/reference/using-api/api-concepts/

This has similar benefits to streaming parsing (#254) for peak RAM on client, but also reduces peak RAM on server, and avoids some annoyances like client timing out while server builds response.
Would be great in at some point in kubeclient.
Do we still also want streaming parsing?  Don't have an opinion yet.

First I want to figure out if same block-based API as in #254 is also good for chunking?  Or do we need something lower-level that gives user more control?

cben:
Seems this is supported starting with kubernetes 1.8 API.

cben:
2 open questions:

1. How should whole-collection metadata, notably resource_version, be returned?
    #254 in current form first yields each item, at the end returns a hash with `resourceVersion` & `kind` keys.
    For streaming parse this was motivated by not assuming k8s always puts `items` key last in the JSON.
    For chunking, we do have metadata in each chunk, and parse whole chunk at a time, so could also expose it earlier.

2. what to do with 410 Gone mid-iteration?

    > Like a watch operation, a continue token will expire after a short amount of time (by default 5 minutes) and return a 410 Gone if more results cannot be returned. In this case, the client will need to start from the beginning or omit the limit parameter.
    >  -- [api-concepts](https://kubernetes.io/docs/reference/using-api/api-concepts/)

    How should this scenario be handled by kubeclient?  Raise exception?  Does this preclude access to whole-collection metadata?

Perhaps we need a 2-step interface like `watch`:
```
iter = client.iter_pods
puts "can access #{iter.resourceVersion} early?"
begin
  iter.each { |pod| ... }
  puts "can access #{iter.resourceVersion} at the end"
rescue
  puts "can still access #{iter.resourceVersion} if lost it mid-way!"
end
```
robust but awkward :-(
would prefer simpler 1-step `client.get_pods { ... }` if possible.

agrare:
@cben I took a crack at this over in https://github.com/abonas/kubeclient/pull/356
[...]

PR #356, opened 2018-10-17
Add support for API paging with limit and continue

agrare:
Implement API chunking support as specified in
https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks

Add limit and continue as parameters to get_entities and add continue as an optional attribute (default to nil) to EntityList.

Sample usage:
```
continue_token = nil
begin
  options = {:limit => 100}
  options.merge!(:continue => continue_token) unless continue_token.nil?

  entities = connection.get_pods(options)
  break if entities.nil?

  continue_token = entities.continue

  # do something with the chunk of entities returned
end until entities.continue.nil?
```

Fixes: https://github.com/abonas/kubeclient/issues/283


agrare:
cc @cben @Ladas

abonas:
> Add limit and continue as parameters to get_entities and get_entities

I guess you meant get and watch? :)

agrare:
> I guess you meant get and watch? :)

Yes thank you!

cben:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#watch-list-65 :

`continue` — .... This field is not supported if watch is true. ...
`limit` — .... This field is not supported if watch is true. ...

so not relevant for watch.

agrare:
> Watch? What does chunking mean in context of watch? Also, I don't see changes related to watch.

Ah okay they shared the same arguments as get [ref](https://github.com/abonas/kubeclient/blob/061a734c901a1f203c3d8a38bbace01784fc9f00/lib/kubeclient/common.rb#L45) so I thought that was always true.  I'll split those up then so limit/continue aren't possible args for watch

agrare:
@cben fixed

agrare:
@cben as for the higher-level vs lower-level interface, I actually like being able to process the chunks rather than just getting an iterator because I want to save the smaller chunks as well.

I can definitely see the benefit of a "nicer" iterator though where the caller doesn't have to worry about it, but this is closer to how the API is implemented.

agrare:
@cben I pushed a version which always uses paging and handles the continue internally, either yielding the batches or returning the full list if no block is given.  This makes it easier on the callers but I haven't found a nice way of cleanly handling the format_response since it parses the body and I had to parse the JSON to get the continue out of the payload.

lib/kubeclient/entity_list.rb
```diff
@@ -5,12 +5,17 @@ module Common
...
-      def initialize(kind, resource_version, list)
+      def initialize(kind, resource_version, list, continue = nil)
         @kind = kind
         # rubocop:disable Style/VariableName
         @resourceVersion = resource_version
+        @continue = continue
         super(list)
       end
+
+      def last?
```
cben: Nice!  I think this is worth keeping if we return to low-level version.

agrare: Yeah absolutely, this made the calling code a lot cleaner

agrare:
> I now lean towards the low-level version, which does the job, is 100% opt-in, and has no open questions.

Yeah I agree, it is simple and lets us get this functionality in without changing the default behavior.
I'll work on the docs/changelog/tests now, thanks @cben !

agrare:
> docs (with usage example; mention risk of 410 Gone; mention servers may ignore it). changelog if possible.

Added, please check that they meet with what you expected.  I added an Unreleased section to the changelog similar to how we handle per-sprint manageiq changelogs, when you cut another release you can just replace that with the release number.

> some tests? I'd include both responses from server supporting limit/continue, and a server that ignores limit.

I added a simple test where two pages are returned.  I am going to add a test for 410 Gone condition as well.  I'll need to try to find a pre-1.9 cluster to test a server that doesn't support chunking though, any hints?

README.md
```diff
@@ -345,6 +345,29 @@ You can specify multiple labels (that option will return entities which have bot
...
+Get all entities of a specific type in chunks:<br>
+
+```ruby
+continue = nil
+options = {
+  limit: 1_000
+}
+
+begin
+  options[:continue] = continue unless continue.nil?
```
cben: I propose documenting that user may always pass `continue: nil` with same result as omitting it.
Then user could do:
```
continue = nil
begin
  entities = client.get_pods(limit: 1_000, continue: continue)

  continue = entities.continue
end until entities.last?
```

P.S. googling this [`begin ... end until ...` construct](https://rosettacode.org/wiki/Loops/Do-while#Ruby), turns out Matz [regrets it and considers removing it](http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-core/6745).  So let's write this as `loop` with `break if entities.last?` inside.

README.md
```diff
@@ -345,6 +345,29 @@ You can specify multiple labels (that option will return entities which have bot
 pods = client.get_pods(label_selector: 'name=redis-master,app=redis')
 ```
 
+Get all entities of a specific type in chunks:<br>
```
cben: nit: the `<br>` is unnecessary, especially as it's followed by empty line.

README.md
agrare: :+1: it looks like kubernetes does handle being passed a nil continue option on the first get so yeah this is much cleaner

agrare: Done

README.md
agrare: Done

agrare:
@cben can you take another look?  I updated the readme and added tests for 410 gone

README.md
cben: kubernetes doesn't see the nil, it's dropped by
```
SEARCH_ARGUMENTS.each { |k, v| params[k] = options[v] if options[v] }
```

cben:
LGTM :100:

agrare:
Thanks so much @cben!

Issue #360, opened 2018-11-05
custom resources & istio


f4tq:
Hi,
First off, many thanks for kubeclient.  It works very well.

I've been able to - mostly - use kubeclient to discover and instantiate all istio objects via kubeclient but there are a couple exceptions.  

The [istio 1.0.3 custom resource definitions ](https://github.com/istio/istio/blob/1.0.3/install/kubernetes/helm/istio/templates/crds.yaml) in question raise an exception due to the kind for a bunch of their definitions not having a capital letter in `kind` -  rbac.istio.io and configs.istio.io.
I'm able fix it like this by just using kind in [lib/kubeclient/common.rb](https://github.com/f4tq/kubeclient/blob/master/lib/kubeclient/common.rb#L141):
```
-      prefix = kind[0..kind.rindex(/[A-Z]/)] # NetworkP
+      prefix = kind =~ /[A-Z]/ ? kind[0..kind.rindex(/[A-Z]/)] : kind # NetworkP
```
I'll submit a PR for this if you'd like but given that some of the generated functions might not follow your pattern, I thought I'd ask first.  I also see that you have a istio fork so you may have seen this already.  Please advise.

Here is how I'm instantiating the client to the using the api group/version form:
```ruby
  istioRbacClient = Kubeclient::Client.new(
    "#{context.api_endpoint}/apis/rbac.istio.io",
    "v1alpha1" ,
    ssl_options: context.ssl_options,
    auth_options: context.auth_options
  )
```
AND
```ruby 
  istioConfigClient = Kubeclient::Client.new(
    "#{context.api_endpoint}/apis/config.istio.io",
    "v1alpha2" ,
    ssl_options: context.ssl_options,
    auth_options: context.auth_options
  ) 
```

The exception before the change above looks like : 
```
Resource: https://XXXXX.hcp.eastus2.azmk8s.io/apis/config.istio.io/v1alpha2
parse_definition kind noop name: noops
Couldn't read /output/packages/kubeconfig.json ["/var/lib/gems/2.3.0/gems/kubeclient-4.0.0/lib/kubeclient/common.rb:136:in `parse_definition'", "/var/lib/gems/2.3.0/gems/kubeclient-4.0.0/lib/kubeclient/common.rb:481:in `block in load_entities'", "/var/lib/gems/2.3.0/gems/kubeclient-4.0.0/lib/kubeclient/
[...]

PR #361, opened 2018-11-07
Tolerate lowercase kubenetes `kind`s

f4tq:
- Istio stresses the type system by using unexpected kind definitions as explained in [issue 360](https://github.com/abonas/kubeclient/issues/360).

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -138,7 +138,7 @@ module Kubeclient
       # See: https://github.com/kubernetes/kubernetes/issues/8115
       kind = kind[0..-2] if %w[Endpoints SecurityContextConstraints].include?(kind)
 
-      prefix = kind[0..kind.rindex(/[A-Z]/)] # NetworkP
+      prefix = kind =~ /[A-Z]/ ? kind[0..kind.rindex(/[A-Z]/)] : kind # NetworkP
       m = name.match(/^#{prefix.downcase}(.*)$/)
       m && OpenStruct.new(
         entity_type:   kind, # ComponentStatus
--- a/test/test_common.rb
+++ b/test/test_common.rb
@@ -33,6 +33,8 @@ class CommonTest < MiniTest::Test
       BuildConfig build_config
       Image image
       ImageStream image_stream
+      dogstatsd dogstatsd
+      HTTPAPISpecBinding httpapispec_binding
     ].each_slice(2) do |singular, plural|
       assert_equal(Kubeclient::ClientMixin.underscore_entity(singular), plural)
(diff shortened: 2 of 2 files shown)


f4tq:
The tests that are failing also fail on a clean master. 

Looks like it has to do with certs.

lib/kubeclient/common.rb
```diff
@@ -138,7 +138,7 @@ def self.parse_definition(kind, name)
       # See: https://github.com/kubernetes/kubernetes/issues/8115
       kind = kind[0..-2] if %w[Endpoints SecurityContextConstraints].include?(kind)
 
-      prefix = kind[0..kind.rindex(/[A-Z]/)] # NetworkP
+      prefix = kind =~ /[A-Z]/ ? kind[0..kind.rindex(/[A-Z]/)] : kind # NetworkP
```
cben: This ternary assumes we'll see either `kind: CamelCase` or `kind: lowercase`, right?
What if we get `kind: "lowerCamelCase"`?
Also, even if using a lowercase kind as-is works, it feels pretty accidental given how we look for capital letters in `prefix`...

The purpose of this weird `prefix` dance is automatically adapting to the plural ending, which is sometimes `s` and sometimes `es` (and sometimes nothing e.g. Endpoints=endpoints but we special-case those above).

Except it doesn't extract just the `s` / `es` — `m[1]` ends up with the whole lowercase end.
To take a fuller example:
```
kind = "OAuthAccessToken"
name = "oauthaccesstokens"
prefix = kind[0..kind.rindex(/[A-Z]/)]      #  "OAuthAccessT"
m = name.match(/^#{prefix.downcase}(.*)$/)  # #<MatchData "oauthaccesstokens" 1:"okens">
# Using new underscore algorithm from #355
underscore_entity(kind)                  # "o_auth_access_token"
underscore_entity(prefix)                # "o_auth_access_t"
underscore_entity(prefix) + m[1]         # "o_auth_access_tokens"
```

Can we rewrite the whole thing to work in a clearer way?

cc @eatwithforks

f4tq: All the ternary does is check if there are any upper case letters.  If there are, do the original action.  If there aren't, then take `kind` as is.  

This way you don't get a negative index which was causing an uncaught exception (and crash).

There are 2 tests. This fix allows any kind, irregardless of casing, to make it through.

cben: Fair enough.  This is a strict improvement.  I'll merge this and see if I can rewrite it clearer separately...

cben:
close-cycting to re-run tests after #364 test fix.

Issue #362, opened 2018-11-08
support kube_config exec


cjhodges77:
Is there a plan to support the kube_config exec function? as in:
```console
users:
- name: user-name
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1alpha1
      command: heptio-authenticator-aws
      args:
```

motymichaely:
@cjhodges77 I needed this as well so I just implemented it ;)

cben:
Closed by #363 (thanks Moty again).

PR #363, opened 2018-11-08
Adds support for exec credential plugin

motymichaely:
Hey!

This PR offers a partial implementation of the Kube Config ExecCredential plugin (#362).

This was Inspired by: https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/exec/exec.go

Things to consider: 
1. Add support for Client Certificate and Key data fields.
1. Credentials refresh
1. Caching (with token expiration support)


motymichaely:
@cben I believe we should add a warning in README.  According to [this client-go issue comment](https://github.com/kubernetes/client-go/issues/388#issuecomment-377417057), exec credential will eventually replace any Azure, GCP and alike credential providers (which makes sense).

I am currently trying to figure out a way to implement token refreshment mechanism. There will be cases in which the client is initialized with a token that expires after an hour and a refreshing mechanism should come into play before the next API call is made.  BTW - I think this is also true for GoogleApplicationDefaultCredentials token. 

Any idea?

rhodrid:
I need this for EKS on AWS which uses the `aws-iam-authenticator` helper. For now I've forked but would like to see this in master.

cben:
Sorry for silence, I'm trying to land some discovery related stuff before fully context-switching to this...

>  exec credential will eventually replace any Azure, GCP and alike credential providers (which makes sense).

> I need this for EKS on AWS which uses the aws-iam-authenticator helper

Thanks for the context.  Yeah, this is important.
I'd still like to see an opt-out — IMHO a library should not offer dangerous functionality (especially "run arbitrary processes") without _also_ offering a safe "just parse" functionality.
Let's see.  As currently written `config.context(name)` will automatically do everything, including running `exec` commands.  With the passive-sounding `.context` name, I find that a bit ... surprising :-(
Perhaps opt-in like `.context(name, obtain_auth: true)` might be better from purely "does what it says" perspective?

Now, about refreshment.That's a bigger effort, may be better to land this first.
It opens some questions about object lifetime.  So far `Client` is effectively immutable.  Do you expect a single `Client` to

cben:
(sent too soon, edited now ^^)

lib/kubeclient/exec_credentials.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        # Verify apiVersion?
+        api_version = opts['apiVersion']
+        if api_version && api_version != creds['apiVersion']
+          raise "exec plugin is configured to use API version #{api_version}, " \
+            "plugin returned version #{creds['apiVersion']}"
+        end
+
+        raise 'exec plugin didn\'t return a status field' if creds['status'].nil?
+        raise 'exec plugin didn\'t return a token' if creds['status']['token'].nil?
```
cben: These string exceptions should become some exception class.
`KubeException` is deprecated because it's in global namespace, we've documented "The gem raises Kubeclient::HttpError or subclasses now."  But `Kubeclient::HttpError` is inappropriate here.
This is not very important, until we actually try to remove `KubeException`.

What's more interesting is understand when this can be raised.  If we'll support auto-renewal, then ANY request such as `get_pods` would potentially also raise these!
What exceptions does `Config` raise presently?  I see `KeyError`, and one string 'Unknown kubeconfig version'.
You're adding some `ArgumentError` and strings.

Perhaps add `Kubeclient::ConfigError` for all config problems?

BTW, if we add opt-out, need to choose error for refusing to exec commands too.

motymichaely: @cben I agree, `KubeClient::ConfigError` can be a good candidate for configuration errors.

motymichaely:
@cben For refreshing, I was thinking about the following:

1. Having bearer_token accept callable method which will return a valid token. We can then change [this line](https://github.com/motymichaely/kubeclient/blob/feature/exec-credential/lib/kubeclient/config.rb#L123) to a block, that will be called [here](https://github.com/motymichaely/kubeclient/blob/feature/exec-credential/lib/kubeclient/common.rb#L488):

```ruby
def bearer_token(bearer_token)
  token = bearer_token.respond_to?(:call) ? bearer_token.call : bearer_token
  @headers ||= {}
  @headers[:Authorization] = "Bearer #{token}"
end
```
2.  Then, adding a retry mechanism to [handle_exception](https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/common.rb#L115) when a 403 exception occurs (?). The retry mechanism can then refresh the token by calling `bearer_token(@auth_options[:bearer_token])` again and retry the call:

```ruby
def handle_exception
  retry_count = 0
  begin
    yield
  rescue RestClient::Exception => e
    if e.http_code == 403 
      retry_count += 1
      retry if retry_count < (@auth_options[:max_retries] || 1)
    end
    json_error_msg = begin
      JSON.parse(e.response || '') || {}
    rescue JSON::ParserError
      {}
    end
    error_klass = e.http_code == 404 ? ResourceNotFoundError : HttpError
    raise error_klass.new(e.http_code, err_message, e.response)
  end
end
```

Thoughts?

cben:
#### Do we need retry, or do we have sufficient info to renew before it expires?

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#input-and-output-formats

> - If an expiry is included, the bearer token and TLS credentials are cached until the expiry time is reached, or if the server responds with a 401 HTTP status code, or when the process exits.
> - If an expiry is omitted, the bearer token and TLS credentials are cached until the server responds with a 401 HTTP status code or until the process exits.

OK, we'll need ability to retry.

#### Is it always bearer tokens?
No, executed plugin might choose to return `clientCertificateData` & `clientCertificateData`.  We don't know which it will be until we run it.
=> Can't replace `bearer_token` with a block.  It's not even enough to allow whole `auth_config` to be an active object, because client certs come via `ssl_options`?
=> The PR as currently written assumes it's always `ExecCredentials.token`.  (I'm fine with merging this part first if you want.)

#### Does caching belong in `Client` or elsewhere?
It should be possible to use one config shared by many `Client` objects (for different API groups) and obtain / renew the credentials only once.
So let's say we have a new auth object.
I feel `Client` should ask the auth object for connection options *every time* (in case it expired or was renewed via another Client).
`handle_exception` will still try to handle 401 but it'd do something like `@auth_config.renew if @auth_config.respond_to?(:renew)`.
TODO: consider thread safety (requests from different threads hitting expiration or 401 simultaneously).

### Are we OK with plugins blocking for TTY input?
> When run from an interactive session, stdin is exposed directly to the plugin. Plugins should use a TTY check to determine if it’s appropriate to prompt a user interactively.

I guess that's OK.

cben:
Did I mention that it's simpler to first land this without renewal? :wink:

f4tq:
Hi,

I just pulled this PR and rebased up and it works great on all my istio 1.0.4  eks (k8 10) & aks (k8 11.0.4)  clusters.

Without this PR, kubeclient doesn't work with eks - period.

Kindly, take this PR and perfect it later.  

👍

lib/kubeclient/exec_credentials.rb
```diff
@@ -0,0 +1,60 @@
...
+  # Inspired by https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/exec/exec.go
+  class ExecCredentials
+    class << self
+      def token(opts)
+        require 'open3'
+        require 'json'
+
+        raise ArgumentError, 'exec options are required' if opts.nil?
+
+        cmd = opts['command']
```
cben: note for future (not blocker): if `command` is relative path, the Go implemetation resolves it relative to the kubeconfig file
https://github.com/kubernetes/kubernetes/pull/59495/files/6463e9efd9ba552e60d2555a3e6526ef90196473#diff-4c107b9e9f7f10a98e5c52f66b952e01R562

cben:
Sorry, should have merged before.  Though I wanted to make some changes — at least document — before releasing.
I've discovered we already have a documented way to opt-out from any external-file lookups, albeit the opt-out didn't work :-/ — fixed in #372.
So I felt I must extend it to external-file execution as well.
=> all done; merging this and opening followup PR.

I'm not converting string exceptions to class(es).  The `Config` code already raises a lot of non-KubeException/HttpError errors (tweaked README to not sound like it

cben:
If you want to see this released, it'd help if someone can review my followup #375.

(at the moment I'm focused on fixing an already released regression (#376) but hopefully finishing that soon)

PR #366, opened 2018-11-22
Simpler derivation of singular & plural method names

### LINKED ISSUE #307: .get_endpoints.kind == "Endpoint" vs .get_endpoint(...).kind == "Endpoints"
When getting a single entity, we return .kind as-is:
```
pry(main)> kclient.get_endpoint('docker-registry', 'default', as: :raw)[0..39]
=> "{\"kind\":\"Endpoints\",\"apiVersion\":\"v1\",\"m"
pry(main)> kclient.get_endpoint('docker-registry', 'default').kind
=> "Endpoints"
pry(main)> kclient.get_security_context_constraint('nonroot').kind
=> "SecurityContextConstraints"
```

For collections, k8s does not return individual .kind in each item, only a global list kind.
We do some magic to convert that to a singular .kind, without the `List`.
Somewhat confusingly, we do this on the EntityList pseudo-array, no .kind on individual item:
```
pry(main)> kclient.get_endpoints(as: :raw)[0..39]
=> "{\"kind\":\"EndpointsList\",\"apiVersion\":\"v1"
pry(main)> kclient.get_endpoints.kind
=> "Endpoint"
pry(main)> kclient.get_endpoints.first.kind
=> nil
pry(main)> kclient.get_security_context_constraints.kind
=> "SecurityContextConstraint"
```

The **immediate bug** here is "Endpoints" vs "Endpoint" and  "SecurityContextConstraints" vs "SecurityContextConstraint" mismatch, because the single is already plural in k8s (see https://github.com/kubernetes/kubernetes/issues/8115#issuecomment-101319887, https://github.com/kubernetes/kubernetes/issues/18622, #81, #261 for background).
I'm not 100% sure which way to fix it, but I think the reason we "fixed" the double plural is merely we needed distinct `get_foo` / `get_foos` *method names* — which is a kubeclient concern that should not leak into `.kind`, it should always be a valid k8s API kind.

---

The wider question I want to discuss (not necessarily here) is what's the purpose of this magic?
- The best use I envision is making it easy to feed the kind back to the API, which is not yet applicable much in kubeclient but is discussed on #208 etc.
  Towards that, I'd like us to provide the non-List .kind (and apiVersion?) **inside** every entity of the array, such that you don't have to care whether you obtained one from via List or Get operation.  (And then potentially restore `get_foos.kind` to be "FooList".)

### LINKED ISSUE #367: create_security_context_constraint doesn't work
This results from same mess as #307 but distinct problem.  We had a workaround for this for `create_endpoint` but `create_security_context_constraint` is broken.   It's an openshift resource, available both in 'v1' core api and a separate api group, neither works:
```
testing_scc = Kubeclient::Resource.new(
  metadata: {
    name: 'teleportation'
  },
  runAsUser: {
    type: 'MustRunAs'
  },
  seLinuxContext: {
    type: 'MustRunAs'
  }
)      

client = Kubeclient::Client.new(context.api_endpoint, context.api_version, ssl_options: context.ssl_options, auth_options: context.auth_options)

pry(main)> client = Kubeclient::Client.new(context.api_endpoint, 'v1', ssl_options: context.ssl_options, auth_options: context.auth_options)
pry(main)> client.create_security_context_constraint(testing_scc)
Kubeclient::HttpError: SecurityContextConstraint in version "v1" cannot be handled as a SecurityContextConstraints: no kind "SecurityContextConstraint" is registered for version "v1"
from /home/bpaskinc/kubeclient/lib/kubeclient/common.rb:125:in `rescue in handle_exception'
Caused by RestClient::BadRequest: 400 Bad Request
from /home/bpaskinc/myenv/rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/rest-client-2.0.2/lib/restclient/abstract_response.rb:223:in `exception_with_response'

pry(main)> sclient = Kubeclient::Client.new(context.api_endpoint + '/apis/security.openshift.io', 'v1', ssl_options: context.ssl_options, auth_options: context.auth_options)
pry(main)> sclient.create_security_context_constraint(testing_scc)
Kubeclient::HttpError: SecurityContextConstraint in version "v1" cannot be handled as a SecurityContextConstraints: no kind "SecurityContextConstraint" is registered for version "security.openshift.io/v1"
from /home/bpaskinc/kubeclient/lib/kubeclient/common.rb:125:in `rescue in handle_exception'
Caused by RestClient::BadRequest: 400 Bad Request
from /home/bpaskinc/myenv/rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/rest-client-2.0.2/lib/restclient/abstract_response.rb:223:in `exception_with_response'
```

Fixing this (with tests) in #366.

cben:
Simpler logic, more explicit about intent (e.g. for the 2 special kinds hardcode the final result instead of intermediate step), added explanations.

Also fixes #307 - `get_security_context_constraints.kind`, `get_endpoints.kind` are now plural as in kubernetes.
Also fixes #367 - `create_security_context_constraint` now works.  (tested live too)

Added lots of tests.
 - [x] added changelog.

@f4tq @eatwithforks please review (easier by commits).

After this, I'll rebase #355 and change it to support both new and old names...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Kubeclient release versioning follows [SemVer](https://semver.org/).
 
 ### Fixed
 - Support custom resources with lowercase `kind` (#361).
+- `create_security_context_constraint` now works (#366).
+- `get_security_context_constraints.kind`, `get_endpoints.kind` are now plural as in kubernetes (#366).
 
 ### Added
 - Add support for retrieving large lists of objects in chunks (#356).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,11 @@ Kubeclient release versioning follows [SemVer](https://semver.org/).
 
 ## Unreleased
 
+### Fixed
+- Support custom resources with lowercase `kind` (#361).
+
 ### Added
-- Add support for retrieving large lists of objects in chunks (#356)
+- Add support for retrieving large lists of objects in chunks (#356).
 
 ## 4.0.0 — 2018-07-23
 
@@ -86,6 +89,6 @@ No changes since 2.5.0, fixed packaging mistake.
 
 ### Added
 
-- `as: raw` option for `get_*` methods returning a string (#262 via #271)
+- `as: raw` option for `get_*` methods returning a string (#262 via #271).
 
 ## 2.4.0 - 2017-05-10
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -435,7 +435,7 @@ module Kubeclient
 
     private
 
-    # Format ditetime according to RFC3339
+    # Format datetime according to RFC3339
     def format_datetime(value)
       case value
       when DateTime, Time
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -132,21 +132,38 @@ module Kubeclient
     end
 
     def self.parse_definition(kind, name)
-      # "name": "componentstatuses", networkpolicies, endpoints
-      # "kind": "ComponentStatus" NetworkPolicy, Endpoints
-      # maintain pre group api compatibility for endpoints and securitycontextconstraints.
-      # See: https://github.com/kubernetes/kubernetes/issues/8115
-      kind = kind[0..-2] if %w[Endpoints SecurityContextConstraints].include?(kind)
+      # Kubernetes gives us have 3 inputs:
+      #   kind: "ComponentStatus"
+      #   name: "componentstatuses"
+      #   singularName: "componentstatus" (usually kind.downcase)
+      # and want to derive singular and plural method names, with underscores:
+      #   "component_status"
+      #   "component_statuses"
+      # kind's CamelCase word boundaries determine our placement of underscores.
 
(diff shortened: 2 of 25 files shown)


cben:
There was also a workaround for `create_endpoint` that I now removed, but was missing for `create_security_context_constraint`, so I believe this PR also fixed that, adding test...

cben:
All done now, PTAL.

cben:
@f4tq @eatwithforks @grosser @NickLaMuro @agrare looking for someone to review this...

cben:
Ouch, I think I caused a regression here for plurals which are not a pure suffix.
For example kind: `WordBoundary`, plural name: `wordboundaries`.
The previous algorithm only assumed a match between `WordB` and `wordb` and would have worked.  The new one expects `"wordboundaries".start_with?("wordboundary")` which doesn't hold :-(
Well now I know why the strange till-last-capital-letter was there :man_facepalming: 

cc @masayag this might be a blocker for #373 — I'd prefer not to complicate it further before we fix this.

cben:
opened #376 with details.  will fix ASAP, but having a very busy week.

PR #373, opened 2018-12-08
Replace '-' with '_' in method names

masayag:
When names of methods aren't conventional, a minor tweak can allow using
the auto-generated methods.

For instance, the CNI defines entity named `NetworkAttachmentDefinition`.
Its resource is `network-attachment-definition`, and to access the

`/apis/k8s.cni.cncf.io/v1/namespaces/default/network-attachment-definitions/ovs-foreman`

```
{
  "namespaced"=>true,
}
```

```

```
This produces unsupported generated method names such as `get_network-attachment-definitions`
and `watch_network-attachment-definitions` which prevents client to access that resource.

The PR proposes to replace dashes with underscores to allow generation
of legal methods names, regardless of the weird naming.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -150,13 +150,19 @@ module Kubeclient
         # But how?  If it differs from kind.downcase, kind's word boundaries don't apply.
         singular_name = kind.downcase
 
-        if name.start_with?(kind.downcase)
+        if name.start_with?(singular_name)
           plural_suffix = name[kind.downcase.length..-1]             # "es"
           singular_underscores = ClientMixin.underscore_entity(kind) # "component_status"
           method_names = [singular_underscores, singular_underscores + plural_suffix]
         else
-          # Something weird, can't infer underscores for plural so just give them up
-          method_names = [singular_name, name]
+          underscored_name = name.tr('-', '_')
+          singular_underscores = ClientMixin.underscore_entity(kind)
+          method_names = if underscored_name.start_with?(singular_underscores)
+                           [singular_underscores, underscored_name]
+                         else
+                           # fallback to lowercase, no separators for both names
+                           [singular_name, underscored_name.tr('_', '')]
+                         end
         end
       end
 
--- a/test/test_common.rb
+++ b/test/test_common.rb
@@ -74,4 +74,19 @@ class CommonTest < MiniTest::Test
     formatted = client.send(:format_datetime, value)
     assert_equal(formatted, '2018-04-30T19:20:33.000000000+00:00')
   end
+
+  def test_parse_definition_with_unconventional_names
+    %w[
+      TestWithDash test-with-dashes test_with_dash test_with_dashes
+      TestUnderscore test_underscores test_underscore test_underscores
+      TestMismatch other-odd-name testmismatch otheroddname
+      MixedDashMinus mixed-dash_minuses mixed_dash_minus mixed_dash_minuses
+      Noseparator noseparators noseparator noseparators
+      SameUptoWordboundary sameup-toword-boundarys sameuptowordboundary sameuptowordboundarys
+    ].each_slice(4) do |kind, plural, expected_single, expected_plural|
+      method_names = Kubeclient::ClientMixin.parse_definition(kind, plural).method_names
+      assert_equal(method_names[0], expected_single)
+      assert_equal(method_names[1], expected_plural)
+    end
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Kubeclient release versioning follows [SemVer](https://semver.org/).
 
 ## Unreleased
 
+- For resources that contain dashes in name, the generated method names will replace the dash in names with underscores.
+
 ### Security
 - Really made `Kubeclient::Config.new(data, nil)` prevent external file lookups.
   README documented this since 3.1.1 (#334) but alas that was a lie — absolute paths always worked.


cben:
Ooh, interesting find!
Can you add a test case?  And a note in CANGELOG.
Also note rubocop complaint (`bundle exec rake` to run locally).

In this case, it's maybe better to rely on `singularName` which is more obviously symmetric with `name`, rather than hoping underscores inferred from capitalization of `kind` matches the dashes in `name`?
Though `singularName` is optional, so we'd also have to fallback onto `kind`, so maybe simpler as you wrote...
OK either way.

masayag:
@cben in terms of reducing any potential of regressions, I'd like to stay with my proposed PR, when the impacts are clear to me.

Fixed travis complains (use 'tr' instead of 'gsub' for performance).

+ added test for that special case

masayag:
@cben when can we expect for a new release that will include this fix ?

cben:
Hopefully tomorrow.

Link to the weird CRD: https://github.com/intel/multus-cni/blob/ca15454d5b76f42ad1c9173959a32ec452536d40/examples/crd.yml

lib/kubeclient/common.rb
```diff
@@ -156,7 +156,7 @@ def self.parse_definition(kind, name)
           method_names = [singular_underscores, singular_underscores + plural_suffix]
         else
           # Something weird, can't infer underscores for plural so just give them up
```
abonas: is the comment still valid? it might be read as confusing along with code that actually adds underscores

masayag: Pls let me know if the updated comment makes more sense or requires any improvements / being removed entirely.

abonas: yes it does, thanks :)
I would say that "something weird" part is too generic and redundant, but it was here before, so I'm not insisting on taking it out.

abonas:
@cben I'm leaving the merge to you, don't want to interfere with your release plans :)

test/test_common.rb
```diff
@@ -74,4 +74,11 @@ def test_format_datetime_with_time
     formatted = client.send(:format_datetime, value)
     assert_equal(formatted, '2018-04-30T19:20:33.000000000+00:00')
   end
+
+  def test_parse_definition_with_dashes_in_resource_name
+    method_names = Kubeclient::ClientMixin.parse_definition('TestNameWithDashes',\
```
cben: Let's use a singular `kind` here — `'TestNameWithDash'`

masayag: @cben fixed.

test/test_common.rb
```diff
@@ -74,4 +74,11 @@ def test_format_datetime_with_time
     formatted = client.send(:format_datetime, value)
     assert_equal(formatted, '2018-04-30T19:20:33.000000000+00:00')
   end
+
+  def test_parse_definition_with_dashes_in_resource_name
+    method_names = Kubeclient::ClientMixin.parse_definition('TestNameWithDash',\
+                                                            'test-name-with-dashes').method_names
+    assert_equal(method_names[0], 'testnamewithdash')
+    assert_equal(method_names[1], 'test_name_with_dashes')
```
cben: Wait, I missed this before, so we'll have `get_testnamewithdash` method for single entity but `get_test_name_with_dashes` for the collection?
I don't like this.

- One direction is for this case to want consistent in `get_test_name_with_dash`, `get_test_name_with_dashes`.
  - one way we could obtain this is to use `[singular_underscores, name.tr('-', '_')]`.  IFF kind's word boundaries match the plural dashes, a bit tricky to detect this...
  - another way is to use singularName!  IFF it's specified, AND we're in "something weird" case so we give up relying on `kind`.  `[singular_name.tr('-', '_'), name.tr('-', '_')]`.

- another direction is to drop the dashes, giving us `get_testnamewithdash`, `get_testnamewithdashes`.

The first direction is prettier — after all somebody went to non-standard trouble :wink: :innocent: to give us word boundaries in both singular and plural case — but I'm not sure how to detect this situation reliably.

PS. I think *predictable* method names are more important consideration than *pretty*.  A dumber algorithm may be better if it's easier to explain?

masayag: @cben, have a look at the updated  version and see if that's make sense to have.
Added few examples to test.
Please let me know if you'd like to include more cases or alter this PR.

lib/kubeclient/common.rb
```diff
@@ -150,13 +150,19 @@ def self.parse_definition(kind, name)
...
-        if name.start_with?(kind.downcase)
+        if name.start_with?(singular_name)
           plural_suffix = name[kind.downcase.length..-1]             # "es"
           singular_underscores = ClientMixin.underscore_entity(kind) # "component_status"
           method_names = [singular_underscores, singular_underscores + plural_suffix]
         else
-          # Something weird, can't infer underscores for plural so just give them up
-          method_names = [singular_name, name]
+          stripped_name = name.tr('-', '').tr('_', '') # clear separators
+          if stripped_name.start_with?(singular_name)
```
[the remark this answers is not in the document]
masayag: Did you mean the opposite ? Since `singular_underscores` is shorten than `name` and fails for all tests.

```
name.tr('-', '_').start_with?(singular_underscores)
```

However, with this change, your suggested test case
`      SameUptoWordboundary sameup-toword-boundarys same_upto_wordboundary sameup_toword_boundarys
` fails.

This test (and other) passes without changing the proposed code.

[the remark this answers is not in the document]
cben: Sorry, yes, opposite.

The test "case" was not really a suggested case to add as-is, just example of `stripped_name.start_with?(singular_name)` behavior, which I think is not good:
```
get_same_upto_wordboundary 
get_sameup_toword_boundarys
```
And I think desired behavior here is the last fallback of dropping the dashes:
```
get_sameuptowordboundary 
get_sameuptowordboundarys
```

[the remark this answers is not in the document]
masayag: Updated the PR to produce the desired naming. Add test with proper expectation for it.

masayag:
@cben should this be merged to master first or should I create a PR for 4.1 branch ?

PR #374, opened 2018-12-09
Add http_max_redirects option

kuahyeow:
There are times when we do not want the client to follow any redirects. This PR adds a `http_max_redirects` option (exposed from `rest-client`) to Kubeclient::Client.

Set `http_max_redirects: 0` to not follow any redirects which will raise an
exception.

### DIFF
--- a/lib/kubeclient/watch_stream.rb
+++ b/lib/kubeclient/watch_stream.rb
@@ -8,6 +8,7 @@ module Kubeclient
         @uri = uri
         @http_client = nil
         @http_options = http_options
+        @http_options[:http_max_redirects] ||= Kubeclient::Client::DEFAULT_HTTP_MAX_REDIRECTS
         @formatter = formatter
       end
 
@@ -38,15 +39,31 @@ module Kubeclient
 
       private
 
+      def max_hops
+        @http_options[:http_max_redirects] + 1
+      end
+
+      def follow_option
+        if max_hops > 1
+          { max_hops: max_hops }
+        else
+          # i.e. Do not follow redirects as we have set http_max_redirects to 0
+          # Setting `{ max_hops: 1 }` does not work FWIW
+          false
+        end
+      end
+
       def build_client
+        client = HTTP::Client.new(follow: follow_option)
+
         if @http_options[:basic_auth_user] && @http_options[:basic_auth_password]
-          HTTP.basic_auth(
+          client = client.basic_auth(
             user: @http_options[:basic_auth_user],
             pass: @http_options[:basic_auth_password]
           )
-        else
-          HTTP::Client.new
         end
+
+        client
       end
 
       def using_proxy
--- a/README.md
+++ b/README.md
@@ -218,6 +218,15 @@ client = Kubeclient::Client.new(
 )
 ```
 
+### Redirects
+
+You can optionally not allow redirection with kubeclient. For example:
+
+```ruby
+client = Kubeclient::Client.new(
+  'https://localhost:8443/api/', http_max_redirects: 0
+)
+```
 
 ### Timeouts
 
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -556,7 +556,8 @@ module Kubeclient
         basic_auth_user: @auth_options[:username],
         basic_auth_password: @auth_options[:password],
         headers: @headers,
-        http_proxy_uri: @http_proxy_uri
+        http_proxy_uri: @http_proxy_uri,
+        http_max_redirects: http_max_redirects
       }
 
       if uri.scheme == 'https'
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -34,6 +34,7 @@ module Kubeclient
     }.freeze
 
     DEFAULT_HTTP_PROXY_URI = nil
+    DEFAULT_HTTP_MAX_REDIRECTS = 10
 
     SEARCH_ARGUMENTS = {
       'labelSelector' => :label_selector,
@@ -52,6 +53,7 @@ module Kubeclient
     attr_reader :ssl_options
     attr_reader :auth_options
     attr_reader :http_proxy_uri
+    attr_reader :http_max_redirects
     attr_reader :headers
     attr_reader :discovered
 
@@ -64,6 +66,7 @@ module Kubeclient
       socket_options: DEFAULT_SOCKET_OPTIONS,
       timeouts: DEFAULT_TIMEOUTS,
       http_proxy_uri: DEFAULT_HTTP_PROXY_URI,
+      http_max_redirects: DEFAULT_HTTP_MAX_REDIRECTS,
       as: :ros
     )
       validate_auth_options(auth_options)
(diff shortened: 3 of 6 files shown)


kuahyeow:
I'm not sure if exposing as `http_max_redirects: <integer>` is the best idea. Alternative would be have a simpler `http_follow_redirects: <boolean>` option.

cben:
Thanks!
Does this also cover places we use `http` gem instead of `rest-client` — watch, and maybe logs(?)
http gem has `max_hops` option: https://github.com/httprb/http/blob/da68f40182f4230cad0a43b4196e7906c34516df/lib/http/redirector.rb#L41

rest-client is unmaintained, so it seems likely at some point we'll switch (#237)...
Do all gems support "redirect but only *k* times"?  Would exposing a boolean be more future-proof?

OTOH, are there situations where one wants to *increase* default max redirects?  rest-client and httpclient default to 10, http gem to only 5, what if one needs more?
=> I'm fine with exposing an integer knob.  (And if so let's set it to 10 for get/watch parity)

kuahyeow:
Oh, I did not know about `http` usage. Will look into that.

kuahyeow:
`watch` seems to [raise on 30x](https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/watch_stream.rb#L19), so it seems moot to pass in redirect options. This seems to be the only place where `http` gem is  used

`get_pod_logs` uses the `rest-client` gem so it should honor the option.

> OTOH, are there situations where one wants to increase default max redirects? rest-client and httpclient default to 10, http gem to only 5, what if one needs more?
> => I'm fine with exposing an integer knob. (And if so let's set it to 10 for get/watch parity)

OK, have updated `DEFAULT_HTTP_MAX_REDIRECTS` to `10`

cben:
> watch seems to raise on 30x, so it seems moot to pass in redirect options. 

Untested, but I think that's the response code of the last request, after the gem handled redirects internally?  And when max_hops is exceeded, it'll raise
TooManyRedirectsError or EndlessRedirectError.  Kubeclient will not see 3xx.

That line goes back to #142 which first switched to http gem.  It was not discussed there.
I dont think `< 300`  test had any intent to raise on redirects, it's just a way to check for 2xx statuses.

kuahyeow:
> Untested, but I think that's the response code of the last request, after the gem handled redirects internally? And when max_hops is exceeded, it'll raise
TooManyRedirectsError or EndlessRedirectError. Kubeclient will not see 3xx.

Interesting, I have added a test to try this out. Without modifications to `WatchStream`, it fails. So I think `http` by default does not follow redirects and then we do that check for `<300`

```
  1) Error:
TestPodLog#test_watch_pod_log_with_redirect:
Kubeclient::HttpError: Found
    /Users/tkuah/code/kubeclient/lib/kubeclient/watch_stream.rb:20:in `each'
    /Users/tkuah/code/kubeclient/test/test_pod_log.rb:96:in `with_index'
    /Users/tkuah/code/kubeclient/test/test_pod_log.rb:96:in `test_watch_pod_log_with_redirect'
```


```
+
+  def test_watch_pod_log_with_redirect
+    expected_lines = open_test_file('pod_log.txt').read.split("\n")
+
+    stub_request(:get, %r{/namespaces/default/pods/[a-z0-9-]+/log\?.*follow})
+      .to_return(status: 302, headers: { location: 'http://localhost:1234/api/namespaces/default/pods/redis-master-pod/log' })
+
+    stream = client.watch_pod_log('redis-master-pod', 'default')
+    stream.to_enum.with_index do |notice, index|
+      assert_instance_of(String, notice)
+      assert_equal(expected_lines[index], notice)
+    end
+  end
```

To confirm, I tried out `http` in `irb`:

```
irb(main):003:0> client = HTTP::Client.new
=> #<HTTP::Client:0x00007fd4533503c0 @default_options=#<HTTP::Options:0x00007fd453350118 @response=:auto, @proxy={}, @timeout_class=HTTP::Timeout::Null, @timeout_options={}, @socket_class=TCPSocket, @nodelay=false, @ssl_socket_class=OpenSSL::SSL::SSLSocket, @ssl={}, @keep_alive_timeout=5, @headers=#<HTTP::Headers {}>, @cookies={}, @encoding=nil, @features={}>, @connection=nil, @state=:clean>
irb(main):004:0> client.request(:get, 'https://httpstat.us/302')
=> #<HTTP::Response/1.1 302 Found {"Cache-Control"=>"private", "Location"=>"https://httpstat.us", "Server"=>"Microsoft-IIS/10.0", "X-Aspnetmvc-Version"=>"5.1", "Access-Control-Allow-Origin"=>"*", "X-Aspnet-Version"=>"4.0.30319", "X-Powered-By"=>"ASP.NET", "Set-Cookie"=>"ARRAffinity=77c477e3e649643e5771873e1a13179fb00983bc73c71e196bf25967fd453df9;Path=/;HttpOnly;Domain=httpstat.us", "Date"=>"Wed, 19 Dec 2018 11:03:21 GMT", "Connection"=>"close", "Content-Length"=>"0"}>
```

```
irb(main):005:0> client.follow.request(:get, 'https://httpstat.us/302')
=> #<HTTP::Response/1.1 200 OK {"Cache-Control"=>"private", "Content-Length"=>"7959", "Content-Type"=>"text/html; charset=utf-8", "Server"=>"Microsoft-IIS/10.0", "X-Aspnetmvc-Version"=>"5.1", "Access-Control-Allow-Origin"=>"*", "X-Aspnet-Version"=>"4.0.30319", "X-Powered-By"=>"ASP.NET", "Set-Cookie"=>"ARRAffinity=77c477e3e649643e5771873e1a13179fb00983bc73c71e196bf25967fd453df9;Path=/;HttpOnly;Domain=httpstat.us", "Date"=>"Wed, 19 Dec 2018 11:04:19 GMT", "Connection"=>"close"}>
```

So to be consistent, we should follow redirects for `WatchStream` too ? If OK, I can perform the change.

cben:
Oh, good findings.

Yes, we want watch to be consistent with other methods
 (with one exception: we dont want watch to timeout, deliberately)

lib/kubeclient/watch_stream.rb
```diff
@@ -8,6 +8,7 @@ def initialize(uri, http_options, formatter:)
         @uri = uri
         @http_client = nil
         @http_options = http_options
+        @http_options[:http_max_redirects] ||= Kubeclient::Client::DEFAULT_HTTP_MAX_REDIRECTS
```
kuahyeow: Referring to `Kubeclient::Client::DEFAULT_HTTP_MAX_REDIRECTS` feels slightly icky here, but this is the most direct way to get a consistent default

cben: I _think_ this could be omitted as WatchStream is normally passed what http_options returns.  I'll play with it later today.

lib/kubeclient/watch_stream.rb
```diff
@@ -38,15 +39,31 @@ def finish
...
+          { max_hops: max_hops }
+        else
+          # i.e. Do not follow redirects as we have set http_max_redirects to 0
+          # Setting `{ max_hops: 1 }` does not work FWIW
+          false
+        end
+      end
+
       def build_client
+        client = HTTP::Client.new(follow: follow_option)
```
cben: Perhaps this could be moved into constructor, or just even http_options.

cben:
Merged, if I want I'll do a followup PR.

@kuahyeow Let me know if you want a new gem release with this.

kuahyeow:
Thanks @cben ! :)

> Let me know if you want a new gem release with this.

Yes please, a new gem release would be much appreciated (so that we can drop our monkey-patch 😅)

cben:
:heavy_check_mark: released 4.2.2.

kuahyeow:
Thanks @cben - it has been awesome working with you! 💚

PR #375, opened 2018-12-09
Credential plugins: lookup command relative to config's dir, nil = opt-out

cben:
Follow-up to #363.

Refine lookup to match [Go client's behavior](https://github.com/kubernetes/kubernetes/pull/59495#discussion_r171138995): Distinguish 3 cases:
- absolute (e.g. `/path/to/foo`)
- `$PATH`-based (e.g. `curl`)
- relative to config file's dir, or specified base dir (e.g. `./foo`)

If base dir explicitly set to nil, refuse to execute external commands, matching existing opt-out from file lookups (#372).

Document security aspects of credential plugins.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,3 +25,5 @@ Security/MarshalLoad:
 Style/MethodCallWithArgsParentheses:
   IgnoredMethods:
   - require_relative
+Style/RegexpLiteral:
+  Enabled: false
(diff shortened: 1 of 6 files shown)


cben:
@motymichaely @KnVerey @rhodrid @f4tq please review.

jeremywadsack:
@cben Looks good. I'll be excited when GCP moves this this instead of the customer provider.

cben:
Released 4.2.0.

PR #382, opened 2018-12-26
Fallback to lower-case method names

masayag:
When names of methods aren't conventional, there will be an attempt to
produce method names based on the following logic:

1. If the plural name without separators starts with lower-cased 'kind',
   use underscores as separators (assuming plural has separators)
2. In case of a mismatch, remove any separators and plural name will be
   the lower-case of it.

For instance, the CNI defines entity named NetworkAttachmentDefinition.
Its resource is 'network-attachment-definition', so to access the

/apis/k8s.cni.cncf.io/v1/namespaces/default/network-attachment-definitions

{
  "namespaced"=>true,
}

This produces unsupported generated method names such as 'get_network-attachment-definitions'
and 'watch_network-attachment-definitions' which prevents client to access that resource.

Using the PR, the generated method names will be:
'get_network_attachment_definition' and 'get_network_definitions'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,14 @@ Notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 Kubeclient release versioning follows [SemVer](https://semver.org/).
 
+## Unreleased
+
+### Fixed
+- For resources that contain dashes in name, there will be an attempt to resolve the method name based on singular name prefix or by replacing the dash in names with underscores.
+
 ## 4.1.1 — 2018-12-17
 
 ### Fixed
-
 - Fixed method names for non-suffix plurals such as y -> ies  (#377).
 
 ## 4.1.0 — 2018-11-28 — REGRESSION
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -163,8 +163,7 @@ module Kubeclient
             method_names = [prefix_underscores + singular_suffix,      # "network_policy"
                             prefix_underscores + plural_suffix]        # "network_policies"
           else
-            # Something weird, can't infer underscores for plural so just give them up
-            method_names = [singular_name, name]
+            method_names = resolve_unconventional_method_names(name, kind, singular_name)
           end
         end
       end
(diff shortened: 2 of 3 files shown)


masayag:
@cben Please review

cben:
reviewing.  no additional PRs needed.

CHANGELOG.md
```diff
@@ -4,10 +4,14 @@ Notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 Kubeclient release versioning follows [SemVer](https://semver.org/).
 
-## 4.1.1 — 2018-12-17
+## 4.1.1 — 2018-12-26
```
cben: can't have 4.1.1 twice.  Make this `## Unreleased` or `## 4.2.1`, I'll set the version & date in a later PR for the release version bump.

hmm, you're still based on v4.1.z branch.  you should merge/rebase fresh master, put the new version on top, before 4.2.0 (it'll become 4.2.1).
(unless you have a need for this fix to also be released on 4.1 branch?)

masayag: fog-kubevirt has the following dependency on kubeclient: "kubeclient", "~> 4.1.0"
So I'd like it to be included and released for 4.1.z (more than have it on 4.2.0 in terms on necessity)

masayag: @cben PR for master is here: https://github.com/abonas/kubeclient/pull/383

lib/kubeclient/common.rb
```diff
@@ -176,6 +175,23 @@ def self.parse_definition(kind, name)
       )
     end
 
+    def self.resolve_unconventional_method_names(name, kind, singular_name)
+      if name.start_with?(singular_name)
+        plural_suffix = name[kind.downcase.length..-1]             # "es"
+        singular_underscores = ClientMixin.underscore_entity(kind) # "component_status"
```
cben: if it makes sense, could you update comments on the right to some realistic example that would get into this function?
I think ComponentStatus / componentstatuses is covered by normal `name.start_with?(prefix.downcase)` flow.

cben:
LGTM :100: thanks for sticking through the iterations, which were largely my fault!

masayag:
Thank you @cben

cben:
released 4.1.2.

PR #383, opened 2018-12-26
Fallback to lower-case method names

masayag:
When names of methods aren't conventional, there will be an attempt to
produce method names based on the following logic:

1. If the plural name without separators starts with lower-cased 'kind',
   use underscores as separators (assuming plural has separators)
2. In case of a mismatch, remove any separators and plural name will be
   the lower-case of it.

For instance, the CNI defines entity named NetworkAttachmentDefinition.
Its resource is 'network-attachment-definition', so to access the

/apis/k8s.cni.cncf.io/v1/namespaces/default/network-attachment-definitions

{
  "namespaced"=>true,
}

This produces unsupported generated method names such as 'get_network-attachment-definitions'
and 'watch_network-attachment-definitions' which prevents client to access that resource.

Using the PR, the generated method names will be:
'get_network_attachment_definition' and 'get_network_definitions'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 Kubeclient release versioning follows [SemVer](https://semver.org/).
 
+## Unreleased
+
+### Fixed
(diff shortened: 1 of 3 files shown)


cben:
same as #382, LGTM.
will release this in 4.2.1 soon.

cben:
released.

PR #384, opened 2018-12-26
Allow using Gemfile.dev.rb for development

masayag:
In order to allow developers introducing local gems for development
purposes (i.e. debug), a file named `Gemfile.dev.rb` can be created
and include gems settings.


masayag:
@cben Do you recommend on adding an empty or commented `Gemfile.dev.rb` ?
Are you okay with `Gemfile.dev.rb` or would you prefer `Gemfile.local.rb` ?

cben:
WAT is going on in that Travis build?!

masayag:
> WAT is going on in that Travis build?!

Not sure, seems like the change to `.rubocop.yml` triggered rubocop for all sources and not only for latest changes ? I haven't experienced that locally:

masayag@masayag: ~/work/kubeclient (gemfile_dev)$ bundle exec rake rubocop
Running RuboCop...
Inspecting 41 files
.........................................

41 files inspected, no offenses detected

------
I'll try to use a different 'Exclude' syntax...maybe this is the case. else, I'll just 'Gemfile.dev.rb' only to  `Style/MethodCallWithArgsParentheses:`

cben:
Don't care about .dev vs .local
Don't think you can add an empty one to git, while still ignoring it (which is the whole purpose here, to be able to add gems like `byebug` but keep clean `git status`)

masayag:
> Don't think you can add an empty one to git, while still ignoring it (which is the whole purpose here, to be able to add gems like byebug but keep clean git status)

TA-DA ;-)
'git add -f Gemfile.dev.rb' did the trick.

So we can discuss the content of `Gemfile.dev.rb` - should it be empty or left with commented lines.

masayag:
@cben do you see any further changes needed ?

masayag:
@cben travis was unhappy with adding 'Exclue: [Gemfile.dev.rb]' under the 'AllCops' elements.
When I moved the exclusion to a specific rule, it seems to be satisfied with the change.

Issue #268, opened 2017-09-13
Allow passing headers per request


grosser:
Atm I'm trying to update statefulset so I need to do this hack-around which is a bit ugly ...

```
      def deploy
        ....
        with_patch_header do
          client.patch_statefulset name, namespace, [{op: "replace", path: "/spec", value: @template.fetch(:spec)}]
        end
      end

      private
      
      def with_patch_header
        old = client.headers['Content-Type']
        client.headers['Content-Type'] = 'application/json-patch+json'
        yield
      ensure
        client.headers['Content-Type'] = old
      end
```

moolitayer:
We have patch entity, would that serve your use case @grosser ?
```ruby
    def patch_entity(resource_name, name, patch, namespace = nil)
      ns_prefix = build_namespace_prefix(namespace)
      handle_exception do
        rest_client[ns_prefix + resource_name + "/#{name}"]
          .patch(
            patch.to_json,
            { 'Content-Type' => 'application/strategic-merge-patch+json' }.merge(@headers)
          )
      end
    end
```

grosser:
I need to change the `Content-Type` ... I'm using patch_entity already

moolitayer:
Oh right, so it's ` 'Content-Type' => 'application/strategic-merge-patch+json'` vs `'application/json-patch+json'`? Can you explain what is the difference? I think it was discussed in https://github.com/abonas/kubeclient/pull/161 but don't remember the distinction.

grosser:
idk ... was following an example and the default one did not work ... have not looked any deeper :D

cben:
doc: https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/

Are there other use cases than patch to allow direct setting of headers?
My instinct is we're clearly missing functionality in `patch_*()` and should add it, and defer the general escape hatch until proven necessary.

Current signature is `patch_foo(name, patch, namespace = nil)`.  It's awkward that we have name & namespace not adjacent :-(, and I'd like to avoid adding a 4th positional arg to it.  Cf. #312.

stormbeta:
@cben That awkwardness with the method signature is why I didn't try to add it in a PR myself - wasn't sure what the best way to handle it would be.

It's also worth noting that some CustomResourceDefinitions may require this if they don't support strategic merge.

cben:
[...]

PR #390, opened 2019-01-24
Add merge and json patch

### LINKED ISSUE #357: Support json patch & json merge patch in addition to strategic patch
`Error message: the body of the request was in an unknown format - accepted media types include: application/json-patch+json,application/merge-patch+json`

Using 

`@atcloud.patch_service_check(check.metadata.name, {metadata: {annotations: {key: 'value'}}}, check.metadata.namespace)`

Any ideas?

tsontario:
Closes #357 

Following the conversation on #357 and #268 , I figured I could start working on this feature as my first contribution to kubeclient. As @cben proposed, I'm simply adding `json_patch_*` and `merge_patch_*` methods to the client.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -5,7 +5,7 @@ module Kubeclient
   # Common methods
   # this is mixed in by other gems
   module ClientMixin
-    ENTITY_METHODS = %w[get watch delete create update patch].freeze
+    ENTITY_METHODS = %w[get watch delete create update patch json_patch merge_patch].freeze
 
     DEFAULT_SSL_OPTIONS = {
       client_cert: nil,
@@ -203,6 +203,7 @@ module Kubeclient
       namespace.to_s.empty? ? '' : "namespaces/#{namespace}/"
     end
 
+    # rubocop:disable  Metrics/BlockLength
     def define_entity_methods
       @entities.values.each do |entity|
         # get all entities of a type e.g. get_nodes, get_pods, etc.
@@ -238,11 +239,23 @@ module Kubeclient
           update_entity(entity.resource_name, entity_config)
         end
 
-        define_singleton_method("patch_#{entity.method_names[0]}") do |name, patch, namespace = nil|
-          patch_entity(entity.resource_name, name, patch, namespace)
+        define_singleton_method("patch_#{entity.method_names[0]}") \
+        do |name, patch, namespace = nil|
+          patch_entity(entity.resource_name, name, patch, 'strategic-merge-patch', namespace)
+        end
+
+        define_singleton_method("json_patch_#{entity.method_names[0]}") \
+        do |name, patch, namespace = nil|
+          patch_entity(entity.resource_name, name, patch, 'json-patch', namespace)
+        end
+
+        define_singleton_method("merge_patch_#{entity.method_names[0]}") \
+        do |name, patch, namespace = nil|
+          patch_entity(entity.resource_name, name, patch, 'merge-patch', namespace)
         end
       end
     end
+    # rubocop:enable  Metrics/BlockLength
 
     def self.underscore_entity(entity_name)
       entity_name.gsub(/([a-z])([A-Z])/, '\1_\2').downcase
@@ -381,13 +394,13 @@ module Kubeclient
       format_response(@as, response.body)
     end
 
-    def patch_entity(resource_name, name, patch, namespace = nil)
+    def patch_entity(resource_name, name, patch, strategy, namespace)
       ns_prefix = build_namespace_prefix(namespace)
       response = handle_exception do
         rest_client[ns_prefix + resource_name + "/#{name}"]
           .patch(
             patch.to_json,
-            { 'Content-Type' => 'application/strategic-merge-patch+json' }.merge(@headers)
+            { 'Content-Type' => "application/#{strategy}+json" }.merge(@headers)
           )
       end
       format_response(@as, response.body)
--- a/test/test_service.rb
+++ b/test/test_service.rb
@@ -273,4 +273,58 @@ class TestService < MiniTest::Test
       data['metadata']['annotations']['key'] == 'value'
     end
   end
+
+  def test_json_patch_service
+    service = Kubeclient::Resource.new
+    name = 'my-service'
+
+    service.metadata = {}
+    service.metadata.name      = name
+    service.metadata.namespace = 'development'
+
+    stub_core_api_list
+    expected_url = "http://localhost:8080/api/v1/namespaces/development/services/#{name}"
+    stub_request(:patch, expected_url)
+      .to_return(body: open_test_file('service_json_patch.json'), status: 200)
+
+    patch = [
+      { 'op' => 'add', 'path' => '/spec/type', 'value' => 'LoadBalancer' }
+    ]
+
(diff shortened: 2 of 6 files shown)


lib/kubeclient/common.rb
```diff
@@ -393,6 +401,30 @@ def patch_entity(resource_name, name, patch, namespace = nil)
...
+      ns_prefix = build_namespace_prefix(namespace)
+      response = handle_exception do
+        rest_client[ns_prefix + resource_name + "/#{name}"]
+          .patch(
+            patch.to_json,
+            { 'Content-Type' => 'application/merge-patch+json' }.merge(@headers)
+          )
+      end
+      format_response(@as, response.body)
+    end
```
cben: The considerations from #357 that led to separate public `*patch` methods don't apply for these internal undocumented methods.  Consider merging them into one method with 5 args.  (Also, you can make `namespace` mandatory).

test/test_service.rb
```diff
@@ -273,4 +273,66 @@ def test_patch_service
...
+    stub_request(:patch, expected_url)
+      .to_return(body: open_test_file('service_patch.json'), status: 200)
+
+    patch = {
+      metadata: {
+        annotations: {
+          key: 'value'
+        }
+      }
+    }
```
cben: JSON patch takes different format — list of {op, path, value} tuples.  https://tools.ietf.org/html/rfc6902
I think here you want:
```
patch = [
  {op: "replace", path: "/metadata/annotations/key", value: "value"}
]
```

test/test_service.rb
```diff
@@ -273,4 +273,66 @@ def test_patch_service
...
```
cben: (This one can stay same as strategic `patch` :+1:
When arrays are involved *result* might differ but we're testing kubeclient not k8s here, expressing this difference would require extra effort with mock results but won't make the test any stronger.)

lib/kubeclient/common.rb
```diff
@@ -241,6 +241,14 @@ def define_entity_methods
...
           patch_entity(entity.resource_name, name, patch, namespace)
         end
+
+        define_singleton_method("json_patch_#{entity.method_names[0]}") do |name, patch, namespace = nil|
+          json_patch_entity(entity.resource_name, name, patch, namespace)
+        end
+
+        define_singleton_method("merge_patch_#{entity.method_names[0]}") do |name, patch, namespace = nil|
+          merge_patch_entity(entity.resource_name, name, patch, namespace)
+        end
```
cben: Travis failed on rubocop.

```
lib/kubeclient/common.rb:207:7: C: Metrics/BlockLength: Block has too many lines. [30/25]
```
It's this block, of all the `define_singleton_method`s.  It's perfectly readable.
Can you find a way to suppress this?

```
lib/kubeclient/common.rb:245:101: C: Metrics/LineLength: Line is too long. [105/100]
        define_singleton_method("json_patch_#{entity.method_names[0]}") do |name, patch, namespace = nil|
                                                                                                    ^^^^^
lib/kubeclient/common.rb:249:101: C: Metrics/LineLength: Line is too long. [106/100]
        define_singleton_method("merge_patch_#{entity.method_names[0]}") do |name, patch, namespace = nil|
                                                                                                    ^^^^^^
```
You can copy the ugly ` \ ` style from `get_` and `delete_` above :-/
Could also try creating shorter vars for `entity.method_names[1]` and `entity.method_names[0]`.

Or apply `rubocop:disable` comments around this whole function.  These only add noise to the code :-(, but this particular block has been annoying contributors for some time...

timothysmith0609: Usually I'm not a fan of going around rubocop but I think with regards to the block length it makes sense to do so.

Concerning line length, 100 seems a bit draconian to me. I think something closer to 120 still maintains readability and gets us that much more space. For now, I'll adopt the `/` convention to stay in line with the rest of the code

test/test_service.rb
timothysmith0609: >but we're testing kubeclient not k8s here

In this case the line feels a little blurry, since I'm quite interested that the response being sent back by the server is valid. It looks like `test_patch_service` is completely stubbing out the `patch` call. In fact, I can remove the `content-type` header from the code and `test_patch_service` will still pass.

One approach I could take is to create a fixture by executing the code against a cluster so at least we have a valid snapshot. I see that VCR is used in some places in this repo but understand that it has its own drawbacks as well.

I'm interested to hear your thoughts here. In the worst case I'll get a working test that won't fail if a regression is introduced, since we're completely stubbing that part out. On the other hand, it does at least show that the client is discovering the appropriate method and making it available on resources 🤷‍♂️

lib/kubeclient/common.rb
timothysmith0609: >you can make namespace mandatory

I'm surprised it wasn't to begin with since it would seem that [`build_ns_prefix(namespace)`](https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/common.rb#L202) would crash in that case (called from[here](https://github.com/abonas/kubeclient/blob/master/lib/kubeclient/common.rb#L385)).

Making it mandatory is probably the best bet, since then it's really the responsibility of the caller at that point

cben: Sorry, I was unclear. Sure, for the public methods is must be optional.  
I meant on the `patch_entity` internal method, it's always passed a namespace arg, even if nil.  
(And by mandatory I meant at signature level, dropping the `= nil`, but still doing `build_namespace_prefix` supporting nil)

Never mind, fine either way.

test/test_service.rb
```diff
@@ -273,4 +273,63 @@ def test_patch_service
...
+    service.metadata.namespace = 'development'
+
+    stub_core_api_list
+    expected_url = "http://localhost:8080/api/v1/namespaces/development/services/#{name}"
+    stub_request(:patch, expected_url)
+      .to_return(body: open_test_file('service_patch.json'), status: 200)
+
+    patch = [
+      { 'op' => 'replace', 'path' => '/metadata/annotations/key1', 'value' => 'value' },
+      { 'op' => 'add', 'path' => '/metadata/annotations/key2', 'value' => 'value' }
```
cben: in principle, once you use functionally different patch here, ideally you'd mock a different return result than `service_patch.json`.  If you don't want that, nor want to patch only `/metadata/annotations/key`, please add a comment on the `to_return` saying it's inaccurate result for this patch.

test/test_service.rb
cben: > I can remove the content-type header from the code and test_patch_service will still pass.

Oh, something ate my comment with suggestions for testing Content-type.  (untested, only googled)
One way is add after the `stub_request`:
```
  .with(headers: { 'Content-Type' => '...'})
```
another is extend the condition in `assert_requested`:
```
  req.headers['Content-Type'] == '...' &&
  data[...] == ...
```

timothysmith0609:
Ok, this is ready for another round of review.

- [x] Updated README
- [x] Updated CHANGELOG
- [x] Separate fixtures for different patches
- [x] Ran fixture tests locally against minikube instance
- [x] Remove `namespace = nil` in [patch_entity](https://github.com/abonas/kubeclient/pull/390/files#diff-104b3d1973988933ffc83feada470da2R397) (since it is in fact redundant)

tsontario:
Squashed and ready to go. Thanks for the solid feedback and looking forward to the next PR.

cben:
released in gem 4.3.0.

PR #391, opened 2019-01-26
Positional and params support

### LINKED ISSUE #312: Harmonize positional vs keyword namespace: args
In some method we take namespace as optional positional arg, in some as optional keyword `namespace:` arg.  This is error-prone, and doesn't scale as we want to add more optional args.  (It'd be especially awkward if you'd need to pass dummy nil namespace when dealing with a non-namespaced entity such as nodes...)

Long-term I'd like to support keyword `namespace:` everywhere, initially still supporting positional where needed for compatibility, eventually deprecating positional.

tsontario:
### Goals

Aims to close #312 (Harmonize positional vs keyword namespace args).

For now, I'm looking for validation of the approach or whether or not we want to try solving this a different way.

### Proposed Solution

This solution leverages (abuses?) the fact that ruby captures all named parameters in **kwargs if all other parameters are provided default values. By doing this, we add the ability to supply named parameters to `get`, `delete`, `patch`, etc. while still supporting the previous unnamed parameter implementation. This should provide the means to gracefully deprecate unnamed parameters for these methods moving forward. Once that's done, we can explicitly put the named parameters in the method signatures and clean up the logic for dealing with both cases.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -221,12 +221,12 @@ module Kubeclient
 
         # get a single entity of a specific type by name
         define_singleton_method("get_#{entity.method_names[0]}") \
-        do |name, namespace = nil, opts = {}|
+        do |name = nil, namespace = nil, **opts|
           get_entity(entity.resource_name, name, namespace, opts)
         end
 
         define_singleton_method("delete_#{entity.method_names[0]}") \
-        do |name, namespace = nil, opts = {}|
+        do |name = nil, namespace = nil, **opts|
           delete_entity(entity.resource_name, name, namespace, opts)
         end
 
@@ -238,8 +238,9 @@ module Kubeclient
           update_entity(entity.resource_name, entity_config)
         end
 
-        define_singleton_method("patch_#{entity.method_names[0]}") do |name, patch, namespace = nil|
-          patch_entity(entity.resource_name, name, patch, namespace)
+        define_singleton_method("patch_#{entity.method_names[0]}") \
+        do |name = nil, patch = nil, namespace = nil, **opts|
+          patch_entity(entity.resource_name, name, patch, namespace, opts)
         end
       end
     end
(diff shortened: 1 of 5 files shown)


timothysmith0609:
>I'd like unrecognized options to still raise an error.

Definitely agree with you there and it shouldn't be difficult to enforce this (via your suggestions or some similar mechanism). I think we can probably do away with `**args` and just default `name:, namespace:, etc` to nil. This will increase the complexity of the methods in the short term (via duplication of name/namespace in both positional and keyword sections) but provides a convenient deprecation path without excessive code additions.

Do you have an idea of what the deprecation timeline will look like and is there an established manner in communicating these changes? Do we just use the CHANGELOG or actually print warnings to users?

Regarding the tests, I wouldn't pay too much attention to them now as they're just there to sanity-check my solution. I'm conflicted about using the same cassette for both approaches since I prefer to fully encapsulate my tests, but I could be convinced that this case would benefit from your idea.

cben:
So far we haven't bothered with active warnings, simply CHANGELOG + breaking change in major release.  Better to add new way & document upcoming change ahead of that when possible, of course.  
We could do warnings if desired.

Do you mean deprecating positional name, namespace?  
Why?
I'm not sure there is payoff to justify the incompatibility.  For observable future I'd keep the exisiting inconsistent support for positional.
We can encourage consistent keyword style in docs.

tsontario:
>Do you mean deprecating positional name, namespace?
Why?

I think I need some clarification here. My goal with this PR is to support _both_ positional and keyword parameters for `name, namespace` in the hope that, at some later date, we will fully discontinue positional parameters. What I don't want is to support both positional/keywords in perpetuity. The change to keywords is clearly a breaking change and would be a feature of a major release.

To reiterate, I'm ok with supporting the inconsistent as-is for now, but at some point we should enforce keywords-only everywhere.

cben:
Understood.  I don't yet see a significant price to supporting both in perpetuity, but it's quite possible technical debt will pile up with time...  We don't need to agree on that now as this is good direction for the code anyway :+1: :grinning:

tsontario:
>I don't yet see a significant price to supporting both in perpetuity

For me, this PR is about finding a canonical way to pass the parameters to the client functions (whether that ends up being positional or params is not really important at this point). I'm uncomfortable providing multiple means of doing the same thing for too long, and want to do it here only as a means to provide a graceful deprecation path. 

In particular, the means in which I want to go about this is to provide an around alias on top of the current get/watch/etc. operations that accepts both inputs and, once deprecation is complete, we can just eliminate those around aliases and the underlying methods will have the proper contract. Keeping that around for a long time is dangerous as there's a lot of indirection for very little gain (e.g. it's hard to see that we're in fact aliasing the methods in the first place)

cben:
Can you resolve conflicts, and check rubocop?
(feel free to increase, or locally disable, limits like PerceivedComplexity.  They're a useful reminder to humans but unsuitable as hard CI-enforced rules.)

This is good stuff and I'd like to land it; we can always improve later.

timothysmith0609:
Apologies, I've been quite busy lately so this fell off my radar. I'll try to take a look at this in the next 2 weeks and iron everything out to get it in shape for release.

tsontario:
Closing this as I don't foresee myself completing this work.

Issue #392, opened 2019-02-15
Automatically support GCP ApplicationDefaultCredentials


timothysmith0609:
Continuation of #213 

It seems like a good idea to have Kubeclient automatically default to `GoogleApplicationDefaultCredentials` when the provided context's user has the `auth-provider: 'gcp'` entry. Currently, specifying application default credentials must be done explicitly. This is problematic for tools that don't know a priori what cloud provider is being targeted (e.g. [Shopify/kubernetes-deploy](https://github.com/Shopify/kubernetes-deploy)). 

Our current solution in kubernetes deploy is to subclass `Kubeclient::Config` and add our own [shim](https://github.com/Shopify/kubernetes-deploy/pull/418/files#diff-2afdf2c15d5f6b0538e430a95ad61e3bR6). I don't see a strong argument against fixing this upstream here, since the required code change is small and the default can always be overridden by the end user in `Client.new :auth_options`.

Thoughts? I know there's been a lot of discussion on this issue in the past, and `exec` support for GCP is apparently in the pipeline, though it's hard to say how far out that is.

cben:
Generally :+1: to make config Just Work!
And avoid users having to subclass Config - I gather that's proving brittle (and it wasn't consciously designed for extension by subclassing)...

I see you wrote on https://github.com/Shopify/kubernetes-deploy/issues/406 "The ability to check for this case is hidden behind the kubeclient interface so overriding is really our only choice at the moment" -- can you elaborate? Would this fully resolve that problem?

There was some concern last time about taking on google gem dependencies which are pretty large (bigger then kubeclient's current deps iirc?).  The current "soft dependency" situation is partially compromise for that but it's not too pretty :-/

That's not a strict blocker - if we need the dependencies, so be it.

timothysmith0609:
>The ability to check for this case is hidden behind the kubeclient interface so overriding is really our only choice at the moment" -- can you elaborate?

The issue for kubernetes-deploy is that it's agnostic about what cloud provider the end-user is targeting. For example, if we knew we would always be targetting a gcp cluster, then we could use the kubeclient ApplicationDefaultCredentials as it stands. But we need a way to find out, at runtime, which cloud provider we are dealing with. This information is encapsulated by private methods in the `Config` class (namely, `fetch_user_auth_options` or, at a higher level, `fetch_context`). The fact that `@kfcg` is private spurred the need for our subclassing and overriding strategy.

[...]

Issue #393, opened 2019-02-16
Authentication renewal


cben:
Some credentials have short shelf life. GCP auth is good for about a hour, I think(?).  `exec` results may return expiration time, or may simply return error at some point.

Since one might have passed around a Kubeclient::Client object to many places,
it would be desirable to be able to renew auth - manually or automatically - without creating a new Client.

cben:
# 1. Mutable `Client` or external mutable object holding credentials?

Would a first step of making `Client` mutable, at least for auth_options, be helpful?
No.  Presently you need separate `Client` object per API group, but you want to renew credentials only once.  Having to keep track of each Client and patching them all would negate the gains.

- Conclusion: we want a new object responsible for auth (not giving it a name yet).
  Many Client objects will be able to point to a single auth object.
- Conclusion: we want `Client` to ask the auth object for credentials *before every API request*.

cben:
## 2. Will the current way users call construct `Client` by passing in `auth_options: config.context.auth_options` etc. need to change?

Probably.  We might be able to smuggle an auth object instead of passive data under a new key in auth_options, but not sure I want that.

Anyway, this will be a huge enough win to warrant an API change, I'm not worried about this for now :see_no_evil:

cben:
Kubernetes plans to change service account tokens to short-lived: https://github.com/kubernetes/kubernetes/issues/70679
This even simple in-cluster config will require on-the fly auth updates.

cben:
For BoundServiceAccountTokenVolume, reloading from file implemented in #532 (currently on every request! feedback wanted whether that needs optimizing).

PR #394, opened 2019-02-19
gcp automatic default credentials

timothysmith0609:
Follow-up of #392 

- Adds a check for gcp auth provider and defaults to using GoogleApplicationDefaultCredentials.
- Also added a specific error if `googleauth` gem isn't present on the importing application.
- Brought in Mocha as a dev dependency for mocking. I can use a bare mock if you prefer
- Add a unit test + test kubeconfig file

cc @cben

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,7 +296,7 @@ kubeclient can use `googleauth` gem to authorize.
 This requires the [`googleauth` gem](https://github.com/google/google-auth-library-ruby) that is _not_ included in
 `kubeclient` dependencies so you should add it to your bundle.
 
-If you use `Config.context(...).auth_options` and the kubeconfig file has `user: {auth-provider: {name: gcp}}`, kubeclient will automatically try this (raising LoadError if you don't have `googleauth`).
+If you use `Config.context(...).auth_options` and the kubeconfig file has `user: {auth-provider: {name: gcp}}`, kubeclient will automatically try this (raising LoadError if you don't have `googleauth` in your bundle).
 
 Or you can obtain a token manually:
 
--- a/README.md
+++ b/README.md
@@ -166,29 +166,6 @@ namespace = File.read('/var/run/secrets/kubernetes.io/serviceaccount/namespace')
 ```
 You can find information about tokens in [this guide](https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod) and in [this reference](http://kubernetes.io/docs/admin/authentication/).
 
-#### Google's Application Default Credentials
-
-On Google Compute Engine, Google App Engine, or Google Cloud Functions, as well as `gcloud`-configured systems
-with [application default credentials](https://developers.google.com/identity/protocols/application-default-credentials),
-you can use the token provider to authorize `kubeclient`.
-
-This requires the [`googleauth` gem](https://github.com/google/google-auth-library-ruby) that is _not_ included in
-`kubeclient` dependencies so you should add it to your bundle.
-
-```ruby
-require 'googleauth'
-
-auth_options = {
-  bearer_token: Kubeclient::GoogleApplicationDefaultCredentials.token
-}
-client = Kubeclient::Client.new(
-  'https://localhost:8443/api/', 'v1', auth_options: auth_options
-)
-```
-
-Note that this token is good for one hour. If your code runs for longer than that, you should plan to
-acquire a new one.
-
 ### Non-blocking IO
 
 You can also use kubeclient with non-blocking sockets such as Celluloid::IO, see [here](https://github.com/httprb/http/wiki/Parallel-requests-with-Celluloid%3A%3AIO)
@@ -308,6 +285,34 @@ Kubeclient::Client.new(
   auth_options: context.auth_options
 )
 ```
+
+
+#### Google's Application Default Credentials
+
+On Google Compute Engine, Google App Engine, or Google Cloud Functions, as well as `gcloud`-configured systems
+with [application default credentials](https://developers.google.com/identity/protocols/application-default-credentials),
+kubeclient can use `googleauth` gem to authorize.
+
+This requires the [`googleauth` gem](https://github.com/google/google-auth-library-ruby) that is _not_ included in
+`kubeclient` dependencies so you should add it to your bundle.
+
+If you use `Config.context(...).auth_options` and the kubeconfig file has `user: {auth-provider: {name: gcp}}`, kubeclient will automatically try this (raising LoadError if you don't have `googleauth`).
+
+Or you can obtain a token manually:
+
+```ruby
+require 'googleauth'
+
+auth_options = {
+  bearer_token: Kubeclient::GoogleApplicationDefaultCredentials.token
+}
+client = Kubeclient::Client.new(
+  'https://localhost:8443/api/', 'v1', auth_options: auth_options
+)
+```
+
+Note that this returns token good for one hour. If your code runs for longer than that, you should plan to
+acquire a new one.
 
 #### Security: Don't use config from untrusted sources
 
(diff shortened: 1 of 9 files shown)


tsontario:
> Metrics/PerceivedComplexity: Perceived complexity for fetch_user_auth_options is too high. [8/7]

This rubocop violation feels a bit too strict for me

lib/kubeclient/google_application_default_credentials.rb
```diff
@@ -5,7 +5,13 @@ module Kubeclient
   class GoogleApplicationDefaultCredentials
     class << self
       def token
-        require 'googleauth'
+        begin
+          require 'googleauth'
+        rescue LoadError
+          puts "Gem 'googleauth' not found. Kubeclient does not include the googleauth gem, "\
+          'you must include it in your own application'
```
cben: :+1: on the idea.  Not certain `puts` is ideal but don't know what's better.
I suggest adding a clarification to the that specifically "to support auth-provider `gcp`, you must include it ..."

cben:
Looks great! :clap:
No objection to mocha.
Yeah, rubocop is too draconian.  Feel free to bump PerceivedComplexity limit to what you consider sane in .rubocop.yml. :arrow_up: 
Can you add a note in CHANGELOG?

lib/kubeclient/google_application_default_credentials.rb
timothysmith0609: I suppose it's a specific enough thing that we can make our own error class for this

cben:
rebased to solve conflict (force-pushed your branch).

rubocop warns "Lint/InheritException: Inherit from RuntimeError instead of LoadError."
Introduced https://github.com/rubocop-hq/rubocop/pull/3427, motivation seems to be that default `rescue` copes with StandardError and won't catch this (and specific preference for RuntimeError is just that's a safer change https://github.com/rubocop-hq/rubocop/issues/3127#issuecomment-365211903).
Since this was LoadError before, nothing changes and `GoogleDependencyError < LoadError` is fine :+1:

cben:
Thanks LGTM :100:

I just pushed an update to the README, can you please review that?

cben:
cc @jeremywadsack, if you still care about gcp, please review (at least the README)

README.md
```diff
@@ -309,6 +286,34 @@ Kubeclient::Client.new(
 )
 ```
 
+
+#### Google's Application Default Credentials
+
+On Google Compute Engine, Google App Engine, or Google Cloud Functions, as well as `gcloud`-configured systems
+with [application default credentials](https://developers.google.com/identity/protocols/application-default-credentials),
```
cben: Can we claim all these 4 scenarios (GCE, AppEngine, Functions, gcloud) should work out of the box?
I wasn't sure so wrote "if ... kubeconfig file has `user: {auth-provider: {name: gcp}}`" to be safe...

README.md
```diff
@@ -309,6 +286,34 @@ Kubeclient::Client.new(
...
+auth_options = {
+  bearer_token: Kubeclient::GoogleApplicationDefaultCredentials.token
+}
+client = Kubeclient::Client.new(
+  'https://localhost:8443/api/', 'v1', auth_options: auth_options
+)
+```
+
+Note that this returns token good for one hour. If your code runs for longer than that, you should plan to
+acquire a new one.
```
cben: Q: how do you acquire a new one, in the automagic Config `gcp` scenario?
Calling `.context` again will do this, right?

README.md
jeremywadsack: All those systems are documented by Google to include application default credentials. See https://cloud.google.com/docs/authentication/production#obtaining_credentials_on_compute_engine_kubernetes_engine_app_engine_flexible_environment_and_cloud_functions and https://cloud.google.com/docs/authentication/production#obtaining_credentials_on_app_engine_standard_environment

jeremywadsack: FYI, I think the link above should be updated to this: https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application

README.md
jeremywadsack: Call `Kubeclient::GoogleApplicationDefaultCredentials.token` again.

I think we've discussed before that `kubeclient` doesn't have any support for renewing tokens. Is that something that's now covered with `exec` functionality?

README.md
```diff
@@ -309,6 +286,34 @@ Kubeclient::Client.new(
...
+
+auth_options = {
+  bearer_token: Kubeclient::GoogleApplicationDefaultCredentials.token
+}
+client = Kubeclient::Client.new(
+  'https://localhost:8443/api/', 'v1', auth_options: auth_options
+)
+```
+
+Note that this returns token good for one hour. If your code runs for longer than that, you should plan to
```
jeremywadsack:
```suggestion
Note that this returns a token good for one hour. If your code requires authorization for longer than that, you should plan to
```

jeremywadsack:
@timothysmith0609 @cben I'm really glad to see this change. I think that having the gem "just work" in many environments makes it easier for people to use. It helps people get started with the gem and then they can tune more if they need to.

Tangentially, I would suggest that we configure this to "just work" [inside a Kubernetes cluster](https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod). That is detect and implement the code in the [README section "Inside a Kubernetes Cluster"](https://github.com/abonas/kubeclient/blob/9d2ece748da6906f382102f9c46093e25fae5f2e/README.md#inside-a-kubernetes-cluster) when no other authorization is provided.

timothysmith0609:
>That is detect and implement the code in the [README section "Inside a Kubernetes Cluster"](https://github.com/abonas/kubeclient/blob/9d2ece748da6906f382102f9c46093e25fae5f2e/README.md#inside-a-kubernetes-cluster) when no other authorization is provided.

That sounds like a good idea. Looking at how client-go implements [rest.InClusterConfig](https://github.com/kubernetes/client-go/blob/master/rest/config.go#L399), looks like we can just lift that wholesale. In fact, that's pretty much what I've seen most apps doing (unsurprisingly). 

I propose to push that work out to a separate issue, however, and keep this change squarely scoped to just the gcp change. I think adding the in-cluster ability as a new method on `Kubeclient::Config` should be sufficient, and we'll leave the responsibility of calling the appropriate method to the caller (since sometimes clients are run in-cluster, but with a mounted `kubeconfig` to speak with other clusters). This has the nice benefit of mirroring the conventions of the client-go package as well

README.md
timothysmith0609: Calling `.context` will generate a new token, since it will end up calling `Kubeclient::GoogleApplicationDefaultCredentials.token`, as Jeremy points out. I think the farthest we should go in that direction is to capture the error, if possible, and let the caller handle it.

timothysmith0609:
@cben README changes seem good to me (made a minor edit for clarity re: LoadError). Are there any outstanding concerns? As mentioned above, @jeremywadsack 's suggestion to provide a simple mechanism for in-cluster config is a good idea (and the next logical step); let's push that out into the next issue.

jeremywadsack:
@timothysmith0609 Agree that in-cluster should be a separate PR. For reference see also our gem that implements this (in ruby): https://github.com/keylimetoolbox/resque-kubernetes/blob/v2.0.0/lib/resque/kubernetes/context/well_known.rb

And yes, we should do this as a last resort. If there's a provided configuration use that; if not, check for GCP; final step is to check for in-cluster.

cben:
Hi, was sick for a few days, thanks for the polish :smile: 
Added a test that `.context()` does call `.token` (and `.auth_options` doesn't) and documented this.
Going to merge & release.

cben:
Opened #395 for in-cluster config.
Opened #393 for authentication renewal.  I'll brain-dump my thoughts there later, but any ideas welcome.

cben:
released in gem 4.3.0.

PR #396, opened 2019-02-25
Implement OIDC Auth Provider

rhysm:
Adds ability for Kubeclient to utilize the oidc auth-provider both by
re-using the id-token that is inside the kubeconfig if it is not expired
and refreshing it if the token has passed expiry.


cben:
Sorry for delay, reviewing...
This is cool!
k9s docs: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens

lib/kubeclient/oidc_auth_provider.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Kubeclient
+  # Uses OIDC id-tokens and refreshes them if they are stale.
+  class OIDCAuthProvider
+    class OpenIDConnectDependencyError < LoadError # rubocop:disable Lint/InheritException
+    end
+
+    class << self
+      def token(auth_provider)
```
cben: [optional nit, your call] The name `auth_provider` sounds like this would contain the ['auth-provider'] part of the config, but it contains the ['auth-provider']['config'] sub-part.  Maybe `config`? `auth_provider_config` (ouch)?  `provider_config`?

lib/kubeclient/oidc_auth_provider.rb
```diff
@@ -0,0 +1,45 @@
...
+          raise OpenIDConnectDependencyError,
+                'Error requiring openid_connect gem. Kubeclient itself does not include the ' \
+                'openid_connect gem. To support auth-provider oidc, you must include it in your ' \
+                "calling application. Failed with: #{e.message}"
+        end
+
+        issuer_url = auth_provider['idp-issuer-url']
+        discovery = OpenIDConnect::Discovery::Provider::Config.discover! issuer_url
+
+        id_token = OpenIDConnect::ResponseObject::IdToken.decode auth_provider['id-token'],
```
cben: What if the config doesn't contain `id-token` at all?  (Is that possible?)
As we can always obtain a new one, let's allow this?

README.md
```diff
@@ -314,6 +314,23 @@ client = Kubeclient::Client.new(
...
+id-tokens when they have expired.
+
+This requires the [`openid_connect` gem](https://github.com/nov/openid_connect) which is not included in
+the `kubeclient` dependencies so should be added to your own applications bundle.
+
+The OIDC Auth Provider will not perform the initial setup of your `$KUBECONFIG` file. You will need to use something
+like [`dexter`](https://github.com/gini/dexter) in order to configure the auth-provider in your `$KUBECONFIG` file.
+
+If the id-token specified in your `$KUBECONFIG` file has not expired the OIDC Auth Provider will simply use that token.
+If it has expired then the provider will refresh the token and use that.
```
cben: The wordings "will refresh" ... "will not automatically refresh" sounds confusing on a quick read.
I propose here saying:
> If you use `Config.context(...).auth_options` and the kubeconfig file has `user: {auth-provider: {name: gcp}}`, kubeclient will automatically obtain a token (or use `id-token` if still valid)

to me "obtain" doesn't imply ability to keep refreshing it, nor sounds as action done *to* the token so less expectation it'll write it back to the config file (the disclaimer below is still good).

The other change is becase this section talks abstractly about "the OIDC Auth Provider" but I wanted to say specifically how to invoke this (`Client` doesn't do this, `Config` does)

cben: (one day if we implement #393, we can reword these sections in terms of "refresh" or "renewal")

README.md
```diff
@@ -314,6 +314,23 @@ client = Kubeclient::Client.new(
...
+
+Currently tokens are only valid for one hour and the provider will not automatically refresh them. Fresh id-tokens are also
```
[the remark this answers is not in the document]
rhysm: Updated for clarity.

README.md
```diff
@@ -314,6 +314,23 @@ client = Kubeclient::Client.new(
 Note that this returns a token good for one hour. If your code requires authorization for longer than that, you should plan to
 acquire a new one, by calling `.context()` or `GoogleApplicationDefaultCredentials.token` again.
 
+#### OIDC Auth Provider
+
+If the cluster you are using has OIDC authentication enabed you can use the `openid_connect` gem to refresh
+id-tokens when they have expired.
```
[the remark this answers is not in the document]
rhysm: Updated for clarity.

lib/kubeclient/oidc_auth_provider.rb
```diff
@@ -0,0 +1,45 @@
...
+          token_endpoint: discovery.token_endpoint,
+          userinfo_endpoint: discovery.userinfo_endpoint
+        )
+        client.refresh_token = auth_provider['refresh-token']
+        client.access_token!.id_token
+      end
+
+      def expired?(id_token)
+        # If token expired or expiring within 60 seconds
+        Time.now.to_i + 60 > id_token.exp.to_i
```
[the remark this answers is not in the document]
rhysm: The openid_connect gem performs the same check except for the 60 second grace so I believe it will be fine.

lib/kubeclient/oidc_auth_provider.rb
rhysm: Good point, I've gone with `provider_config`.

lib/kubeclient/oidc_auth_provider.rb
rhysm: I'm not sure how you'd get into that state but `kubectl` supports it so I've implemented it too.

README.md
rhysm: Updated README for more clarity, let me know what you think.

cben:
BTW, a security question:
Kubeclient supports opt-out from some "dangerous" config handling — reading local files, and `exec:`.
(For scenarios where you take config from untrusted source, which I warn against anyway, and I haven't seen anyone doing, but on general principles...)
Do you think this opt-out should also disable OIDC (and GCP #394) auth-providers?

On one hand, making a network request is not risky to your machine(?).  
On the other, that's still "acting on data" vs passively reading it.  However the goal of Config is to feed it to Client which will act on a lot of it by network requests anyway so that's hardly a criterion...

cben:
released in gem 4.3.0.

PR #402, opened 2019-03-03
Add some clarrification on Apps api group

blijblijblij:


cben:
Thanks!  However listing all api groups here will obviously not scale...  Also, I suspect you didn't you come to the docs already knowing that what you want is "A Group Api with app object"...

I'd like to add some brief refresher what are API groups and how to figure out what group one needs (using `kubectl`, and from k8s api docs)...
Keeping this open as reminder but unlikely to merge this as-is.

PR #404, opened 2019-03-28
adds authentication helper for aws eks

smcavallo:
Adds a helper class which can retrieve a token which can be used to authenticate against aws eks

Signed-off-by: S.Cavallo <smcavallo@hotmail.com>


cben:
Cool, reviewing...

lib/kubeclient/aws_eks_credentials.rb
```diff
@@ -0,0 +1,47 @@
...
+          require 'aws-sigv4'
+          require 'base64'
+          require 'cgi'
+        rescue LoadError => e
+          raise AmazonEksDependencyError,
+                'Error requiring aws gems. Kubeclient itself does not include the following ' \
+                'gems: [aws-sdk-core,aws-sigv4]. To support auth-provider eks, you must ' \
+                "include it in your calling application. Failed with: #{e.message}"
+        end
+        # https://github.com/aws/aws-sdk-ruby/pull/1848
```
cben: Does this mean one needs a sufficiently new  aws-sdk-core?  Is ainimal version worth documenting?

README.md
```diff
@@ -287,6 +287,37 @@ Kubeclient::Client.new(
...
+
+auth_options = {
+  bearer_token: Kubeclient::AmazonEksCredentials.token(credentials, eks_cluster)
+}
+client = Kubeclient::Client.new(
+  'https://localhost:8443/api/', 'v1', auth_options: auth_options
+)
+```
+
+Note that this returns a token good for one hour. If your code requires authorization for longer than that, you should plan to
```
cben: Is it indeed 1 hour? (or just copy pasted)

README.md
```diff
@@ -287,6 +287,37 @@ Kubeclient::Client.new(
...
+On Amazon EKS, when using aws-iam-authentication, with [credentials](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Credentials.html),
+kubeclient can use `aws-sdk-core` gem to retrieve a token to authenticate against EKS.
+
+This requires a set of gems which are _not_ included in
+`kubeclient` dependencies (`aws-sdk-core` and `aws-sigv4`) so you should add them to your bundle.
+
+To obtain a token:
+
+```ruby
+require 'aws-sdk-core'
```
cben: Can this be omitted?  Its required in token().

README.md
```diff
@@ -287,6 +287,37 @@ Kubeclient::Client.new(
...
+# Use keys
+credentials = Aws::Credentials.new(access_key, secret_key)
+# Or a profile
+credentials = Aws::SharedCredentials.new(profile_name: 'default').credentials
+
+auth_options = {
+  bearer_token: Kubeclient::AmazonEksCredentials.token(credentials, eks_cluster)
+}
+client = Kubeclient::Client.new(
+  'https://localhost:8443/api/', 'v1', auth_options: auth_options
```
cben: nit: is localhost realistic when running on EKS?  What's a typical hostname (or use example.com) or port?

README.md
```diff
@@ -287,6 +287,37 @@ Kubeclient::Client.new(
 ```
 
 
+#### Amazon EKS Credentials
+
+On Amazon EKS, when using aws-iam-authentication, with [credentials](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Credentials.html),
```
cben: Can you clarify whether it's for cluster running on EKS, or the code calling Kubeclient, or both?

lib/kubeclient/aws_eks_credentials.rb
```diff
@@ -0,0 +1,47 @@
...
+          http_method: 'GET',
+          url: 'https://sts.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15',
+          body: '',
+          credentials: credentials,
+          expires_in: 60,
+          headers: {
+            'X-K8s-Aws-Id' => eks_cluster
+          }
+        )
+        kube_token = 'k8s-aws-v1.' + Base64.urlsafe_encode64(presignedUrlString.to_s).chomp('==')
```
cben: Do you have reference for removing the `==` padding?  I think it's there in base64 because without it sometimes the decoding is ambiguous (?), so I worry if this works now but might break if length changes...

lib/kubeclient/aws_eks_credentials.rb
smcavallo: Now that you mention this, the *only* requirement is aws-sigv4.  aws-sdk is not required.  consumers can use any version of aws-sdk v2 or aws-sdk-core v3 as long as it contains a credentials object it can pass through to the aws-sig4 pre-signer class.  i'll update the documentation

README.md
smcavallo: one minute - sorry - copy/paste

README.md
smcavallo: it's the https endpoint for your eks cluster (same as you'd put in your kubeconfig.  i've updated the docs.

README.md
smcavallo: i've removed it from token but will be required for anyone constructing the Aws::Credentials object

lib/kubeclient/aws_eks_credentials.rb
smcavallo: when calling Base64.urlsafe_encode64 on the presignedurl, the result ends in `==` 
On the aws side, the request will not be authorized unless the `==` doesn't exist at the end.  I don't know of a better solution - this is what we need to do when generating our tokens for them to work correctly.

README.md
smcavallo: i'll put a blurb in there.  it's possible to use aws-iam-authentication with kubernetes clusters which are self-hosted on EC2.  I've never tested these tokens in that configuration.

cben:
Thanks again.

1 minute is really short, for the interested #393 is brainstorming on adding auto renewal to kubeclient.

PR #406, opened 2019-04-08
automatically load aws credential library

smcavallo:
automatically load aws credential library
Signed-off-by: S.Cavallo <smcavallo@hotmail.com>

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -1,6 +1,7 @@
 require 'json'
 require 'rest-client'
 
+require 'kubeclient/aws_eks_credentials'
 require 'kubeclient/common'
 require 'kubeclient/config'
 require 'kubeclient/entity_list'


lib/kubeclient.rb
```diff
@@ -1,6 +1,7 @@
 require 'json'
 require 'rest-client'
 
+require 'kubeclient/aws_eks_credentials'
```
cben: This will require adding it as a dependency for everyone, no?

cben: Aah wait, it's our helper not external library.  I need to look at this when I'm less tired :-)

cben: LGTM :+1:
Our README says one can use `Kubeclient::AmazonEksCredentials` doesn't say one would need to require it (which would be strange, `require 'kubeclient` is enough for everything else).
aws_eks_credentials.rb only requires `aws-sigv4` when actually used.

PR #407, opened 2019-04-15
Consider OIDC tokens with an unknown kid to be expired

benlangfeld:
When the authorization server rotates key IDs, we might encounter a token with a kid which the server no longer reports in discovery. Currently, json-jwt raises an exception and the user must expend a lot of effort debugging. We can help by simply assuming the token is expired and forcing token refresh.


cben:
Sorry, very busy week :-/  Hope to review soon...

benlangfeld:
@cben Done :)

PR #408, opened 2019-04-16
Persist refreshed tokens

benlangfeld:


cben:
Friendly ping :wave:, as #407 has been merged, would you like to rebase and advance this?

AFAICT, this prepares some APIs but doesn't yet write back a kubeconfig file, right?

cben:
BTW, preparing some APIs can be a great stand-alone PR :+1:, not saying you have to do it all at once.

cben:
Do'h I finally get it.  Currently, (once `id-token` from Config is expired) each `OIDCAuthProvider.token` call creates a fresh `OpenIDConnect::Client.new(...)` so the `refresh_token` it returns gets discarded entirely. :frowning_face: 

- [ ] This PR doesn't yet persist to a file (and that'd deserve separate discussion whether it needs to be opt-in) but:
- [x] it does bundle refresh_token with the returned access token :+1:
- [x] it does bundle it into `auth_options[:refresh_token]` that goes into `Config::Context`  object (and typically from there into `Client` object(s)). :+1:
- [ ] I don't see it being used for next renewal yet?

As I now mentioned on #606, need to find right place where to save it... current README advice to renew auth by re-creating all objects, even the `Config`, is bad.

PR #413, opened 2019-05-28
update http client version

joshlarsen:
I see there is some discussion about changing the http client. In the mean time, perhaps we can bump the current client to a more recent version.

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -34,5 +34,5 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'rest-client', '~> 2.0'
   spec.add_dependency 'recursive-open-struct', '~> 1.0', '>= 1.0.4'
-  spec.add_dependency 'http', '~> 3.0'
+  spec.add_dependency 'http', '~> 4.0'
 end
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -34,5 +34,5 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'rest-client', '~> 2.0'
   spec.add_dependency 'recursive-open-struct', '~> 1.0', '>= 1.0.4'
-  spec.add_dependency 'http', '~> 4.0'
+  spec.add_dependency 'http', '>= 3.0', '< 5.0'
 end


cben:
Let's see, reading https://github.com/httprb/http/blob/master/CHANGES.md...

Have you tried running with 4.0?

cben:
- [ ] https://github.com/httprb/http/pull/467 Drop Ruby 2.2 support
  kubeclient still supports 2.2 and later.  Can we support both http versions with `http >= 3.0`?
  Ruby 2.2 and 2.3 [are EOL](https://www.ruby-lang.org/en/downloads/branches/), so I'm open to dropping 2.2 or both, but that can only happen in a major release.

joshlarsen:
@cben I am running 4.1 and no issues so far. I agree `http >= 3.0` seems like a good compromise to support both.

cben:
- [x] https://github.com/httprb/http/pull/446
  > The previous style: .timeout(:global|:per_operation, hash_of_options) doesn't work anymore.

  Not a problem, kubeclient only uses `http` gem for watch and deliberately doesn't set any timeouts.

Other 4.0.0 changes don't sound relevant.

cben:
- [x] https://github.com/httprb/http/pull/436
  Raise `ConnectionError` when writing to socket fails (insdead of `Errno::EPIPE` or similar).
  
  At one point, kubeclient was catching `Errno::EBADF` but the scenario was different, not relevant.
  Also #315 changed it to catch `StandardError` which would cover `HTTP::ConnectionError`.

kubeclient.gemspec
```diff
@@ -34,5 +34,5 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'rest-client', '~> 2.0'
   spec.add_dependency 'recursive-open-struct', '~> 1.0', '>= 1.0.4'
-  spec.add_dependency 'http', '~> 3.0'
+  spec.add_dependency 'http', '~> 4.0'
```
cben: Can you change this to:
```suggestion
  spec.add_dependency 'http', '>= 3.0', '< 5.0'
```

Hopefully that'd automagically make ruby2.2 to use http 3 (http 4 declares `required_ruby_version = ">= 2.3"`).
And leave the option to use 3 for anyone that has other constraints.

cben:
Weird, the 2.2.0 job https://travis-ci.org/abonas/kubeclient/jobs/545301561#L488 
said `Installing http 4.0.0`

cben:
Aha, 4.0.0 metadata still declared 2.2 support.
Later rubies get `Fetching http 4.1.1`

cben:
Asked on https://github.com/httprb/http/commit/190d4ebb97bbc5eb202f20439971f5608ab76c51

PR #414, opened 2019-05-29
Travis: also test on Mac, only run rubocop once

cben:
Running rubocop on all rubies is pointless, as it's designed to separate "ruby it runs on" from "rubie(s) the linted source is target to".  All runs produce exactly same warnings, if not it's a rubocop bug.


cben:
Mac ruby 2.2 failing with:
```
Ignoring executable-hooks-1.4.2 because its extensions are not built.  Try: gem pristine executable-hooks --version 1.4.2
Ignoring gem-wrappers-1.3.2 because its extensions are not built.  Try: gem pristine gem-wrappers --version 1.3.2
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/rubygems.rb:241:in `bin_path': can't find gem bundler (>= 0.a) (Gem::GemNotFoundException)
	from /usr/local/bin/bundle:22:in `<main>'
```

Note `ruby/2.3.0/` in there, that's weird, as we're trying to run ruby 2.2!

Whatever, I'll just skip 2.2 on mac.
And while I'm there, I'll limit rubocop to only 1 run.

cben:
hmm, I wonder if I can kubeclient works on other interpreters like jruby, rubinius etc

PR #420, opened 2019-09-24
Added resourceVersion to acceptable get options

jkohen:
ResourceVersion is useful to continue a previous get, and making it 0 forces a local api server request instead of an etcd fetch, which is more efficient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,6 +447,12 @@ You can specify multiple labels (that option will return entities which have bot
 pods = client.get_pods(label_selector: 'name=redis-master,app=redis')
 ```
 
+You can get entities at a specific version by specifying a parameter named `resource_version` (named `resourceVersion` in Kubernetes server):
+
+```ruby
+pods = client.get_pods(resource_version: '0')
+```
+
 Get all entities of a specific type in chunks:
 
 ```ruby
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -39,7 +39,7 @@ module Kubeclient
     SEARCH_ARGUMENTS = {
       'labelSelector'   => :label_selector,
       'fieldSelector'   => :field_selector,
-      'resourceVersion' => :resource_version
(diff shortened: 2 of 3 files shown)


jkohen:
I've tested this with a client, and now I can use resource_version in the options for `get_pods`.

richm:
Can you point me at a link to the documentation where it describes the behavior of the query with `resourceVersion: 0` with respect to the etcd/cache lookup?
Do you have any idea of the performance gain of this approach?

jkohen:
@richm https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#list-pod-v1-core
Explains the difference between 0 and unset. Kubelet uses 0. In an internal benchmark by the Google Kubernetes team we are seeing 40ms vs 4s.

cben:
Thanks!  I'm a bit surprised we didn't support this already :-)

Let's check naming consistency:

- `client.watch_foos(resource_version: 123)`
- `client.get_foos(resource_version: 123)` added here
- `EntityList` objects (returned in default `as: :ros` mode) have `.resourceVersion` attribute.  
    In other modes, you get data as-is, i.e. `[:metadata][:resourceVersion]` or `["metadata"]["resourceVersion"]`.

This is a bit confusing but not much we can do about it.
(Not worth breaking compatibility, but could add aliases somewhere.  We can't change the as-is modes, and I think EntityList deviating from them would only be more confusing.  I'm not sure  changing the methods to accept `resourceVersion: 123` is beneficial as all other kubeclient APIs do use lower_case style?)
Anyway, out of scope, here you're mirroring existing watch interface :+1:

Can you document this in README?
Please add a test.  (see similar `test_entities_with_label_selector`)

jkohen:
Done & Done. Thanks for the quick review, Beni!

cben:
Will hopefully make a release tomorrow.

cben:
released in 4.5.0 (#421)

Issue #332, opened 2018-05-27
Expose {create,delete,...}_entity methods to manipulate objects of dynamic kind


cben:
If you have objects whose `kind` you don't know ahead of time, you can't used discovery-created methods like `create_service`, `get_services`.
Computing the method name e.g. `client.send("get_#{kind}", ...)` is messy, especially if you need the plural name :-(

We already have generic methods: `get_entity`, `get_entities`, `create_entity`, `update_entity`, `patch_entity`, `delete_entity`, `watch_entities` — except they're not documented in README and **not considered public** (and I believe we already changed interface once based on this)

Main use case is using objects from a yaml file (or from template processing). Related #208, #187, #325, #329.  (I felt before those would be good to expose, now found convincing use case)

- [ ] review the API is good
  - [ ] `watch_entities` does double singular/plural duty depending of if `:name` is given — is this elegant?
    So does `watch_pods` BTW — there is no `watch_pod`.
  - [ ] is namespace passed consistently? #312
- [ ] document as public

stiller-leser:
Some initial feedback from my side. I just tried `create_entity` and noticed a couple of things:

- Isn't it redundant that I need to specify `kind` and `name`, since they are present in `entity`? Maybe we could overload `create_entity` here with two diffrent kinds of parameters here? For example `create_entity(entity_type, resource_name, entity_config)` and `create_entity(entity)`. The latter one could then be used with `YAML.load_stream(file).first` (basically just dumping the yaml file content in a loop into `create_entity` as a first step. Later an additional method like `create_entities_from_yaml` could be provided.
- The method should check if the keys in the hash are strings or symbols, or at least raise a warning. Currently string-keys lead to `undefined method '[]' for nil:NilClass` because of https://github.com/abonas/kubeclient/blob/c637dc1d44363ea3facf2c893225a844f806dbb3/lib/kubeclient/common.rb#L311
- In the same line I would additional fallback to the default K8S namespace `default`

Happy to provide more feedback (or even PR, if I can free up some time and get the connection to my cluster working.)

Thanks for your help in #329!

cben:
Feedback from @stiller-leser, not sure why it disappeared:

> Some initial feedback from my side. I just tried `create_entity` and noticed a couple of things:
>
[...]

PR #422, opened 2019-10-18
Add ability to patch status of an object

a-chernykh:


cben:
Nice. 2 small concerns:

1. Do we also need JSON merge patch and JSON patch variants?  Might be worth it for orthogonality alone; also, CRDs don't support strategic merge.

2. Are there any kind pairs "Foo" and "FooStatus"?  Then method names would collide.  I agree it's unlikely.  Could you grep https://github.com/cben/kubernetes-discovery-samples (I'm away from computer today)?

cben:
BTW, are these really adding functionality vs doing patch on the whole object only touching `status` part?  I think (untested) that works too.

I suspect the API has `foo/status` endpoints mostly for ease of whole-sale replacement (PUT).
Or maybe for separate RBAC?  (E.g. an operator having read access to resources themselves and write to their status).  In that use case, I agree this adds functionality.

cben:
An alternative could be to not provide these [for now] as generated method names, first as generic method taking the kind of object as parameter. 

That is, implement at least patch part of generic methods #332, plus status variants.  (Or maybe same method abusing `kind: Foo/status` but probably not, I prefer #332 layer to match k8s semantics closely)

Don't take this as hard pushback, I'm looking for your opinion, but I also know we want a #332 layer anyway so I'm interested in nudging people in that direction 😉 

BTW, kubeclient is generally lacking about `foo/bar` sub-resources.  We have a few specialized methods here and there...
Do you think there is use case for more methods like `update_foo_status`, `get_foo_status`, `watch_foo_status`...?

If you have any broader vision how to support sub-resources, please write 👍

andreychernih:
Hi, @cben, I appreciate fast turnaround reviewing this PR and thank you for maintaining such a critical open-source project.

To be honest, this was a quick and dirty change I needed to make in order to benefit from custom Pod conditions (trying to fill in some gaps in Kubernetes inability to run slow jobs without interruptions during auto-scaling and deployments). 

Unfortunately, I was not able to use the regular `PATCH /pods/<pod_name>` endpoint in order to update custom status, therefore, I had to implement this `_patch` method which uses `PATCH /pods/<pod_name/status`. This endpoint has a separate RBAC as well. I tested this change in our AWS EKS cluster and it works as expected.

cben:
Oh i now remembered there is a starightforward way to add *full* set of operations on these `foo/status` sub-resources:
I believe we're already getting them listed in discovery, and we're deliberately ignoring ones that contain a slash.  We could stop ignoring those and decide how to map them to method names, e.g. replace slash with underscores.

cben:
Sorry, I effectively blocked this by saying "there is a better way" but never implementing it :-|
I won't have time to work on this until 5.0 release (#435).
Why don't I just merge your PR for now and extend later?  I probably could.  But I wanted to verify the more general approach would result in interface as this "special case" approach...

Are in interested in tackling the general approach yourself?
- modify `next if resource['name'].include?('/')` line in `load_entities` to allow things ending in `/status`.
- modify `parse_definition` to translate that to `_status` suffix
- tests...

Alternatively, please ping me periodically :wink:

shuhaowu:
This is required to patch status right? Trying to update a single status field by re-POSTING the entire resource doesn't seem to actually change the field in the status.

cben:
> The PUT and POST verbs on objects MUST ignore the "status" values, to avoid accidentally overwriting the status in read-modify-write scenarios. A /status subresource MUST be provided to enable system components to update statuses of resources they manage.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

PR #423, opened 2019-11-27
Remove padding from Base64 token in AmazonEksCredentials

mastamark:
The Base64 encoding of the signature request sometimes comes back with 1 or 2 equal signs (or 'padding' - https://parsiya.net/blog/2017-08-06-tldr-base64/#padding).  This change should be safe as the '=' is only for padding and is safe to remove.

### DIFF
--- a/lib/kubeclient/aws_eks_credentials.rb
+++ b/lib/kubeclient/aws_eks_credentials.rb
@@ -38,7 +38,7 @@ module Kubeclient
             'X-K8s-Aws-Id' => eks_cluster
           }
         )
-        kube_token = 'k8s-aws-v1.' + Base64.urlsafe_encode64(presigned_url_string.to_s).sub(/=(.*)?/, '') # rubocop:disable Metrics/LineLength
+        kube_token = 'k8s-aws-v1.' + Base64.urlsafe_encode64(presigned_url_string.to_s).sub(/=*$/, '') # rubocop:disable Metrics/LineLength
         kube_token
       end
     end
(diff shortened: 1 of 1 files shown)


lib/kubeclient/aws_eks_credentials.rb
```diff
@@ -38,7 +38,7 @@ def token(credentials, eks_cluster)
             'X-K8s-Aws-Id' => eks_cluster
           }
         )
-        kube_token = 'k8s-aws-v1.' + Base64.urlsafe_encode64(presigned_url_string.to_s).chomp('==')
+        kube_token = 'k8s-aws-v1.' + Base64.urlsafe_encode64(presigned_url_string.to_s).sub(/=(.*)?/, '') # rubocop:disable Metrics/LineLength
```
cben: Nice catch, thanks :+1:

If, for whatever reason, there is padding in the middle of the encoded data, this regexp will discard it together with everything that follows.  I don't think I ever saw that, but as [Wikipedia says](https://en.wikipedia.org/wiki/Base64#Output_padding):

> The padding character is technically not needed for decoding  ...  One exception, where padding characters are technically required, is when multiple Base64 encoded files have been concatenated. 

This really shouldn't happen here, but do you mind changing this to the obviously safe `.sub(/=*$/, '')`?

P.S. ruby 2.3+ has `Base64.urlsafe_encode64(..., padding: false)` but not 2.2 which we still support so can't use it.  That's implemented simply by `str.delete("=")`, so I guess I'm being over-pedantic.

mastamark: Done!

cben:
Sorry for delay releasing.  I intend to release 4.6.0 with a couple new features soon.  Let me know if anybody also wants a 4.5.1 bugfix release.

PR #425, opened 2019-12-13
make using watcher easier/safer by passing a block directly

grosser:
I just want to do this ... having to know about watcher is unnecessary overhead and it not automatically closing itself is trouble some too, so make it simple to do the right thing:

```ruby
watch_nodes do |node|
```

/cc @cben

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,8 +596,7 @@ client.all_entities
 It is possible to receive live update notices watching the relevant entities:
 
 ```ruby
-watcher = client.watch_pods
-watcher.each do |notice|
+client.watch_pods do |notice|
   # process notice data
 end
 ```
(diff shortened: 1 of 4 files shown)


cben:
Will review next week but sounds great.

grosser:
I updated the docs/tests

cben:
Code looks excellent :100:  

The guaranteed closing is a win :+1:  
IIUC now if you run the loop till the server stops sending, the http connection is cleaned up, but if you `break` or crash in the middle you can leak open connections, right?
- [ ] Note that in "interrupt from another thread" scenario, this is still a risk.  Should we document "be careful to always call `.finish`?
    Could we harden `WatchStream#each` to always close?  Is that backward-compatible?  It maybe currently possibile to call `.each`, break/crash in the middle, then call `.each` to consume more...  But that's not reliable as `buffer` is a local variable, is lost when you break.  I'll open an issue.  Out of scope here.

Are there any considerations wrt. watch resumption (#273)?

- One thing we learnt there is currently the starting resourceVersion is baked into a WatchStream instance, so you need a new one every time anyway.

- An intermediate low-level improvement could be for kubeclient to just track the last-seen resourceVersion, with user still having to write a loop calling `WatchStream`.
  The best place to expose I can think of is on the WatchStream object, but with this new interface you won't have access to it, so the *correct* way to watch  However, just tracking the version is not a big improvement, I don't really want to stop there...

- Hopefully, kubeclient can implement the full watch resumption (as long as known version is still valid, 410 will still exit to caller).
  I don't see a strong reason to make it opt-in, I think we can enable it for all existing callers (probably providing opt-out for low-level use?).  Either way that should be orthogonal to the shortcut here — it should be applicable to both new `client.watch_foo { ... }` style and existing code doing `watcher = client.watch_foo; watcher.each { ... }`.  The latter will require changing WatchStream to support multi-use (or keep it single use but have watch_foo return a multi-use wrapper over it?)

PR #426, opened 2019-12-17
Support limitBytes query parameter for get_pod_log

dwilkins:
some log lines could be really long and produce more data than a
ruby application is willing to process. Adding `limit_bytes` as a
parameter to `get_pod_log` will help protect against large sized
logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ matrix:
   exclude:
     - os: osx
       rvm: "2.2.0"
+    - os: osx
+      rvm: "2.3.0"
   include:
     - os: linux
       rvm: "2.5.0"
--- a/README.md
+++ b/README.md
@@ -673,6 +673,12 @@ client.get_pod_log('pod-name', 'default', tail_lines: 10)
 # => "..."
 ```
 
+Kubernetes can fetch a specific number of bytes from the log, but the exact size is not guaranteed and last line may not be terminated:
+```ruby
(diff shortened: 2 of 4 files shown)


cben:
LGTM on the feature :+1:, thanks!

Goes back far, apparently since kubernetes 1.1.

---

OSX failure: https://travis-ci.org/abonas/kubeclient/jobs/626345033
```
Libraries missing for ruby-2.3.0: /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib,/usr/local/opt/openssl/lib/libssl.1.0.0.dylib. Refer to your system manual for installing libraries
Mounting remote ruby failed with status 10, stopping installation.
Gemset '' does not exist, 'rvm ruby-2.3.0 do rvm gemset create ' first, or append '--create'.
The command "rvm use 2.3.0 --install --binary --fuzzy" failed and exited with 2 during .
```
Googling these errors didn't give any useful leads :frowning_face: 
Let me check if is a fluke or consistent...

You're right that ruby 2.2 and 2.3 are [end of life](https://www.ruby-lang.org/en/downloads/branches/). I'll drop them entirely in next major release; for now I'm keeping gemspec declaration that we're `'>= 2.2.0'` but ok with excluding on travis if it's recurring as fighting it is not productive.

cben:
Same errors in https://www.travis-ci.org/cben/kubeclient/jobs/627225675 from master.
OK, merging.

Issue #432, opened 2020-02-11
Kubeclient fails on Ruby 2.7.0


benlangfeld:
Need a fix for this: https://github.com/aetherknight/recursive-open-struct/issues/62

cben:
Unrelated warning seen on 2.7:
```
lib/kubeclient/config.rb:33: Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
test/test_config.rb:128: Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
test/test_config.rb:137: Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
```

cben:
Tests fixed by recursive-open-struct 1.1.1 🙏, which current kubeclient gemspec allows.  So nothing to do on our side :tada:
Added 2.7 to kubeclient's test matrix #433.

cben:
Update: #439 bumped recursive-open-struct to 1.1.1+.
Will eventually release in 5.0 (blockers: #435).
Let me know if anyone wants a 4.6.1 (4.7.0?) backport.

PR #433, opened 2020-02-12
Test on ruby 2.7.0

cben:
First step to understanding #432...


cben:
`recursive-open-struct` 1.1.1 with a fix was just released, let's see if it helps...
https://github.com/aetherknight/recursive-open-struct/issues/62#issuecomment-597468960

PR #436, opened 2020-03-11
Document watch and upcoming changes

cben:
Turns out watch was barely documented, while it very much needs documentation :grin:
Preparation for 5.0 changes proposed in https://github.com/abonas/kubeclient/pull/275#issuecomment-591928097

Also, the whole "Examples" section (which is really the main documentation) is messy, took a small step towards organizing it...

@grosser @qingling128 @mszabo @Stono please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ client = Kubeclient::Client.new(
 
 ### Timeouts
 
-Watching never times out.
+Watching configures the socket to never time out (however, sooner or later all watches terminate).
 
 One-off actions like `.get_*`, `.delete_*` have a configurable timeout:
 ```ruby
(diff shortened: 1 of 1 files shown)


cben:
- In several places here, I neglected to say which exception / how exactly to detect 410 errors.  I don't really want people to now write code catching `HTTP` errors and to check `notice.code`, I hope to fix them asap (and I intend to have a separate sub-exception for 410 because of its importance).

- I didn't mention the new ["BOOKMARK" type](https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks).  
  Kubeclient doesn't know how to opt-in yet.  Once we track last seen version, we could opt-in and honor these _transparently_ without yielding them to the block.  Will decide how to cross that bridge when we get there.
  I wondered if I should add generic advice to ignore unknown `type` for future compatibility?  But if people did that now, we'd all be ignoring `"type": "ERROR"` :laughing:  And it seems k8s will make any new types opt-in.

cben:
BTW, the terminology "watch notice" comes from `Kubeclient::WatchNotice`, which we dropped in #299.  The k8s golang & opeapi type name is `WatchEvent`, so maybe should start using that?
 https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#watchevent-v1-meta

Issue #435, opened 2020-03-10
5.0 imcompatible changes plan


cben:
- [x] changed `under_score` algorithm https://github.com/abonas/kubeclient/pull/355
- [ ] change watch exceptions behavior https://github.com/abonas/kubeclient/pull/275#issuecomment-591928097
- [ ] drop EOL ruby versions
- [ ] I'm thinking about making all methods not documented in README `private`?
  Specifically `get_entities`, `patch_entity` etc?

cben:
- What about Remove deprecated `KubeException` #293?
  => Not now. As explained there, `HttpError` is not sufficient, we need some top-level exception.  Should probably introduce a `Kubeclient::Error` but that's a compatible change, and should give users time to adopt it before dropping `KubeException`.

cben:
Apologies for delays.  Working from home with kids has been... challenging.
I hope to get a lot done next week.

grosser:
I'd propose: 
- get rid of most of the api:
User has to set the api version and resource anyway, so much more efficient and simple to call `client.get("nodes")` without doing the extra api discovery 🤷 
- remove struct and return hashes ... it's an expensive crutch :D

cben:
> get rid of most of the api

That will be an entirely different gem :smile:
There is a non-breaking road though: first implement #332, then much later see how well people adopted new style and how people feel about dropping existing API.

> remove struct and return hashes

Now that it's simple matter of ``as: :parsed` / `as: :parsed_symbolized`, I don't see a large enough payoff to justify breaking compatibility.
We could do small nudges like make `as:` required in `Client` contructor to stop encouraging a bad choice.  Or even make people require and inject recursive-open-struct and stop depending on it?
But _forcing_ people with lots of working `.foo.bar` code to rewrite everything to `[:foo][:bar]` form would be a disservice IMHO.

If you care about this, start by changing README to prominently explain why the new modes are recommended.

In short, both of these points sound more "8.0" (and maybe never) then "5.0" to me :)

cben:
- [ ] `.rest_client` is definitely going private (or away) to unblock #466.

grosser:
#332 sounds like a good step forward
... requiring `as` also sounds good (with or without removing struct
runtime dependency)

grosser:
stumbled into the whole url parsing mess again 😞 
so add to the list:
- make version required
- make version `apps/v1` instead of mashing up host/apis/apps and v1
- dynamically set `api` or `apis` depending on version == v1
[...]

PR #439, opened 2020-03-31
bump recursive-open-struct for Ruby 2.7 compatibility

breunigs:
1.1.1 adds support for 2.7 while keeping support for older Rubies,
cmp. https://github.com/aetherknight/recursive-open-struct/pull/64#issuecomment-597469281

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -33,6 +33,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'jsonpath', '~> 1.0'
 
   spec.add_dependency 'rest-client', '~> 2.0'
-  spec.add_dependency 'recursive-open-struct', '~> 1.0', '>= 1.0.4'
+  spec.add_dependency 'recursive-open-struct', '~> 1.1', '>= 1.1.1'
   spec.add_dependency 'http', '>= 3.0', '< 5.0'
 end


cben:
Hi, sorry I forgot about this.  
1.1.1 is already _allowed_ by the current kubeclient spec.  So is releasing a new kubeclient version with this requirement useful?  
My thought was that people doing fresh `bundle install` or `bundle update` would get 1.1.1 anyway (unless some _other_ library is preventing that, e.g. by requiring 1.0.z), and people that don't do  `bundle update` can't benefit from a new version of anything :wink:  
Moreover, if you have something else holding back recursive-open-struct, you'll not get the new kubeclient version + 1.1.1, and you'll not get an error highlighting the 2.7 problem.  Bundler will just silently _hold back kubeclient_ to 4.6.0 that still allows old recursive-open-struct :disappointed: 

Am I missing anything?

I do agree it's slightly wrong to _allow_ 1.0.z which won't work on 2.7 :thinking: 
(in an ideal world, https://rubygems.org/gems/recursive-open-struct/versions/1.1.0 and previous would specify ruby version range < 2.7, preventing their use, but they don't specify any range, and I don't think it's possible to retroactively constrain them...)

P.S. I recently landed some incompatible changes in preparation for a 5.0.0.  Not a problem, if this is useful to people, I can backport only this onto 4.6.0 to release 4.6.1 (or or 4.7.0)...

breunigs:
No, I think your assessment is right. You can of course "force" this issue by using `bundle update kubeclient --conservative` which I think is what most of these automated version bumping robots do. 

The other option I see would be to use the `install_if` hook (around since Bundler 1.10) and use that to determine which min version of the gem is required. Even in this case people might run into the issue that an older kubeclient is selected which _doesn't_ specify the right min version, so we run into the same issue again.

I don't have a strong opinion on this one. It will help with the aforementioned `--conservative` update, but it's equally fair to wait and see how many people run into this issue in the first place. I'm guessing as time passes, this issue will solve itself anyway.

Long story short, your call. Feel free to close this PR if you'd prefer to keep it as is.

cben:
Thanks for the details.  Yeah, why not, merging.
Will eventually release in 5.0 (blockers: #435).
Let me know if anyone wants a 4.6.1 backport.

cben:
Released in https://rubygems.org/gems/kubeclient/versions/4.7.0

PR #441, opened 2020-04-22
Add userinfo.email to default scopes in gcp auth provider

cjoudrey:
I was attempting to use `kubeclient` to connect to a GKE cluster by leveraging the `gcp` auth provider.

I had granted the service account access to certain resources through RBAC policies in the destination cluster yet I kept getting this error:

```
User "10345xxxxxxxxx" cannot patch resource "xxxxxxx" in API group "xxxxx"
in the namespace "xxxxx": Required "container.thirdPartyObjects.update" permission.
```

After digging into the issue, we realized the default scope for the `gcp` auth provider is missing `https://www.googleapis.com/auth/userinfo.email`.

Here's an excerpt of [`client-go`'s `gcp` auth provider](https://github.com/kubernetes/kubernetes/blob/a3ccea9d8743f2ff82e41b6c2af6dc2c41dc7b10/staging/src/k8s.io/client-go/plugin/pkg/client/auth/gcp/gcp.go#L49-L55):

```golang
	// defaultScopes:
	// - cloud-platform is the base scope to authenticate to GCP.
	// - userinfo.email is used to authenticate to GKE APIs with gserviceaccount
	//   email instead of numeric uniqueID.
	defaultScopes = []string{
		"https://www.googleapis.com/auth/cloud-platform",
		"https://www.googleapis.com/auth/userinfo.email"}
```

There's also some clarification about this [in GCP's docs](https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl#authentication):

> All GKE clusters are configured to accept Google Cloud user and service account identities, by validating the credentials presented by `kubectl` and retrieving the email address associated with the user or service account identity. As a result, the credentials for those accounts must include the `userinfo.email` OAuth scope in order to successfully authenticate.

@cben as a follow-up to this PR, it might make sense to pass `config` to `GoogleApplicationDefaultCredentials.token` such that users can customize the scopes. Here's [an example from `client-go`](https://github.com/kubernetes/kubernetes/blob/a3ccea9d8743f2ff82e41b6c2af6dc2c41dc7b10/staging/src/k8s.io/client-go/plugin/pkg/client/auth/gcp/gcp.go#L62-L75):

```golang
// {
//   'auth-provider': {
//     # Required
//     "name": "gcp",
//
//     'config': {
//       # Authentication options
//       # These options are used while getting a token.
//
//       # comma-separated list of GCP API scopes. default value of this field
//       # is "https://www.googleapis.com/auth/cloud-platform,https://www.googleapis.com/auth/userinfo.email".
// 		 # to override the API scopes, specify this field explicitly.
//       "scopes": "https://www.googleapis.com/auth/cloud-platform"
```

### DIFF
--- a/lib/kubeclient/google_application_default_credentials.rb
+++ b/lib/kubeclient/google_application_default_credentials.rb
@@ -16,7 +16,12 @@ module Kubeclient
                 'googleauth gem. To support auth-provider gcp, you must include it in your ' \
                 "calling application. Failed with: #{e.message}"
         end
-        scopes = ['https://www.googleapis.com/auth/cloud-platform']
+
+        scopes = [
+          'https://www.googleapis.com/auth/cloud-platform',
+          'https://www.googleapis.com/auth/userinfo.email'
+        ]
+
         authorization = Google::Auth.get_application_default(scopes)
         authorization.apply({})
         authorization.access_token


cjoudrey:
👋😄Hi @cben! Just a gentle nudge. I was wondering if you had time to take a look at this.

cben:
Hi, thanks for the nudge, these days I definitely need it :)

I know almost nothing about GCP.  Thanks for the references, this makes sense.
Do you have an idea whether this is a new requirement?  I'm assuming when GCP support was added, it worked for the author?

Anyway, merging.  It might take me some time to release this, please do nag me!
Do you think this counts as bugfix or enhancement from SemVer perspective?

cjoudrey:
Sorry for the radio silence @cben. I completely missed this notification. 😢 

> Do you have an idea whether this is a new requirement? I'm assuming when GCP support was added, it worked for the author?

My understanding is that there's seemingly two ways of setting up user permissions for a GKE cluster.

The original code worked fine if you wanted to use [IAM permissions](https://cloud.google.com/kubernetes-engine/docs/reference/api-permissions) to grant users access to GKE clusters. For example, you could create a Google service account and grant it access to different Kubernetes resource types like `container.pods.list` to grant it access to list `Pod` resources:

Screen Shot 2020-05-04 at 12 09 14 PM

Where this system breaks down is for custom resources. The granularity of IAM permissions for custom resources is basically: "The user can read/write/update/delete *all* custom resources.". The IAM permissions don't let you pick what `kind` of custom resources the user can edit. It's all or nothing. For example, you cannot say "Grant user read access to `kind: A` but not `kind: B`.

This screenshot of [IAM permissions](https://cloud.google.com/kubernetes-engine/docs/reference/api-permissions) might better explain the above example:

Screen Shot 2020-05-04 at 12 07 56 PM

This is where RBAC policies come into play. They _do_ support that level of granularity.

This brings us back to this PR. In order to define permissions using RBAC policies and authenticate with a Google service account, we need the client to request the `https://www.googleapis.com/auth/userinfo.email` permissions. Behind the scenes, this seems to be  required in order to map Google service account `id` to an `email`. I'm a bit fuzzy on the exact details here.

To go back to your original question, my guess is that the original author was only using IAM permissions and not RBAC policies. For that use case, the client worked fine. 😄 

> Do you think this counts as bugfix or enhancement from SemVer perspective?

I suppose it could go either way. I don't have strong opinions on this.

Maybe it's safer to treat it as an enhancement and say something like GCP provider now supports authenticating with GCP and leveraging RBAC policies to define permissions?

cben:
Released in https://rubygems.org/gems/kubeclient/versions/4.7.0 (#446)

Issue #358, opened 2018-10-29
Support for apply command


jhavero:
Does this gem support applying yaml files declaratively? For example 'kubectl apply -f file.yml'

cben:
Not currently.
You can completely replace an object with `update_pod()` etc.
You can patch specific parts of an object with `patch_pod()` etc. (currently using "strategic merge patch", want to add others, see #357).
There are several hurdles to process a yaml file with many objects — #329.  The next step to make that easier is #332.

But `kubectl apply` has considerate client-side smarts for not trampling over "external" changes to the object — it records the last applied input in an annotation, and then does something akin to 3-way merge of "last applied" — "new to apply" — "actual".
I'm not sure it's productive to replicate this merging in kubeclient...

Any PRs welcome of course :)

I must say at this point, if what you need is _exactly_ the functionality of `apply`, shelling out to `kubectl` is the easy path...

rhysm:
[Server Side Apply](https://kubernetes.io/docs/reference/using-api/api-concepts/#server-side-apply
) is now available as Alpha in Kubernetes 1.14.

>Server Side Apply allows clients other than kubectl to perform the Apply operation, and will eventually fully replace the complicated Client Side Apply logic that only exists in kubectl. If the Server Side Apply feature is enabled, the PATCH endpoint accepts the additional application/apply-patch+yaml content type. Users of Server Side Apply can send partially specified objects to this endpoint. An applied config should always include every field that the applier has an opinion about.

I haven't had a chance to look at it yet but if you can also specify `application/apply-patch+json` to the `PATCH` endpoint then this is supported already specifying `apply-patch` as the strategy to `patch_entity`.

cben:
- [ ] This adds a new `metadata.managedFields` field.  Do we need anything special to support metadata fields?
- [ ] This allows a new `fieldManager` param to various operations, required on apply operation

grosser:
FYI: samson now uses server-side-apply to do this https://github.com/zendesk/samson/pull/3676

benlangfeld:
Abstract server-side apply support is in https://github.com/abonas/kubeclient/pull/448. I didn't address any more generic interface.

PR #448, opened 2020-06-16
Support for server-side apply

benlangfeld:
See https://kubernetes.io/docs/reference/using-api/api-concepts/#server-side-apply

Fixes https://github.com/abonas/kubeclient/issues/358

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,6 +447,9 @@ update_foo(Kubeclient::Resource.new({metadata: {name: 'name', ...}, ...}))  # gl
 
 patch_foo('name', patch, 'namespace')    # namespaced
 patch_foo('name', patch)                 # global
+
+apply_foo(Kubeclient::Resource.new({metadata: {name: 'name', namespace: 'namespace', ...}, ...}), 'myapp')
+apply_foo(Kubeclient::Resource.new({metadata: {name: 'name', ...}, ...}), 'myapp')  # global
 ```
 
 These grew to be quite inconsistent :confounded:, see https://github.com/abonas/kubeclient/issues/312 and https://github.com/abonas/kubeclient/issues/332 for improvement plans.
@@ -701,6 +704,35 @@ patched = client.patch_pod("docker-registry", {metadata: {annotations: {key: 'va
 ```
 
 `patch_#{entity}` is called using a [strategic merge patch](https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch). `json_patch_#{entity}` and `merge_patch_#{entity}` are also available that use JSON patch and JSON merge patch, respectively. These strategies are useful for resources that do not support strategic merge patch, such as Custom Resources. Consult the [Kubernetes docs](https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#use-a-json-merge-patch-to-update-a-deployment) for more information about the different patch strategies.
+
+### Apply an entity
+For example: `apply_pod`
+
+Input parameters - resource (Kubeclient::Resource) representing the desired state of the resource, field_manager (String) to identify the system managing the state of the resource, force (Boolean) whether or not to override a field managed by someone else.
+
+Example:
+
+```ruby
+service = Kubeclient::Resource.new
+service.metadata = {}
+service.metadata.name = "redis-master"
+service.metadata.namespace = 'staging'
+service.spec = {}
+service.spec.ports = [{
+  'port' => 6379,
+  'targetPort' => 'redis-server'
+}]
+service.spec.selector = {}
+service.spec.selector.name = "redis"
+service.spec.selector.role = "master"
+service.metadata.labels = {}
+service.metadata.labels.app = 'redis'
+service.metadata.labels.role = 'slave'
+
+client.apply_service(service, 'myapp')
+```
+
+See https://kubernetes.io/docs/reference/using-api/api-concepts/#server-side-apply
 
 ### Get all entities of all types : all_entities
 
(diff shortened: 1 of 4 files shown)


lib/kubeclient/common.rb
```diff
@@ -415,6 +420,20 @@ def patch_entity(resource_name, name, patch, strategy, namespace)
       format_response(@as, response.body)
     end
 
+    def apply_entity(resource_name, resource, field_manager, force = nil)
```
grosser, line 423:
```suggestion
    def apply_entity(resource_name, resource, field_manager:, force: nil)
```

README.md
```diff
@@ -702,6 +705,35 @@ patched = client.patch_pod("docker-registry", {metadata: {annotations: {key: 'va
...
+### Apply an entity
+For example: `apply_pod`
+
+Input parameters - resource (Kubeclient::Resource) representing the desired state of the resource, field_manager (String) to identify the system managing the state of the resource, force (Boolean) whether or not to override a field managed by someone else.
+
+Example:
+
+```ruby
+service = Kubeclient::Resource.new
+service.metadata = {}
```
grosser, line 717: can we assign a big hash here to reduce the length of the example ?
(or apply a hash directly)

benlangfeld: I just stole a similar example from a few lines up for consistency, but yes, can do.

benlangfeld:
@grosser Feedback addressed.

README.md
```diff
@@ -447,6 +447,9 @@ update_foo(Kubeclient::Resource.new({metadata: {name: 'name', ...}, ...}))  # gl
 
 patch_foo('name', patch, 'namespace')    # namespaced
 patch_foo('name', patch)                 # global
+
+apply_foo(Kubeclient::Resource.new({metadata: {name: 'name', namespace: 'namespace', ...}, ...}), 'myapp')
+apply_foo(Kubeclient::Resource.new({metadata: {name: 'name', ...}, ...}), 'myapp')  # global
```
cben, line 452: I think these should say `field_manager: 'myapp'` right?

Should we mention optional `force:` here?  The focus here is to summarize how name & namespace are passed, not all options.
So let's use `, field_manager: 'myapp', **opts)`, or just `, **opts)`.

README.md
```diff
@@ -702,6 +705,29 @@ patched = client.patch_pod("docker-registry", {metadata: {annotations: {key: 'va
 
 `patch_#{entity}` is called using a [strategic merge patch](https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch). `json_patch_#{entity}` and `merge_patch_#{entity}` are also available that use JSON patch and JSON merge patch, respectively. These strategies are useful for resources that do not support strategic merge patch, such as Custom Resources. Consult the [Kubernetes docs](https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#use-a-json-merge-patch-to-update-a-deployment) for more information about the different patch strategies.
 
+### Apply an entity
```
cben, line 708: Maybe add some context here:

> This is similar to `kubectl apply --server-side` (kubeclient doesn't implement logic for client-side apply).

and move reference link here?  Your call.

lib/kubeclient/common.rb
```diff
@@ -254,6 +254,10 @@ def define_entity_methods
         do |name, patch, namespace = nil|
           patch_entity(entity.resource_name, name, patch, 'merge-patch', namespace)
         end
+
+        define_singleton_method("apply_#{entity.method_names[0]}") do |*args|
+          apply_entity(entity.resource_name, *args)
```
cben, line 259: Should this use `**opts` or `*args, **opts` for Ruby 2.7+ compatibility?
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/#delegation
actually suggests not using `**opts` in code that has to run on Ruby 2.6 or earlier.  But the motivation seems to be a corner case with empty opts that won't happen here if field_manager: is mandatory?  My head hurts thinking about this...

Would it work to define the block with the actual params we expect?
```suggestion
        define_singleton_method("apply_#{entity.method_names[0]}") \
        do |resource, field_manager:, force: true|
          apply_entity(entity.resource_name, resource, field_manager: field_manager, force: force)
```
Not critical, not sure it's better, just thinking aloud...
Current code passed CI, including 2.7, so it's also good.

cben:
Thanks, overall looks great :+1:

lib/kubeclient/common.rb
benlangfeld: > actually suggests not using `**opts` in code that has to run on Ruby 2.6 or earlier. But the motivation seems to be a corner case with empty opts that won't happen here if field_manager: is mandatory? My head hurts thinking about this...

Indeed using `**opts` breaks:

```
ArgumentError: wrong number of arguments (given 1, expected 0)
    /Users/blangfeld/code/kubeclient/lib/kubeclient/common.rb:258:in `block (2 levels) in define_entity_methods'
    /Users/blangfeld/code/kubeclient/lib/kubeclient/common.rb:101:in `method_missing'
```

If we want true forward compatibility with 2.7/3, we need the ruby2_keywords gem, but I'm not sure if that works with procs or just methods.

> Would it work to define the block with the actual params we expect?

I did that, but it [was suggested to do otherwise](https://github.com/abonas/kubeclient/pull/448#discussion_r441097839). I'll do whatever you're willing to merge.

cben:
LGTM :100:

benlangfeld:
Thanks @cben . Is it possible to cut a release?

benlangfeld:
@cben Do you know when this might make it into a release?

cben:
Thanks for nagging :+1:  Released 4.8.0.

PR #450, opened 2020-07-01
Declare runtime dependency on jsonpath for `gcp` provider with `cmd-path`

### LINKED ISSUE #444: jsonpath dependency incorrectly listed
jsonpath is listed as a development dependency https://github.com/abonas/kubeclient/blob/cab82dbefd51d26d2399469a5a3357a3e98df0f1/kubeclient.gemspec#L33
but is directly called by
https://github.com/abonas/kubeclient/blob/cab82dbefd51d26d2399469a5a3357a3e98df0f1/lib/kubeclient/gcp_command_credentials.rb#L27

This seems like a mistake?

cben:
Fixes #444
It was a implicit dependency similar to other auth provider dependencies being "optional", having users depend on them themselves if needed — except we'd never documented this! :facepalm:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Kubeclient release versioning follows [SemVer](https://semver.org/).
 
 ### Added
 - Support for server-side apply (#448).
+- Declare dependency on jsonpath, needed for `gcp` provider with `cmd-path` (#450).
 
 ## 4.7.0 — 2020-06-14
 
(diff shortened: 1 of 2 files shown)


cben:
@Donald47 @lucasmazza sorry for delay, please review.

Donald47:
Thanks @cben

cben:
Released 4.8.0.

PR #451, opened 2020-07-03
Release 4.8.0

cben:
Releasing 4.8.0, includes #448 and #450.


cben:
And merged back into `master` (see #435 why I'm temporarily releasing from v4.y rather than master)

PR #453, opened 2020-07-30
Add support for client-type ExecCredentials

sczizzo:
This library currently supports `token`-type ExecCredentials, but there is actually another flavor for TLS client auth. With this flavor, the `status` field includes PEM-encoded `clientKeyData` and `clientCertificateData` rather than a `token`. Reference:
- https://kubernetes.io/docs/reference/access-authn-authz/authentication/#input-and-output-formats
- https://godoc.org/k8s.io/client-go/pkg/apis/clientauthentication#ExecCredentialStatus

This PR adds support for TLS client auth while maintaining support for token ExecCredentials. To accomplish this, the credentials provided by the `status` field, regardless of flavor, are passed around as if it were a `user`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 Kubeclient release versioning follows [SemVer](https://semver.org/).
 
+## 4.9.0 - 2020-08-03
+### Added
+- Support for `user: exec` credential plugins using TLS client auth (#453)
+
 ## 4.8.0 — 2020-07-03
 
 ### Added
--- a/lib/kubeclient/config.rb
+++ b/lib/kubeclient/config.rb
@@ -43,7 +43,7 @@ module Kubeclient
 
       if user.key?('exec')
         exec_opts = expand_command_option(user['exec'], 'command')
-        user = ExecCredentials.run(exec_opts)
+        user['exec_result'] = ExecCredentials.run(exec_opts)
       end
 
       ca_cert_data     = fetch_cluster_ca_data(cluster)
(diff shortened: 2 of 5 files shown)


lib/kubeclient/exec_credentials.rb
```diff
@@ -34,6 +34,31 @@ def validate_opts(opts)
...
+      end
+
+      def validate_credentials_status(status)
+        raise 'exec plugin didn\'t return a status field' if status.nil?
+
+        has_client_credentials = validate_client_credentials_status(status)
+        return if has_client_credentials
+
+        has_token = status.key?('token')
+        raise 'exec plugin didn\'t return a token' unless has_token
```
cben, line 59: At this point, I think you can say it returned neither token nor a client certificate.

lib/kubeclient/config.rb
```diff
@@ -142,16 +149,15 @@ def fetch_user_key_data(user)
         File.read(ext_file_path(user['client-key']))
       elsif user.key?('client-key-data')
         Base64.decode64(user['client-key-data'])
+      elsif user.key?('clientKeyData')
+        user['clientKeyData']
```
cben, line 153: It's slightly confusing to have 2 similarly named keys, one base64, one not.
Although both spellings come from k8s.

The reuse of `user` is clever but WDYT of putting it in a sub-field instead?  `user.exec_result = ExecCredentials.run(exec_opts)`
It does mean you'll have to check .token in 2 places, but I think the data flow will be more readable.

cben:
This is great, thanks!
Please add a CHANGELOG entry.

lib/kubeclient/exec_credentials.rb
sczizzo: Done. Also added check to ensure a token XOR client creds are returned.

lib/kubeclient/config.rb
sczizzo: I like it. Done.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 Kubeclient release versioning follows [SemVer](https://semver.org/).
 
+## 4.9.0 - 2020-08-03
```
cben, line 7: Well played :wink:.  Released now.

PR #457, opened 2020-08-12
Implement usage of nested APIs

### LINKED ISSUE #418: Request K8s API, But it return bad-request (400)
versions
```
ruby: 2.5.0
k8s client version: 1.13.4
k8s server version: 1.14.5
kubeclient: 4.4.0
```

My code:

```ruby
require 'kubeclient'
require 'json'
require 'open3'

content = File.open("nginx.yaml").read

p suffix_name = File.extname("nginx.yaml")

case suffix_name
when '.yaml' then
    hash = YAML.safe_load(content)
    p "yaml content is:#{hash}"
when '.json' then
    hash = JSON.parse(content)
    p "json content is:#{hash}"
end

hash.each do |key, value|
    puts "#{key} is: #{value}"
    if key == "kind"
    @kind = value.to_s.downcase
    end
end

config = Kubeclient::Config.read("rancher_kube_config.kubeconfig")

kubeconfig_text = File.read("rancher_kube_config.kubeconfig")
config_context = Kubeclient::Config.new(YAML.safe_load(kubeconfig_text), nil)
@config = config_context.instance_eval { @kcfg }
config = @config
deployment = Kubeclient::Resource.new(hash)
cert_store = OpenSSL::X509::Store.new
ssl_options = {
	cert_store: cert_store,
	verify_ssl: OpenSSL::SSL::VERIFY_NONE
}
context = config_context.context
auth_options = {
		bearer_token: context.auth_options[:bearer_token]
}
client = Kubeclient::Client.new('https://{my_host}/k8s/clusters/c-h4mdn/api', auth_options: auth_options, ssl_options: ssl_options)

begin
  @create_entity = client.send("create_#{@kind}", deployment)
rescue Kubeclient::HttpError => e
	p e
	@error_code = e.error_code
	@message = e.message
end

if @create_entity.nil?
	p "false"
else
	p "success"
end

```

The contents of `config.kubeconfig`:

```YAML
apiVersion: v1
kind: Config
clusters:
- name: "k8s"
  cluster:
    server: "https://MY_HOST/k8s/clusters/c-h4mdn"
    certificate-authority-data: ...
- name: "k8s-k8smaster"
  cluster:
    server: "https://MY_HOST:6443"
    certificate-authority-data: ...

users:
- name: "user-8lhbv"
  user:
    token: my_token

contexts:
- name: "k8s"
  context:
    user: "user-8lhbv"
    cluster: "k8s"
- name: "k8s-k8smaster"
  context:
    user: "user-8lhbv"
    cluster: "k8s-k8smaster"

current-context: "k8s"
```

The contents of `nginx.yaml`:

```yaml
apiVersion: apps/v1
kind: Pod
metadata:
  name: nginx-test2
  labels:
     app: nginx
  namespace: default
spec:
     containers:
        - name: nginx
          image: nginx
          imagePullPolicy: IfNotPresent
          ports:
          - containerPort: 80
     restartPolicy: Always
```

The error I'm getting is this, and I think it maybe the `apiVersion` is wrong:

```
#<Kubeclient::HttpError: HTTP status code 400, Pod in version "v1" cannot be handled as a Pod: no kind "Pod" is registered for version "clusters/v1" in scheme "k8s.io/kubernetes/pkg/api/legacyscheme/scheme.go:29" for POST https://MY_HOST/k8s/clusters/c-h4mdn/api/v1/namespaces/default/pods>
```

T4cC0re:
`handle_uri` expected the Kubernetes API to be rooted at `/` if more than 2
components were included in the provided URL. This was done to extract a
potential `api_group` from the URL.

This commit changes that behaviour, to parse that `api_group` via a regex,
rather than leaning on a strict URL pattern.
This commit also takes into account, that OpenShift uses `/oapi`, as well
as including a test-suite for that new API URL handling.

Implements #318
Fixes #418

### DIFF
--- a/test/test_common_url_handling.rb
+++ b/test/test_common_url_handling.rb
@@ -1,0 +1,160 @@
+require_relative 'test_helper'
+
+# URLHandling tests
+class TestCommonUrlHandling < MiniTest::Test
+  def test_no_path_in_uri
+    client = Kubeclient::Client.new('http://localhost:8080', 'v1')
+    rest_client = client.rest_client
+    assert_equal('v1', client.instance_variable_get(:@api_version))
+    assert_equal('', client.instance_variable_get(:@api_group))
+    assert_equal('http://localhost:8080/api/v1', rest_client.url.to_s)
+  end
+
+  def test_with_api_path_in_uri
+    client = Kubeclient::Client.new('http://localhost:8080/api', 'v1')
+    rest_client = client.rest_client
+    assert_equal('v1', client.instance_variable_get(:@api_version))
+    assert_equal('', client.instance_variable_get(:@api_group))
+    assert_equal('http://localhost:8080/api/v1', rest_client.url.to_s)
+  end
+
+  def test_with_api_path_in_uri_other_version
+    client = Kubeclient::Client.new('http://localhost:8080/api', 'v2')
+    rest_client = client.rest_client
+    assert_equal('v2', client.instance_variable_get(:@api_version))
+    assert_equal('', client.instance_variable_get(:@api_group))
+    assert_equal('http://localhost:8080/api/v2', rest_client.url.to_s)
+  end
+
+  def test_with_api_group_path_in_uri
+    client = Kubeclient::Client.new('http://localhost:8080/apis/this_is_the_group', 'v1')
+    rest_client = client.rest_client
+    assert_equal('v1', client.instance_variable_get(:@api_version))
+    assert_equal('this_is_the_group/', client.instance_variable_get(:@api_group))
+    assert_equal('http://localhost:8080/apis/this_is_the_group/v1', rest_client.url.to_s)
+  end
+
+  def test_with_api_group_path_in_uri_other_version
+    client = Kubeclient::Client.new('http://localhost:8080/apis/this_is_the_group', 'v2')
+    rest_client = client.rest_client
+    assert_equal('v2', client.instance_variable_get(:@api_version))
+    assert_equal('this_is_the_group/', client.instance_variable_get(:@api_group))
+    assert_equal('http://localhost:8080/apis/this_is_the_group/v2', rest_client.url.to_s)
+  end
+
+  def test_with_api_path_in_uri_trailing_slash
+    client = Kubeclient::Client.new('http://localhost:8080/api/', 'v1')
+    rest_client = client.rest_client
+    assert_equal('v1', client.instance_variable_get(:@api_version))
+    assert_equal('', client.instance_variable_get(:@api_group))
+    assert_equal('http://localhost:8080/api/v1', rest_client.url.to_s)
+  end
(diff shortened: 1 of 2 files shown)


T4cC0re:
This would also yield to fix https://gitlab.com/gitlab-org/gitlab/-/issues/22043

cben:
This is awesome and very important, thanks a lot for tackling this! :heart:
Thanks especially for the detailed tests.

- [x] The `/api/but/I/want/a/hidden/k8s/api` cases are excellent subtle corner case to check :clap: 
- [ ] Let's add a test for `/k8s/clusters/c-somerancherID/apis/this_is_the_group` for complete coverage of Rancher-style URLs.
  (I expect it works if `api/but/I/want/a/hidden/k8s/apis/this_is_the_group` test works, but the more the merrier)
- [ ] What these tests are checking is that the final re-assembled url is correct.
  Can you also check that `@api_group` and `@api_version` instance vars are correct?
  I believe in some cases the url was already correct but wrong `@api_group` and `@api_version` caused create operations to fail (due to sending non-existent `apiVersion`).
- 'v1' is already the default, so let's use something else in tests e.g. 'v2' to confirm it's respected.
  (I'm certain it is, default is only when omitting 2nd arg entirely, but still would make a stronger test)

T4cC0re:
Thanks for those pointers. I will take a look at that :)

T4cC0re:
@cben I just force-pushed an amended commit, that includes the tests you asked for. Would you mind reviewing it? :)

T4cC0re:
Just saw, that rubocop failed. Will fix that real quick

test/test_common_url_handling.rb
```diff
@@ -0,0 +1,160 @@
...
+    assert_equal('v2', client.instance_variable_get(:@api_version))
+    assert_equal('', client.instance_variable_get(:@api_group))
+    assert_equal('http://localhost:8080/api/v2', rest_client.url.to_s)
+  end
+
+  def test_with_api_group_path_in_uri
+    client = Kubeclient::Client.new('http://localhost:8080/apis/this_is_the_group', 'v1')
+    rest_client = client.rest_client
+    assert_equal('v1', client.instance_variable_get(:@api_version))
+    assert_equal('this_is_the_group/', client.instance_variable_get(:@api_group))
```
cben, line 33: The trailing slash looked surprising to me, but yep, we had it before and we're sending
`hash[:apiVersion] = @api_group + @api_version`
so this is good :heavy_check_mark:.
(Also this representation is convenient for automatically handling both `foo/v1` and core `v1`)

cben (approved):
Perfect :100:
Thanks again!

cben:
I'll release this by Sunday.

cben:
Well, for a sufficiently late definition of "Sunday".  Released 4.9.1.

Issue #460, opened 2020-09-01
Kubeclient does not support certificate authorities with intermediate certificate


jperville:
## The issue

We tried to run a Ruby application (using the kubeclient gem) on a kubernetes cluster which uses a custom CA (the cluster's CA itself is signed by another custom CA, hence the need for intermediate certificates).

Kubeclient (initialized from the KUBECONFIG file) fails with SSL verify errors.

## How to reproduce

First, we create a `KUBECONFIG` file in the container by executing the following script:
```bash
write_client_kubeconfig() {
  KUBECTL=${1:?please provide the KUBECTL environment variable}

  # only needed for writing a kubeconfig:
  master_url=${MASTER_URL:-https://kubernetes.default.svc.cluster.local:443}
  master_ca=${MASTER_CA:-/var/run/secrets/kubernetes.io/serviceaccount/ca.crt}
  token_file=${TOKEN_FILE:-/var/run/secrets/kubernetes.io/serviceaccount/token}

  # set up configuration for openshift client
  if [ -n "${WRITE_KUBECONFIG:-''}" ]; then
      # craft a kubeconfig, usually at $KUBECONFIG location
      ${KUBECTL} config set-cluster master \
            --certificate-authority="${master_ca}" \
            --server="${master_url}"
      ${KUBECTL} config set-credentials account \
            --token="$(cat ${token_file})"
      ${KUBECTL} config set-context current \
            --cluster=master \
            --user=account \
            --namespace="${infra_project}"
      ${KUBECTL} config use-context current
  fi
}

write_client_kubeconfig kubectl
```

Then we try listing services cluster-wide using the `kubectl` binary:

```bash
kubectl get services --all-namespaces
```
This should work, assuming that the current service account is allowed to list services cluster-wide.

Finally, we try listing the same services in Ruby using `kubeclient`:

```bash
mkdir -p /tmp/test
cd /tmp/test

cat <<EOF | tee Gemfile
source 'https://rubygems.org'
gem 'kubeclient', '~> 4.8'
EOF

bundle install --path .bundle

cat <<EOF | tee test.rb
require 'kubeclient'

config = Kubeclient::Config.read(ENV.fetch('KUBECONFIG'))
context = config.context
ssl_options = context.ssl_options
auth_options = context.auth_options

client = Kubeclient::Client.new(
    context.api_endpoint, 'v1',
    ssl_options: ssl_options, auth_options: auth_options
)

[...]

PR #461, opened 2020-09-01
Load cluster ca certificates using OpenSSL::X509::Store#add_file

jperville:
Using `OpenSSL::X509::Store#add_file` instead of `OpenSSL::X509::Store#add_cert` in lib/kubeclient/config.rb
allows to load all certificates concatenated in the cacert file, instead of only the first one.

This PR should fix issue #460.


cben:
LGTM :100:  Thanks for the detailed explanation on #460.

PR #462, opened 2020-09-02
Custom class passed for `as` option

kddeisz:
There are some times where it's useful to be able to pass your own domain objects around instead of a `RecursiveOpenStruct`. You could accomplish this by just letting it return `:raw` and then wrapping it later, as in:

```ruby
MyObject.new(client.get_services(as: :raw))
```

but that can get repetitive, especially if you're trying to do it over every call the client makes. This commit instead adds the ability to pass in your own class, so the equivalent from the previous example would be:

```ruby
client.get_services(as: MyObject)
```

so that you can do your own processing.


cben:
Neat, I like this :+1:

PR #467, opened 2020-11-04
make version required argument to avoid confusion and make readme exa…

grosser:
…mples consistent

there is more work in this area, like `'host/apis/batch', 'v1'` being weird
but I think this is a nice start on removing confusion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 Kubeclient release versioning follows [SemVer](https://semver.org/).
 
+## Unreleased
+### Changed
+- `Kubeclient::Client.new` always required an api version, use for example: `Kubeclient::Client.new(uri, 'v1')`
+
 ## 4.9.1 — 2020-08-31
 ### Fixed
 - Now should work with apiserver deployed not at root of domain but a sub-path,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Kubeclient release versioning follows [SemVer](https://semver.org/).
 
 ## Unreleased
 ### Changed
-- `Kubeclient::Client.new` always required an api version, use for example: `Kubeclient::Client.new(uri, 'v1')`
+- `Kubeclient::Client.new` now always requires an api version, use for example: `Kubeclient::Client.new(uri, 'v1')`
 
 ## 4.9.1 — 2020-08-31
 ### Fixed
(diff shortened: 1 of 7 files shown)


grosser:
@cben thoughts on this ?

cben:
Not a bad idea.
This doesn't gain much, but it small positive change towards untangling the API groups mess :+1:.

- One question I like to ask about breaking changes is whether the required rewrite for users is tightly coupled to the upgrade.  It's better when there is a major version which supports both old and new interface — this allows piece-wise rewrites.
  This satisfies that because people can add explicit `, 'v1'` in previous version :heavy_check_mark:. 

I'd like to hold this back for now and decide close to 5.0 :thinking: release but I'm inclined to include it.

- [ ] We'll need to list this in CHANGELOG, and in "Upgrading past version 5.0" section of README.
- [ ] What happens if someone runs old code omitting version?  Is the error message clear?

cben:
(and yes, I'm totally over-thinking this release :laughing:)

grosser:
- prefer no Upgrading section in the readme, that's for the changelog and gets out of date quick leaving obsolete info in the readme
- users that omit it get a nice error message (newly added raise)

cben:
OK, let's remove the existing Upgrading section then.

grosser:
https://github.com/abonas/kubeclient/pull/476

grosser:
added changelog

test/test_kubeclient.rb
```diff
@@ -34,18 +34,12 @@ def test_no_path_in_uri
     assert_equal('http://localhost:8080/api/v1', rest_client.url.to_s)
   end
 
-  def test_no_version_passed
```
astencel-sumo, line 37: How about we leave this test in and assert it raises that ArgumentError?

grosser: done

astencel-sumo: Rubocop complaining now 👮

grosser: fixed!

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 Kubeclient release versioning follows [SemVer](https://semver.org/).
 
+## Unreleased
+### Changed
+- `Kubeclient::Client.new` always required an api version, use for example: `Kubeclient::Client.new(uri, 'v1')`
```
cben, line 9:
```suggestion
- `Kubeclient::Client.new` now always requires an api version, use for example: `Kubeclient::Client.new(uri, 'v1')`
```

cben:
Yeah, let's do this :rocket:.  Please resolve conflict.

grosser:
rebased

PR #468, opened 2020-11-04
fix some funky looking indents

grosser:
there is more, but this is a start

they always look out of place with the normal indentation and have to change all lines when the first line changes,
so standardize on 2-space indent 🤷


cben:
OK 👍

PR #469, opened 2020-11-04
remove safe_load warnings

grosser:
```
Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
```

### DIFF
--- a/lib/kubeclient/config.rb
+++ b/lib/kubeclient/config.rb
@@ -30,7 +30,8 @@ module Kubeclient
 
     # Builds Config instance by parsing given file, with lookups relative to file's directory.
     def self.read(filename)
-      parsed = YAML.safe_load(File.read(filename), [Date, Time])
+      options = (RUBY_VERSION >= '2.6' ? { permitted_classes: [Date, Time] } : [Date, Time])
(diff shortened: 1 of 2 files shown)


cben:
Thanks!

Let's drop EOL ruby versions.  Will that suffice for tests to pass?

cben:
Close-cycling this to test Github Actions.

Because so far we had only the merge enabling them and https://github.com/abonas/kubeclient/actions/runs/361333140 showed "No jobs were run"...

cben:
Travis is running — and yes, _seriously_ slow :(  Backlog is now at ~2500 jobs: https://www.traviscistatus.com/#week

Actions are not running.  I'm entirely new to Actions but think the workflow file is missing `on:` section.  Probably want `pull_request` and `push` ?
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#push
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

grosser:
@cben ready

cben:
LGTM

PR #471, opened 2020-11-05
remove EOL ruby version support

grosser:
unblocks https://github.com/abonas/kubeclient/pull/469


cben:
Nice, thanks for the small details like timeout doc

PR #477, opened 2020-11-16
use frozen strings

grosser:


lib/kubeclient/watch_stream.rb
```diff
@@ -21,7 +23,7 @@ def each
           raise Kubeclient::HttpError.new(response.code, response.reason, response)
         end
 
-        buffer = ''
+        buffer = +''
```
cben, line 26: TIL https://metaredux.com/posts/2019/05/10/weird-ruby-positive-and-negative-strings.html

A bit too magic to my taste, I'd consider `''.dup` to be clearer, but RuboCop::Cop::Performance::UnfreezeString explains "Unary plus operator is faster than `String#dup`." so never mind :shrug:

PR #479, opened 2020-11-16
remove deprecated exception

grosser:


cben:
See my comment https://github.com/abonas/kubeclient/pull/293#issuecomment-467421832 on similar attempt.
I think we do want a base class (or module) that will cover all our exception non-http errors too.
However, it's good time to move it to our namespace e.g. `Kubeclient::Error`?  I'll try to make an alternative PR soon.

PR #481, opened 2020-11-17
remove jsonpath dependency

grosser:
@cben 
/cc @lucasmazza in case I forgot a usecase

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -32,7 +32,6 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'faraday', '~> 1.1'
   spec.add_dependency 'faraday_middleware', '~> 1.0'
-  spec.add_dependency 'jsonpath', '~> 1.0'
   spec.add_dependency 'recursive-open-struct', '~> 1.1', '>= 1.1.1'
   spec.add_dependency 'http', '>= 3.0', '< 5.0'
 end
--- a/lib/kubeclient/gcp_command_credentials.rb
+++ b/lib/kubeclient/gcp_command_credentials.rb
@@ -8,7 +8,6 @@ module Kubeclient
         require 'open3'
         require 'shellwords'
         require 'json'
-        require 'jsonpath'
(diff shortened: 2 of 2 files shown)


cben:
Sorry for silence.

- Theorethically the format as defined by the Go client supports arbitrary commands and jsonpath:
  https://github.com/kubernetes/kubernetes/blob/v1.20.0/staging/src/k8s.io/client-go/plugin/pkg/client/auth/gcp/gcp.go#L58-L109
  so replacing it with a small subset of jsonpath _could_ be a regression for somebody.
- In practice, when used for GCP I expect the command is always `gcloud config config-helper --output=json`, and the output fields to take are `{.credential.access_token}`, `{.credential.token_expiry}`.
  - [Here](https://github.com/kubernetes-client/python/issues/754) is an example of people that used this with a different command, for a non-GCP use case.  But I guess that's much rarer now that generic `exec` provider exists...
- For comparison, the Python implementation carefully takes `token-key` and `expiry-key` from config, then ...  
  ignores them and hardcodes `data['credential']['access_token']`, `data['credential']['token_expiry']` :laughing: 
  https://github.com/jfrabaute/python-base/blob/b87a5fe119b820e9196f39343b0f5a1546d7465b/config/kube_config.py#L173-L175

I guess this is safe enough in practice :heavy_check_mark: 
However, the null hypotesis is that it's easier to rely on an implementation somebody else maintains & tests :grin:
(1) what's the motivation? (2) if you want this tiny-subset here, please add a couple tests.
I see you also did https://github.com/joshbuddy/jsonpath/pull/130 — are these related?

grosser:
the jsonpath PR was to make it not show warnings, but that was before I realized that this is a mostly unnecessary dependency

grosser:
- motivation is to have 1 less dependency for everyone to install for a feature very few ppl use + less updating/svm etc
- it's already tested with `'token-key' => '{.credential.access_token}'` and `{.credential.token_expiry}` is not used

PR #482, opened 2020-11-20
Use `require_relative` for local files

astencel-sumo:
This makes it easier for me to test locally in my scenario; anyway I think this is the "correct" way to require local files? Correct me if I'm wrong please.


cben:
Not sure which is "better" but as long as it works, sure why not.

PR #488, opened 2021-01-04
Replace http gem with faraday

astencel-sumo:
This pull request replaces the `http` gem with Faraday, in line with a [previous PR](https://github.com/abonas/kubeclient/pull/466) that replaced `rest-client` with Faraday. After this, the only HTTP library used by Kubeclient will be Faraday. This should satisfy https://github.com/abonas/kubeclient/issues/237.

Please let me know what you think, I'm happy to take any advice that will help me push this to completion.

### DIFF
--- a/lib/kubeclient/watch_stream.rb
+++ b/lib/kubeclient/watch_stream.rb
@@ -1,98 +1,70 @@
 # frozen_string_literal: true
 
 require 'json'
-require 'http'
+
 module Kubeclient
   module Common
     # HTTP Stream used to watch changes on entities
     class WatchStream
-      def initialize(uri, http_options, formatter:)
+      def initialize(uri, options, formatter:)
         @uri = uri
-        @http_client = nil
-        @http_options = http_options
-        @http_options[:http_max_redirects] ||= Kubeclient::Client::DEFAULT_HTTP_MAX_REDIRECTS
+        @options = options
+        @headers = options[:headers]
+        @options[:http_max_redirects] ||= Kubeclient::Client::DEFAULT_HTTP_MAX_REDIRECTS
         @formatter = formatter
+
+        @faraday_client = build_client
       end
 
       def each
         @finished = false
-
-        @http_client = build_client
-        response = @http_client.request(:get, @uri, build_client_options)
-        unless response.code < 300
-          raise Kubeclient::HttpError.new(response.code, response.reason, response)
-        end
-
         buffer = +''
-        response.body.each do |chunk|
-          buffer << chunk
-          while (line = buffer.slice!(/.+\n/))
-            yield(@formatter.call(line.chomp))
+
+        begin
+          @faraday_client.get('', nil, @headers) do |request|
+            request.options.on_data = proc do |chunk|
+              buffer << chunk
+              while (line = buffer.slice!(/.+\n/))
+                yield(@formatter.call(line.chomp))
+              end
+              next if @finished
+            end
           end
+        rescue Faraday::Error => e
+          err_message = build_http_error_message(e)
+          response_code = e.response ? (e.response[:status] || e.response&.env&.status) : nil
+          error_klass = (response_code == 404 ? ResourceNotFoundError : HttpError)
+          raise error_klass.new(response_code, err_message, e.response)
         end
-      rescue StandardError
-        raise unless @finished
       end
 
       def finish
         @finished = true
-        @http_client&.close
       end
 
       private
 
-      def max_hops
-        @http_options[:http_max_redirects] + 1
-      end
-
-      def follow_option
-        if max_hops > 1
-          { max_hops: max_hops }
-        else
-          # i.e. Do not follow redirects as we have set http_max_redirects to 0
-          # Setting `{ max_hops: 1 }` does not work FWIW
-          false
-        end
-      end
-
       def build_client
-        client = HTTP::Client.new(follow: follow_option)
+        auth = @options[:auth_options] || {}
+        max_redirects = @options[:http_max_redirects]
 
-        if @http_options[:basic_auth_user] && @http_options[:basic_auth_password]
-          client = client.basic_auth(
-            user: @http_options[:basic_auth_user],
-            pass: @http_options[:basic_auth_password]
-          )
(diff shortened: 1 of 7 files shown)


lib/kubeclient/watch_stream.rb
```diff
@@ -1,98 +1,70 @@
...
-            yield(@formatter.call(line.chomp))
+
+        begin
+          @faraday_client.get('', nil, @headers) do |request|
+            request.options.on_data = proc do |chunk|
+              buffer << chunk
+              while (line = buffer.slice!(/.+\n/))
+                yield(@formatter.call(line.chomp))
+              end
+              next if @finished
```
cben, line 30: The main purpose of `finish` method was to allow interrupting a watch _from another thread_.

The previous implementation by `@http_client&.close` was kinda violent but worked — closing the client would cause the code in `each` to crash in various ways, depending on where exactly it was when it gets closed, and `rescue StandardError` would swallow the crash (when caused by `finish`).

A synchronous `next if @finished` here is much simpler & safer, but I think it'll only stop the loop after it had something to read, which can take unbounded time (until server sends next update)?

cben (commented):
Code looks excellent :clap: :+1:  Haven't tested this yet.

Now that it uses same library as base `Client`, I have a feeling we could share more code,
but not a blocker.

lib/kubeclient/watch_stream.rb
astencel-sumo: Hm yes that's a very good observation. I'll try to confirm this with a test and then fix it, hopefully having come up with a solution. ;)

astencel-sumo:
@cben I have added the test that tries to close the watch connection from another thread when there are no updates from the API server and, just as you described, the `WatchStream#finish` method does not do its job in this scenario.

I then added a call to `Faraday::Connection#close` in the body of the `WatchStream#finish` method, but this didn't help. I'm afraid this is due to the `#close` method not being implemented in the Faraday adapter for `net/http`.

I'm not sure what we can do with this to be honest. Is there a different way to forcibly close the connection? I couldn't find one. Can we get rid of the `finish` functionality? Or should we rather change the HTTP libarary to be able to support the `#finish` use case? We could try some other Faraday backend or switch from Faraday to something else...

cben:
Thanks for looking into this.  Neat test!

Hmm, I'd like to play with this too.
We don't really need `finish` to close it immediately, just "soon" (this was already the semantics but should document it).
Perhaps a non-blocking read with a timeout might help?

I think this is not an immediate deal-breaker.  We can in principle release a breaking 5.0 without `finish` and hopefully restore it in 5.1...

Can you mark the test pending/skipped and document that `finish` no longer works across threads, and I'll merge this.
Or maybe remove it completely — not really useful from inside the watching thread (can just `return` / `break` from the block), and it's better if someone upgrading to 5.0 but still calling `finish` to get an error rather than getting stuck?  WDYT?

cben:
Faraday documents `close` as:
>     # Close any persistent connections. The adapter should still be usable
>     # after calling close.
-- https://github.com/lostisland/faraday/blob/e41668ee591735a8be65fd739cddb7a27518eabd/lib/faraday/adapter.rb#L63-L68

And as far as I can tell, neither [Faraday::Adapter::NetHttpPersistent](https://github.com/lostisland/faraday/blob/master/lib/faraday/adapter/net_http_persistent.rb) nor [Faraday::Adapter::NetHttp](https://github.com/lostisland/faraday-net_http/blob/main/lib/faraday/adapter/net_http.rb) implement `close`, they just inherit the empty base implementation :frowning_face: 

Is there some way to reach in and actually close the connection? Or close underlying OS socket? ("if violence doesn't help, you're not using enough of it")

- With persistent connections, does that risk breaking some unrelated request that was going to reuse same connection?

BTW, is the test failure with persistent adapter or NetHttp?

astencel-sumo:
> And as far as I can tell, neither [Faraday::Adapter::NetHttpPersistent](https://github.com/lostisland/faraday/blob/master/lib/faraday/adapter/net_http_persistent.rb) nor [Faraday::Adapter::NetHttp](https://github.com/lostisland/faraday-net_http/blob/main/lib/faraday/adapter/net_http.rb) implement `close`, they just inherit the empty base implementation ☹️

Yes, exactly!

> Is there some way to reach in and actually close the connection? Or close underlying OS socket? ("if violence doesn't help, you're not using enough of it")

Couldn't find a way, but not an expert in either Ruby or OS/network programming. I wonder if @grosser could help here?

> BTW, is the test failure with persistent adapter or NetHttp?

I only tested this with `net/http` but as noted above, I strongly believe `net/http/persistent` will behave in the same way as it does not implement the `close` method either.

grosser:
Best open an issue in faraday for the close, might be an easy fix.
I touched a bit of code in there, but these edge-cases are always tricky :(

grosser:
this can be closed, master uses faraday already ?

cben:
No, master uses faraday where we previously used `rest-client`, but watching still uses `http`.
It'd great to converge them, although not strictly a release blocker.

cben:
Turns out `on_data` response streaming is presently [supported by only 2 faraday adapters](https://github.com/lostisland/awesome-faraday/#adapters) — the basic Net::HTTP and httpx.  
But it would be sad to generally support only these 2.

- Could we allow watches to use to use a different adapter from other request types [maybe hardcoded Net::HTTP]?
- Should we look for non-faraday solution to watches?

DocX:
> Turns out on_data response streaming is presently [supported by only 2 faraday adapters](https://github.com/lostisland/awesome-faraday/#adapters) — the basic Net::HTTP and httpx.
But it would be sad to generally support only these 2.

What prevents kubeclient to use the specific adapter for its own use? Even if the user's app uses faraday with different adapter, it does not have to interfere with what kubeclient uses. The adapter can be configured everytime we create the faraday instance with `Faraday.new { |f| f.adapter = ... }`.

We already use different client if we don't use faraday in the watch code anyway, so the argument that it would not be aligned with the user's code is IMHO gone.

Whatever kubeclient uses is just internal to kubeclient. 

I guess the only slight disadvantage is that if user uses adapter A, adding kubeclient would add adapter B to their gems?

cben:
Nothing prevents, your analysis is right.

PR #495, opened 2021-03-06
remove vcr tests because hooking into webmock breaks multithreaded tests

grosser:
https://github.com/abonas/kubeclient/pull/494 breaks when VCR is enabled .. somehow it ends up executing the stubs on all threads

can reproduce by adding to test/test_helper.rb
```
require 'vcr'
VCR.configure do |c|
  c.cassette_library_dir = 'test/cassettes'
  c.hook_into(:webmock)
end
```
and running `ruby test/test_informer.rb`

the guestbook test itself seems rather brittle / not reproducible, so I'm thinking best get rid of it instead of pouring more work into it


cben:
I guess you're right.

I was recently thinking to automate some version of k8s that's easy to start, e.g. k3s, to be able to reproducibly re-record it the VCR.  But if we make something like that, we don't need VCR, we could get real end-to-end testing.  As it is, the VCR test is not worth much.

cben:
Don't know yet what's going on with truffleruby.  Added it to #435 checklist so we don't forget.  This PR is obviously safe, merging.

PR #498, opened 2021-04-15
Raise ResourceAlreadyExistsError when resource being created already exists

cjoudrey:
👋 Hello! 😄 

### Context

We've been working on automation to setup new Kubernetes clusters and part the automation involves bootstrapping namespaces. Since this automation can sometimes fail, we try to write our tasks in a way that they are idempotent and thus retryable.

Today, if you attempt to create the same namespace twice using `kubeclient` you'll get a generic `HttpError`:

```ruby
irb(main):005:0> client.create_namespace(Kubeclient::Resource.new(metadata: { name: 'default' }))
Kubeclient::HttpError (HTTP status code 409, namespaces "default" already exists for POST https://xxx.xxx.xxx.xxx/api/v1/namespaces)
```

It's possible to extract the `reason` from this `HttpError` through the `response`, but it leads to inelegant code:

```ruby
rescue Kubeclient::HttpError => e
  response = JSON.parse(e.response.body)
  return if response["reason"] == "AlreadyExists"

  raise
end
```

### This PR

This pull request exposes a new error class (`ResourceAlreadyExistsError`) which does this heavy-lifting and turns the above code into:

```ruby
rescue Kubeclient::ResourceAlreadyExistsError
  # Resource already exists, nothing to do here
  nil
rescue Kubeclient::HttpError => e
  # ...
end
```

The change should be backwards compatible since `ResourceAlreadyExistsError` inherits from `HttpError`.

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -16,6 +16,7 @@ require_relative 'kubeclient/missing_kind_compatibility'
 require_relative 'kubeclient/oidc_auth_provider'
 require_relative 'kubeclient/resource'
 require_relative 'kubeclient/resource_not_found_error'
+require_relative 'kubeclient/resource_already_exists_error'
 require_relative 'kubeclient/version'
 require_relative 'kubeclient/watch_stream'
 
@@ -159,7 +160,15 @@ module Kubeclient
     rescue Faraday::Error => e
       err_message = build_http_error_message(e)
       response_code = e.response ? (e.response[:status] || e.response&.env&.status) : nil
-      error_klass = (response_code == 404 ? ResourceNotFoundError : HttpError)
+
+      error_klass = if response_code == 404
+                      ResourceNotFoundError
+                    elsif response_code == 409 && build_http_conflict_reason(e) == 'AlreadyExists'
+                      ResourceAlreadyExistsError
+                    else
+                      HttpError
+                    end
+
       raise error_klass.new(response_code, err_message, e.response)
     end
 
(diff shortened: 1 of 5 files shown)


lib/kubeclient.rb
```diff
@@ -589,6 +598,16 @@ def api
                                          security_context_constraints]
     }.freeze
 
+    def build_http_conflict_reason(e)
```
cjoudrey, line 601: It's unfortunate that we're parsing the JSON response twice (once in `build_http_error_message` and once here during a 409) but I noticed `build_http_error_message` was a public method so I wanted to avoid making a breaking change. I don't have much context on why it's public, but perhaps in the future it should also be `private`?

cben: My stance for quite a while has been that if it's not documented in the README, it's internal and we're free to break it, even if it's not `private` on the ruby level.
So feel free to change this, and/or make it `private`.

cben: Speaking of which, please document `ResourceAlreadyExistsError` in README.  And if you don't mind, `ResourceNotFoundError` which I see we forgot.

Double parsing is suboptimal but not deal-breaker, I'm OK with merging as-is if you prefer.  Otherwise LGTM :100:, thanks for this PR!

cjoudrey: > Speaking of which, please document `ResourceAlreadyExistsError` in README. And if you don't mind, `ResourceNotFoundError` which I see we forgot.

Ohhh yeah, good call!

Do you have a preference on where to put these?

Perhaps a section `### Errors` within `## Usage`? Perhaps after `### Timeouts`?

cjoudrey:
I added some documentation about errors to `README.md`:

Screen Shot 2021-04-20 at 6 38 07 PM

@cben Let me know if you want me to change anything in the wording or its location in the `README.md`.

Thanks! 😄

cben:
Perfect.

PR #502, opened 2021-05-11
Omit test/ dir from built .gem (v4.y branch)

### LINKED ISSUE #501: Delete kubeclient gem default certificates
Hi,
I have installed kubeclient gem in td-agent. This gem has installed default certificates in test/config/ folder. From https://github.com/abonas/kubeclient/issues/499 I got to know it is used for running unit tests and also the certificates are expired. Having default certificates bundled in a gem is a security risk. Can these certificates be deleted while bundling the gem so that the default certificates are not bundled with kubeclient gem?

cben:
Closes #501.  `test_files` is deprecated.  It was causing all tests to be shipped in the .gem file, which is useless bloat (was questionable tradeoff before when rubygems supported running tests for a downloaded gem, anyway no longer supported): https://github.com/rubygems/guides/issues/90

|   | before (4.9.1) | after |
|---|-------------|-----|
| Compressed | 78KB | 33KB |
| Unpacked | 757KB | 184KB |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 Kubeclient release versioning follows [SemVer](https://semver.org/).
 
+## UNRELEASED
+
+### Removed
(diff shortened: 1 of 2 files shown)


cben:
to be forward-ported to master too.

Issue #483, opened 2020-11-23
test TLS certs expired


pravi:
```
TestResourceQuota#test_get_from_json_v1 = 0.01 s = .
KubeclientConfigTest#test_user_token = /usr/lib/ruby/vendor_ruby/kubeclient/config.rb:33: Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
0.00 s = .
KubeclientConfigTest#test_external_nopath_absolute = 0.00 s = .
KubeclientConfigTest#test_gcp_default_auth = /tmp/autopkgtest-lxc.0kqlxqvn/downtmp/build.PLE/src/test/test_config.rb:128: Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
0.00 s = .
KubeclientConfigTest#test_gcp_default_auth_renew = /tmp/autopkgtest-lxc.0kqlxqvn/downtmp/build.PLE/src/test/test_config.rb:137: Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
0.00 s = .
KubeclientConfigTest#test_user_exec = /usr/lib/ruby/vendor_ruby/kubeclient/config.rb:33: Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
0.01 s = .
KubeclientConfigTest#test_user_password = /usr/lib/ruby/vendor_ruby/kubeclient/config.rb:33: Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
0.00 s = .
KubeclientConfigTest#test_user_exec_nopath = 0.00 s = .
KubeclientConfigTest#test_external = /usr/lib/ruby/vendor_ruby/kubeclient/config.rb:33: Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.
0.00 s = F
[...]

PR #504, opened 2021-05-12
v4.y: Disable tests using expired TLS certs

cben:
Backporting, was disabled on master as part of commit 03c5bfbbde4fd5678a6c6c98fbc34a953eb41dfb.

https://github.com/abonas/kubeclient/issues/483
Instructions how to renew the certs no longer work (they relied on openshift 3.x).  Anyway I'm not certain
what this test actually tests (other than the openssl lib).
One thing that _would_ be valuable to test is https://github.com/abonas/kubeclient/pull/461


cben:
:man_facepalming: I merged this by mistake into master (which was a no-op, zero diff) instead of v4.y.

PR #507, opened 2021-05-19
Allow a region when getting a signer for Aws::Sts

agrare:
While STS does work globally it seems recommended to pass a region to reduce latency and increase redundancy.

https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html

### DIFF
--- a/lib/kubeclient/aws_eks_credentials.rb
+++ b/lib/kubeclient/aws_eks_credentials.rb
@@ -7,7 +7,7 @@ module Kubeclient
     end
 
     class << self
-      def token(credentials, eks_cluster)
+      def token(credentials, eks_cluster, region: 'us-east-1')
         begin
           require 'aws-sigv4'
           require 'base64'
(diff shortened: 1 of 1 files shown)


lib/kubeclient/aws_eks_credentials.rb
```diff
@@ -20,10 +20,9 @@ def token(credentials, eks_cluster)
         end
         # https://github.com/aws/aws-sdk-ruby/pull/1848
```
agrare, line 21: In this PR the author does end up passing in the region from the sts client dynamically: https://github.com/aws/aws-sdk-ruby/pull/1848/files#diff-f99f43d6d557ede5f15cce411662785f51d791a9c26317db749a4a6bb249e03eR54

agrare:
Hey @cben any opinions about this?

cben:
Oops, i missed this entirely. LGTM.

Issue #512, opened 2021-08-11
Watch pods times out/stops recognizes new notices after ~30 mins


khchau7:
client.watch_pods times out/stops recognizing new notices if there is no watch event that occurs for about 30 minutes even with timeouts set (open=60, read=nil) when creating the kubeclient. 
The watch_pods is wrapped in an infinite loop, so if it simply exits or raises an exception, there is a mechanism in place to sleep and then restart the watch. However, we currently don't break out of watch_pods and instead time out and not act on any new notices received.

Might need to set the tcp keep alive (keep_alive_timeout) in the http_options in kubeclient.rb to avoid this. Need to replicate the behavior provided in client-go as specified here: [https://github.com/yangl900/knet](https://github.com/yangl900/knet)
"The k8s client-go by default turns on TCP Keepalive, and the client side will send an ACK packet to API server every 30s. With this, even though the SLB default timeout is 4 minutes, the TCP connection will never be idle and so will never be reset."

Any other ideas as to why this could be occurring?

cben:
https://github.com/yangl900/knet is very interesting, I need to absorb the info there...  
I had "folk knowledge" that whatever you do, apiserver eventually closes inactive connections, but can't point to a source.  Anyway more understanding & tuning are good.

Note that keeping an open connection for many minutes with no activity is likely to cause a "version too old" if it does get disconnected.  Consider also adding kubeclient support to opt-in to watch bookmarks which reduce this issue:
https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks
https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/956-watch-bookmark#motivation

chrisjohnson:
We're seeing this as well. When the actual connection gets closed, the watch command exits with an error (as documented) which we catch and use to restart the listener. But there is something that happens where the listener stops getting notices, but doesn't actually exit, leaving the process more or less stuck. We're going to add a timer-based kill on our side as a workaround but it seems like kubeclient has a condition it needs to account for.

PR #513, opened 2021-08-12
added support for keep alive timeouts #512

khchau7:
See #512: Fix in response to watch pods timing out and not recognizing any new notices after about 30 minutes of no new watch events.

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -53,6 +53,7 @@ module Kubeclient
 
     DEFAULT_HTTP_PROXY_URI = nil
     DEFAULT_HTTP_MAX_REDIRECTS = 10
+    DEFAULT_KEEP_ALIVE_TIMEOUT = 60
 
     SEARCH_ARGUMENTS = {
       'labelSelector'   => :label_selector,
@@ -73,6 +74,7 @@ module Kubeclient
     attr_reader :auth_options
     attr_reader :http_proxy_uri
     attr_reader :http_max_redirects
+    attr_reader :keep_alive_timeout
     attr_reader :headers
     attr_reader :discovered
 
@@ -102,6 +104,7 @@ module Kubeclient
       timeouts: DEFAULT_TIMEOUTS,
       http_proxy_uri: DEFAULT_HTTP_PROXY_URI,
       http_max_redirects: DEFAULT_HTTP_MAX_REDIRECTS,
+      keep_alive_timeout: DEFAULT_KEEP_ALIVE_TIMEOUT,
       as: :ros
     )
       validate_auth_options(auth_options)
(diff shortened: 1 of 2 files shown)


lib/kubeclient/watch_stream.rb
```diff
@@ -11,6 +11,7 @@ def initialize(uri, http_options, formatter:)
         @http_client = nil
         @http_options = http_options
         @http_options[:http_max_redirects] ||= Kubeclient::Client::DEFAULT_HTTP_MAX_REDIRECTS
+        @http_options[:keep_alive_timeout] = 60
```
ganga1980, line 14: this should be like this, this way you can set the value if its being set otherwise it will be default.

@http_options[:keep_alive_timeout] || = Kubeclient::Client::DEFAULT_KEEP_ALIVE_TIMEOUT

ganga1980:
Hi, @cben, can you please review this PR and helping on merging this PR as @khchau7  blocked on this for her project?

lib/kubeclient/watch_stream.rb
ganga1980: keep_alive_timeout only set if its passed in otherwise this shouldnt set to keep no changes to existing behavior.

lib/kubeclient.rb
```diff
@@ -53,6 +53,7 @@ class Client
 
     DEFAULT_HTTP_PROXY_URI = nil
     DEFAULT_HTTP_MAX_REDIRECTS = 10
+    DEFAULT_KEEP_ALIVE_TIMEOUT = 60
```
cben, line 56: The default of `http` gem seems to be 5 seconds: https://github.com/httprb/http/blob/v4.4.1/lib/http/options.rb#L63

I open to setting it higher by default if that's good for k8s watch connections.
This effectively controls how long we keep the watch open when server has no updates, right?
Do you know what's the default in Go or other k8s clients?

OTOH if we're not sure what's best value, we can set it 5 here to keep existing behavior, this PR will let users experiment...

cben:
Is this really controlling what #512 describes?
https://github.com/httprb/http/blob/v4.4.1/lib/http/connection.rb#L184-L188 
https://github.com/httprb/http/blob/v4.4.1/lib/http/client.rb#L113-L116
seems it affects Ruby-level logic closing connection on inactivity, not TCP-level keepalive?

Generally, I'm not against exposing more options :+1:.  We can merge this first and let people experiment with optimal settings.  
Note however that sooner or later, we'd like to replace `http` with `faraday`: #488 — does this setting have a parallel in faraday?  
If raising the timeout helps but the problem is specific to `http` gem, then perhaps we should just set it higher but not expose as new knob.

If you believe this is correct approach, please add a mention of the new option in README and add a CHANGELOG entry (in "Unreleased" section, which is currently horribly incomplete).

lib/kubeclient.rb
```diff
@@ -287,6 +287,7 @@ def define_entity_methods
           # This method used to take resource_version as a param, so
           # this conversion is to keep backwards compatibility
           options = { resource_version: options } unless options.is_a?(Hash)
+          options[timeoutSeconds] = 380 # 6 mins 20 seconds
```
cben, line 290: Did you mean `:timeoutSeconds` ?
This will override any passed-in timeout.  I'm assuming you're just experimenting for now, but this can't be merged.

PR #516, opened 2021-09-19
allow allow_watch_bookmarks param

grosser:
needed for [watch bookmarks](https://kubernetes.io/docs/reference/using-api/api-concepts/#watch-bookmarks)
using them to get a constant heartbeat from the watch so I can stop it if it looks dead using [heartbleed](https://github.com/grosser/heartbleed)
heartbeat comes in every 1 minute after a random initial delay (saw up to 7 minutes) ... api docs say there is no guarantee

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,7 +495,7 @@ You can specify multiple labels (that option will return entities which have bot
 pods = client.get_pods(label_selector: 'name=redis-master,app=redis')
 ```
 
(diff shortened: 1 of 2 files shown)


cben:
LGTM.  Can you document this in README?  If passed, this extends the possible values for `.type` (https://github.com/abonas/kubeclient#watch--receive-entities-updates).

one CI job timed out after 5 hours (?!), but this change alone can't break anything.  re-running...

PR #520, opened 2021-12-10
make missing @kcfg_path self-documenting

grosser:
@cben

### DIFF
--- a/lib/kubeclient/config.rb
+++ b/lib/kubeclient/config.rb
@@ -85,21 +85,21 @@ module Kubeclient
 
     private
 
-    def allow_external_lookups?
-      @kcfg_path != nil
-    end
-
     def ext_file_path(path)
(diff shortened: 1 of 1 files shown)


grosser:
also I'd propose to change readme to only reference `.read` 🤷

grosser:
also I'd propose to change readme to only reference `.read` 🤷

cben:
> also I'd propose to change readme to only reference .read 🤷

I'm fine with emphasizing .read, but I believe a library should offer the _ability_ do work on passed-in data with no connection to the file system.

But note the exception you're raising only refers to .new - thats a fine place to mention only .read (using constructor directly but with file path is really niche...)

cben:
Good to merge as-is if you want.

grosser:
yeah just merge :)

PR #522, opened 2021-12-11
expose kubeconfig hash

mindovermiles262:
Sometimes you need to access the raw kube_config hash. This PR adds a method to do so

### DIFF
--- a/lib/kubeclient/config.rb
+++ b/lib/kubeclient/config.rb
@@ -83,6 +83,10 @@ module Kubeclient
       Context.new(cluster['server'], @kcfg['apiVersion'], ssl_options, auth_options, namespace)
     end
 
+    def kube_config
+      @kcfg
+    end
+
     private
 
     def allow_external_lookups?


cben:
Thanks.  Generally makes sense, I don't think the instance var was deliberately private :+1:

Note that this exposes the original data from kubeconfig file, before all the work done by `.context()` to resolve data:
- referenced external files will not be inlined (e.g. you may see `certificate-authority` file name vs. inline `certificate-authority-data`).
- active credential sources `ExecCredentials` / `GCPAuthProvider` / `OIDCAuthProvider` will not run.
- the data will be static.  This matters because in future we'll want to support auth renewal #393, which besides expiration on above active sources requires re-reading the kubeconfig file (or files referenced from it) when updated.

If you have use case for it with these caveats, LGTM, but please add documentation in README.

Is `.kube_config` clear name for the as-is parsed yaml?  
What do you think of something like `.to_hash`, or maybe `.parsed` / `.as_parsed` to be similar to Client's `as: :parsed` mode?  (not sure these are better, just throwing out ideas...)

cben:
@mindovermiles262 The question still stands — do you think this will be useful with the caveats listed above?  

* One change since then: we've implemented `bearer_token_file` auto-reloading in `Client` (but not an inline `bearer_token`).  
  IIRC, `bearer_token_file` reloading happens just before making a request, so you'll not see it reflected in `Config`'s hash.  You'll see a static file path there, because as I said this returns the parsed yaml as-is, before resolving external file references.

Or should we close this?  What's is your use case for accessing the raw config?

mindovermiles262:
I do not remember why I needed the raw kube config. You're OK to do as you wish with this PR.

Issue #521, opened 2021-12-10
support impersonation


grosser:
set these headers from reading kubectl config or from user-passed config:
```
  "Impersonate-User": "<as>",
  "Impersonate-Group": "<as-groups.join>",
  "Impersonate-Extra-Reason": "<as-user-extra.reason.join>"
```

as-groups and as-user-extra.reason being arrays

did that locally and it works, it's just clumsy so having the config parser automatically do that and the `auth_options` support it would be nice

I can make a PR if this sounds sane :)

cben:
First time I'm hearing about it, so don't have much to say.  Go ahead :)

ref: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation
Is there documentation on where it lives in kubeconfig?  All I see in that doc are kubectl has `--as` and `--as-group` flags.

DocX:
Should be closed by #600

PR #524, opened 2021-12-12
support impersonation via auth_options and in kubectl config

grosser:
fixes https://github.com/abonas/kubeclient/issues/521

TODO: confirm it actually works (can check Monday)

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,6 +430,24 @@ You can read it as follows:
 puts config.context.namespace
 ```
 
+### Impersonation
+
+Impersonation is supported when loading a kubectl config and via the Ruby API, for example:
+
+```ruby
+client = Kubeclient::Client.new(
+  context.api_endpoint, 'v1',
+  auth_options: {
+    as: "admin",
+    as_groups: ["system:masters"],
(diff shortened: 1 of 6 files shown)


cben:
Code looks good, though I don't know much about the feature.
Have you confirmed it actually works?

grosser:
I'm blocked by the ssl being broken in master :(
https://github.com/abonas/kubeclient/issues/525
in theory it should work since that's what I've been doing manually before (setting the headers)
if the ssl issue is hard to fix I can try rebasing this on 4.9 to do an end-to-end test ...

cben:
Just a reminder SSL issues (that I'm aware of) have been fixed.

DocX:
Hi @grosser I've updated your work to latest master and tested it works. I've run into the missing impersonation when working on our internal project. Getting the update makes it work.

I've made new PR here: https://github.com/ManageIQ/kubeclient/pull/600 but feel free to update this PR if you think you would loose the credit :)

grosser:
the less my name is in git blame the better :D

PR #526, opened 2021-12-13
do not fail the build on experimental truffleruby-head

grosser:
fails the build and cancels other stages

Screen Shot 2021-12-12 at 5 50 38 PM


.github/workflows/actions.yml
```diff
@@ -11,15 +11,26 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
```
cben, line 14: Neat!

Issue #527, opened 2022-01-05
4.9.2 and master do not work with AWS GovCloud due to regional hardcoding


snorlaX-sleeps:
The current release of Kubeclient (4.9.2) does not work with AWS GovCloud, resulting in an error like the following:
```
Traceback (most recent call last):
----
	14: from ~/.rvm/gems/ruby-2.6.0/gems/kubeclient-4.9.2/lib/kubeclient.rb:139:in `method_missing'
	13: from ~/.rvm/gems/ruby-2.6.0/gems/kubeclient-4.9.2/lib/kubeclient.rb:187:in `discover'
	12: from ~/.rvm/gems/ruby-2.6.0/gems/kubeclient-4.9.2/lib/kubeclient.rb:661:in `load_entities'
	11: from ~/.rvm/gems/ruby-2.6.0/gems/kubeclient-4.9.2/lib/kubeclient.rb:674:in `fetch_entities'
	10: from ~/.rvm/gems/ruby-2.6.0/gems/kubeclient-4.9.2/lib/kubeclient.rb:160:in `handle_exception'
	 9: from ~/.rvm/gems/ruby-2.6.0/gems/kubeclient-4.9.2/lib/kubeclient.rb:674:in `block in fetch_entities'
	 8: from ~/.rvm/gems/ruby-2.6.0/gems/faraday-1.8.0/lib/faraday/connection.rb:200:in `get'
	 7: from ~/.rvm/gems/ruby-2.6.0/gems/faraday-1.8.0/lib/faraday/connection.rb:516:in `run_request'
	 6: from ~/.rvm/gems/ruby-2.6.0/gems/faraday-1.8.0/lib/faraday/rack_builder.rb:154:in `build_response'
	 5: from ~/.rvm/gems/ruby-2.6.0/gems/faraday_middleware-1.2.0/lib/faraday_middleware/response/follow_redirects.rb:67:in `call'
	 4: from ~/.rvm/gems/ruby-2.6.0/gems/faraday_middleware-1.2.0/lib/faraday_middleware/response/follow_redirects.rb:79:in `perform_with_redirection'
	 3: from ~/.rvm/gems/ruby-2.6.0/gems/faraday-1.8.0/lib/faraday/middleware.rb:18:in `call'
	 2: from ~/.rvm/gems/ruby-2.6.0/gems/faraday-1.8.0/lib/faraday/response.rb:59:in `on_complete'
[...]

PR #528, opened 2022-01-05
Update the AWS STS endpoint to be regional as the method is now regional

snorlaX-sleeps:
Required to support AWS GovCloud. 
Currently the `token` method accepts a region to create the signature for logging into an AWS Cluster, which should make it "quicker" on a regional basis.
However the AWS GovCloud partition does not / is not supported by the URL `https://sts.amazonaws.com` as AWS treats it as a totally separate entity.

This change updates the STS endpoint to match the region provided with the method.

I would be happy to change the PR to update the `token` method (even the parameters) to pass a `gov_cloud` named parameter (similar to the region one), then the original `https://sts.amazonaws.com` could be kept for speed / efficiency for end-users that do not set region and use normal AWS regions - then only set the region in the URL when using the govCloud partition?

[List of AWS STS endpoint URLs](https://docs.aws.amazon.com/general/latest/gr/sts.html)

Fixes https://github.com/abonas/kubeclient/issues/527
Follows on from https://github.com/abonas/kubeclient/pull/507

### DIFF
--- a/lib/kubeclient/aws_eks_credentials.rb
+++ b/lib/kubeclient/aws_eks_credentials.rb
@@ -29,7 +29,7 @@ module Kubeclient
         # https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Sigv4/Signer.html#presign_url-instance_method
         presigned_url_string = signer.presign_url(
           http_method: 'GET',
-          url: 'https://sts.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15',
+          url: "https://sts.#{region}.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15",
           body: '',
           credentials: credentials,
           expires_in: 60,


cben:
from the doc you linked:
> AWS recommends using Regional STS endpoints to reduce latency, build in redundancy, and increase session token validity. Most Regional endpoints are active by default, but you must manually enable endpoints for some Regions, such as Asia Pacific (Hong Kong). You can deactivate STS endpoints for any Regions that are enabled by default if you do not intend to use those Regions.

So sounds like defaulting to per-region endpoint is best practice, not clear that a `gov_cloud: false` param should imply using  `https://sts.amazonaws.com`.  If you think this is useful to control, consider an explicit `use_global_sts` param?
(I never used this, deferring to you on what's actually best)

Meanwhile this LGTM as-is, merging.  (and sorry for my non-responsiveness :snail:)

cben:
close-cycling to rerun CI

snorlaX-sleeps:
Hey @cben - thanks for getting back to me (I had a local copy of the gem so this was not an issue)

> not clear that a gov_cloud: false param should imply using https://sts.amazonaws.com. If you think this is useful to control, consider an explicit use_global_sts param?

The `https://sts.amazonaws.com` endpoint *is regional*, that is just a common endpoint for any commercial (non-gov_cloud or certain other regions) to hit without needing to be region specific.
I think they are equivalent, I just didn't want to change pre-written functionality too much if you preferred.
There is a hard partition between commercial AWS accounts and gov_cloud accounts, so they have to use different services, hence the issue

agrare:
Great fix @snorlaX-sleeps thank you

Issue #530, opened 2022-01-19
bearer_token needs to be reloaded every 1h to avoid expired service account tokens


grosser:
as per https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md since 1.22 (or with BoundServiceAccountTokenVolume=true) tokens will expire every 1h and need to be re-loaded from disk

options:
- add a reload_bearer_auth_token method for users to call
- add an internal Thread the reloads it (... ideally on by default, but tricky)
- check age every time a request is made and reload as needed
- reload on every request (in `def rest_client` and `def http_options`) since it's pretty cheap and usually an in-memory file

PR #532, opened 2022-01-20
refresh bearer token from file to avoid stale tokens when using mount…

grosser:
…ed service account token

fixes https://github.com/abonas/kubeclient/issues/530

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -122,10 +122,7 @@ module Kubeclient
       @http_max_redirects = http_max_redirects
       @as = as
 
-      if (file = auth_options[:bearer_token_file]) # rubocop:disable Style/GuardClause
-        validate_bearer_token_file(file)
-        @auth_options[:bearer_token] = File.read(file).chomp
-      end
+      validate_bearer_token_file
     end
 
     def configure_faraday(&block)
@@ -368,6 +365,7 @@ module Kubeclient
     end
 
     def faraday_client
+      refresh_bearer_token_from_file
       @faraday_client ||= create_faraday_client
     end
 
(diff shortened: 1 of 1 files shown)


lib/kubeclient.rb
```diff
@@ -358,6 +357,12 @@ def create_faraday_client(url = nil)
         if @auth_options[:username] && @auth_options[:password]
           connection.basic_auth(@auth_options[:username], @auth_options[:password])
         end
+
+        refresh_bearer_token_from_file
+        if auth = @headers[:Authorization]
+          req.headers['Authorization'] = auth
```
grosser, line 363: this is kinda sus ... did not see where this was supposed to be coming from before

lib/kubeclient.rb
```diff
@@ -695,11 +698,9 @@ def validate_auth_options(opts)
...
 
     def validate_bearer_token_file
-      msg = "Token file #{@auth_options[:bearer_token_file]} does not exist"
-      raise ArgumentError, msg unless File.file?(@auth_options[:bearer_token_file])
-
-      msg = "Cannot read token file #{@auth_options[:bearer_token_file]}"
-      raise ArgumentError, msg unless File.readable?(@auth_options[:bearer_token_file])
+      file = @auth_options[:bearer_token_file]
+      raise ArgumentError, "Token file #{file} does not exist" unless File.file?(file)
+      raise ArgumentError, "Token file #{file} cannot be read" unless File.readable?(file)
```
grosser, line 703: just cleanup

grosser:
with rebase this is now super clean :D

kjetijor:
`@client = Kubeclient::Client.new(....)` going straight to `@client.api_valid?` seems to not work with this, it'll end up missing any call to refresh_bearer_token_from_file.

grosser:
fixed

cben:
Code is neat :+1:

About performance: you mentioned several options on #530 and this implements the last one:

* reload on every request (in def rest_client and def http_options) since it's pretty cheap and usually an in-memory file

So this will add several syscalls on every request.  But only when config uses a token file, which is probably mainly used in-cluster.
I don't know how to judge if that's Good Enough, but OK let's start with it and optimize later if needed.

PR #537, opened 2022-02-15
bearer_token_file: load inline

kjetijor:
simplify loading the bearer token from file, to just reading it inline. The previous approach with refreshing the token from file from various points, seems to have missed something - we were still flagged by the audit logging on our k8s clusters.

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -706,10 +706,21 @@ module Kubeclient
     end
 
     def http_options(uri)
+      bearer_token = nil
+      if @auth_options[:bearer_token_file]
+        bearer_token_file = @auth_options[:bearer_token_file]
+        if File.file?(bearer_token_file) and File.readable?(bearer_token_file)
+          token = File.read(bearer_token_file).chomp
+          bearer_token = "Bearer #{token}"
+        end
+      elsif @auth_options[:bearer_token]
+        bearer_token = "Bearer #{@auth_options[:bearer_token]}"
+      end
 
       options = {
         basic_auth_user: @auth_options[:username],
         basic_auth_password: @auth_options[:password],
+        authorization: bearer_token,
         headers: @headers,
         http_proxy_uri: @http_proxy_uri,
         http_max_redirects: http_max_redirects
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -354,7 +354,9 @@ module Kubeclient
         if @auth_options[:username]
           connection.request(:basic_auth, @auth_options[:username], @auth_options[:password])
         elsif @auth_options[:bearer_token_file]
-          connection.request(:authorization, 'Bearer', -> { File.read(@auth_options[:bearer_token_file]).chomp })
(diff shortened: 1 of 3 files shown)


lib/kubeclient/version.rb
```diff
@@ -2,5 +2,5 @@
 
 # Kubernetes REST-API Client
 module Kubeclient
-  VERSION = '4.9.2'
+  VERSION = '4.9.3'
```
cben, line 5: please drop the version bump from here.

I agree that token refresh feels worthy of a 4.10.z release :+1:, but current master has backward-incompatible changes not releasable as 4.y, and switch from rest-client to faraday makes this non-trivial to backport.
Need to see what we can do about it :thinking:, but meanwhile let's get it working on master...

@grosser could you review the rest?

lib/kubeclient.rb
```diff
@@ -352,6 +352,8 @@ def create_faraday_client
       Faraday.new(url, options) do |connection|
         if @auth_options[:username]
           connection.request(:basic_auth, @auth_options[:username], @auth_options[:password])
+        elsif @auth_options[:bearer_token_file]
+          connection.request(:authorization, 'Bearer', -> { File.read(@auth_options[:bearer_token_file]).chomp })
```
cben, line 356: Ooh, neat, this make Faraday execute the proc on every request:

https://www.rubydoc.info/gems/faraday/Faraday/Request/Authorization,
https://github.com/lostisland/faraday/blob/main/lib/faraday/request/authorization.rb
> params (Array<String, Proc, call>) — parameters to build the Authorization header. If the type is `:basic`, then these can be a login and password pair. Otherwise, a single value is expected that will be appended after the type. This value can be a proc or an object responding to `.call`, in which case it will be invoked on each request.

lib/kubeclient.rb
```diff
@@ -704,7 +705,6 @@ def return_or_yield_to_watcher(watcher, &block)
     end
 
     def http_options(uri)
-      refresh_bearer_token_from_file
```
cben, line 707: What about `watch_foo` connections?
These don't use faraday yet (pending #488) but `http` gem.  
IIUC #532 ensured watch requests get a fresh token by this line (the name `http_options` literally means "options for `http` gem) and I think dropping this means they won't get fresh tokens, or perhaps won't get a bearer token _at all_?

grosser (approved):
nice approach!
... yeah careful with watch
the audit log should show which request was stale too so you can see if the problem was watch or regular calls

lib/kubeclient.rb
kjetijor: @cben - I'll have a look at watch_*, and unbreak that if I did break it.

kjetijor: watch_foo was broken, both here and on master. It should be less broken now, as we ram through the authorization header from either :bearer_token or :bearer_token_file

grosser:
FYI this fixes https://github.com/ManageIQ/kubeclient/issues/525 too :D

cben:
LGTM

PR #540, opened 2022-02-22
verify_ssl should be converted to a bool for faraday

russell:
in faraday they support `verify` it should be a bool, it also supports
verify_mode which is defined here
https://ruby-doc.org/stdlib-2.5.1/libdoc/openssl/rdoc/OpenSSL/SSL.html

currently to disable certificate verification explicitly we would pass, `OpenSSL::SSL::VERIFY_NONE == 0` so this results in verification being turned on.
```
ssl_options = { verify_ssl: OpenSSL::SSL::VERIFY_NONE }
client = Kubeclient::Client.new(
  'https://localhost:8443/api', 'v1', ssl_options: ssl_options
)
```

Faraday configuration is described here https://github.com/lostisland/faraday/blob/21ac595919b0372b41386ae00640e16f73a20d31/lib/faraday/options/ssl_options.rb#L7

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -345,7 +345,8 @@ module Kubeclient
           cert_store: @ssl_options[:cert_store],
           client_cert: @ssl_options[:client_cert],
           client_key: @ssl_options[:client_key],
-          verify: @ssl_options[:verify_ssl]
+          verify: @ssl_options[:verify_ssl] != OpenSSL::SSL::VERIFY_NONE,
+          verify_mode: @ssl_options[:verify_ssl]
         }
       }
 


russell:
the failures on this branch i got when running locally with 3.1.0, i have no idea, i've not used minitest before, but the error is so strange, that i'm not sure where to start debugging it

lib/kubeclient.rb
```diff
@@ -345,7 +345,8 @@ def create_faraday_client
           cert_store: @ssl_options[:cert_store],
           client_cert: @ssl_options[:client_cert],
           client_key: @ssl_options[:client_key],
-          verify: @ssl_options[:verify_ssl]
+          verify: @ssl_options[:verify_ssl] != OpenSSL::SSL::VERIFY_NONE,
+          verify_mode: @ssl_options[:verify_ssl]
```
cben, line 349: Oh wow good find :clap:

This should fix the issue as long as users pass integers 0=VERIFY_NONE or 1=VERIFY_PEER, which has been the convention kubeclient expects so far :heavy_check_mark:  

But this confusion makes me wonder what happens if a user of kubeclient gets confused about what _they_ should pass.  E.g. if they pass a bool, both false and true will result in `verify: true` because true != 0, and `verify_mode` will pass an unexpected bool into faraday (whose effect may depend on adapter)...

So I'm thinking should we also add validation in the constructor that `ssl_options[:verify_ssl]` is one of VERIFY_NONE/VERIFY_PEER intergers?
WDYT? (but not in this PR, I may want to backport that one validation)

cben:
ruby-head and truffleruby-head are constantly failing, not considered blockers to merge.

LGTM, I'll merge soon, I just want to test it locally...

cben:
opened #543 to track ruby 3.1 support.

lib/kubeclient.rb
russell: that is a good question, further down in the file there is another use of the same verify_ssl value, i think that the most sensible thing would be to validate that it's `OpenSSL::SSL::VERIFY_NONE || OpenSSL::SSL::VERIFY_PEER` we could try and convert `true/false` but maybe it's just better to throw an error and to instead require the correct SSL values, because they could change 🤷 

other usage: https://github.com/ManageIQ/kubeclient/blob/master/lib/kubeclient.rb#L726

PR #541, opened 2022-02-22
add delete_collection api

russell:
Delete collection is like getting a list, except it triggers a delete of the
resources. It returns a list of all resources that match the call.

I based this of the get entities like get_pods because it's roughly
the same API except as a side effect it deletes all the resources.

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -62,6 +62,20 @@ module Kubeclient
       'continue'        => :continue
     }.freeze
 
+    DELETE_COLLECTION_ARGUMENTS = {
+      'continue'             => :continue,
+      'dryRun'               => :dry_run,
+      'fieldSelector'        => :field_selector,
+      'gracePeriodSeconds'   => :grace_period_seconds,
+      'labelSelector'        => :label_selector,
+      'limit'                => :limit,
+      'orphanDependents'     => :orphan_dependents,
+      'propagationPolicy'    => :propagation_policy,
+      'resourceVersion'      => :resource_version,
+      'resourceVersionMatch' => :resource_version_match,
+      'timeoutSeconds'       => :timeout_seconds
+    }.freeze
+
     WATCH_ARGUMENTS = {
       'labelSelector'   => :label_selector,
       'fieldSelector'   => :field_selector,
@@ -294,6 +308,12 @@ module Kubeclient
           delete_entity(entity.resource_name, name, namespace, **opts)
         end
 
+        # delete all entities of a type e.g. delete_pods, etc.
+        define_singleton_method("delete_#{entity.method_names[1]}") \
+        do |options = {}|
+          delete_collection(entity.entity_type, entity.resource_name, options)
+        end
+
         define_singleton_method("create_#{entity.method_names[0]}") do |entity_config|
           create_entity(entity.entity_type, entity.resource_name, entity_config)
         end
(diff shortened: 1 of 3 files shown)


grosser (approved):

lib/kubeclient.rb
```diff
@@ -449,6 +469,32 @@ def delete_entity(resource_name, name, namespace = nil, delete_options: {})
...
+    # Accepts the following options:
+    #   :namespace (string) - the namespace of the entity.
+    #   :label_selector (string) - a selector to restrict the list of returned objects by labels.
+    #   :field_selector (string) - a selector to restrict the list of returned objects by fields.
+    #   :grace_period_seconds (integer) - the duration before an object should be deleted
+    #   :limit (integer) - a maximum number of items to return in each response
+    #   :orphan_dependents (bool) - should the dependent objects be orphaned
+    #   :propagation_policy (string) - one of Foreground|Background|Orphan
+    #   :resource_version (string) - sets a limit on the resource versions that can be served
+    #   :resource_version_match (string) - determines how the resource_version constraint will be applied
```
cben, line 481: lib/kubeclient.rb:481:101: C: Layout/LineLength: Line is too long. [105/100]

Opened #547 to add support `resource_version_match` in other actions.

cben:
Sorry for delay.  Generally looks good :+1:  

Please document this in README.  Including what it returns (the content of the deleted resources?) because it's not entirely obvious.

Let's see, how this compares to delete-single interface:
```
delete_foo('name', 'namespace', opts)    # namespaced
delete_foo('name', nil, opts)            # global
delete_foos(namespace: 'ns', **opts)     # namespaced
delete_foos(**opts)                      # global
```
Not consistent but it mirrors the existing inconsistency between `get_foo` and `get_foos`.
And I do like keyword arguments in new APIs.  So no objection :+1:

* Is this dangerous that this is only 1 letter off from deleting a single resource?  
  If I meant `delete_pod` but accidentally mistype `delete_pods('mypod', 'ns')` will it error, or delete stuff?

Another point worth documenting: What happens if you omit `namespace:` with a namespaced resource, e.g. pods — will it error, or delete all pods in all namespaces?  
I see the unit test covers exactly this scenario — it sends DELETE request to `http://localhost:8080/api/v1/pods`.  (makes sense, `build_namespace_prefix` has no idea  whether it's used with a resource supposed to be namespaced)  
But will that work?  In k8s API I only see [documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#delete-collection-pod-v1-core) for DELETE `/api/v1/namespaces/{namespace}/pods`.

* can you please add a test case using a `namespace:` param?

russell:
> Sorry for delay. Generally looks good 👍

I have rebased the PR since you last looked. DW, i'm even slower ;)

> Please document this in README. Including what it returns (the content of the deleted resources?) because it's not entirely obvious.

It returns the same thing as doing a `get_foos` and the doc suggests looking at get to understand it's potential return values.

> * Is this dangerous that this is only 1 letter off from deleting a single resource?
>   If I meant `delete_pod` but accidentally mistype `delete_pods('mypod', 'ns')` will it error, or delete stuff?

maybe, but it's no more dangerous than any other api IMHO. 

if you mistype it you get

```
[5] pry(main)> client.delete_pods('foo', namespace: 'default')
ArgumentError: wrong number of arguments (given 2, expected 0..1)
from /Users/rsim/Code/zendesk/kubeclient/lib/kubeclient.rb:313:in `block (2 levels) in define_entity_methods'
[6] pry(main)> client.delete_pods('foo', 'default')
ArgumentError: wrong number of arguments (given 2, expected 0..1)
from /Users/rsim/Code/zendesk/kubeclient/lib/kubeclient.rb:313:in `block (2 levels) in define_entity_methods'
```

so it won't accept any positional argument for a name.

> Another point worth documenting: What happens if you omit `namespace:` with a namespaced resource, e.g. pods — will it error, or delete all pods in all namespaces? I see the unit test covers exactly this scenario — it sends DELETE request to `http://localhost:8080/api/v1/pods`. (makes sense, `build_namespace_prefix` has no idea whether it's used with a resource supposed to be namespaced) But will that work? In k8s API I only see [documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#delete-collection-pod-v1-core) for DELETE `/api/v1/namespaces/{namespace}/pods`.

Nothing, you get an error

```
Kubeclient::HttpError: the server does not allow this method on the requested resource
```

> 
> * can you please add a test case using a `namespace:` param?

I have updated all test cases to handle `namespace:` because it's the only working form of the method

cben:
CI failures are just flaky timeouts.  Close-cycling to re-run.

Issue #525, opened 2021-12-13
updating from 4.9 to `master` breaks ssl


grosser:
tried to use
```
gem 'kubeclient', git: "https://github.com/abonas/kubeclient.git"
```
and got:
```
Kubeclient::HttpError: SSL_connect returned=1 errno=0 state=error: certificate verify failed (self signed certificate in certificate chain)
.gem/ruby/3.0.2/bundler/gems/kubeclient-967f4a179eaa/lib/kubeclient.rb:173:in `rescue in handle_exception'
.gem/ruby/3.0.2/bundler/gems/kubeclient-967f4a179eaa/lib/kubeclient.rb:159:in `handle_exception'
.gem/ruby/3.0.2/bundler/gems/kubeclient-967f4a179eaa/lib/kubeclient.rb:466:in `create_entity'
```

... even this did not fix it:
```
ssl_options[:verify_ssl] = OpenSSL::SSL::VERIFY_NONE
```

this blocks testing of https://github.com/abonas/kubeclient/issues/525

cben:
@grosser I'm trying to reproduce now but I don't see this happening.  Can you give more details?
- what ruby version?
- which auth_opitons & ssl_options options are you using? (without sensitive values of course)
- does your cluster use a self-signed cert? (though I agree VERIFY_NONE not supressing that is a bug)
- does it happen on `.discover`? on faraday actions (get/create/update/patch)?  http.rb actions (watch)?  
  (your traceback shows `create_entity` but I'm curious what else)

most usefully, any chance you can bisect when this started?  (at a guess, it's likely we messed up some options during faraday switch)

cben:
If I comment out `certificate-authority-data` in kubeconfig (with a local k0s cluster, self-signed),  I do see SSL error, but that's correct (kubectl agrees), and that error is slightly different: `certificate verify failed (unable to get local issuer certificate)`.

However, what's weird in that situation `context.ssl_options[:verify_ssl]` is already 0 == VERIFY_NONE (yet still getting that error!), and forcing it to VERIFY_NONE makes zero difference.

```ruby
      if cluster_ca_data?(cluster)
        cert_store = OpenSSL::X509::Store.new
        populate_cert_store_from_cluster_ca_data(cluster, cert_store)
        ssl_options[:verify_ssl] = OpenSSL::SSL::VERIFY_PEER
        ssl_options[:cert_store] = cert_store
      else
[...]

PR #545, opened 2022-02-26
[v4.y] Test VERIFY_PEER / VERIFY_NONE work against real cluster

cben:
Followup to https://github.com/ManageIQ/kubeclient/pull/540.  (It fixed https://github.com/abonas/kubeclient/issues/525 which was only broken on master, not v4.y, but good to have these tests here too.)

Adding tests that `Client` honor VERIFY_PEER/VERIFY_NONE by connecting to the temporary cluster during test/config/update_certs_k0s.rb script.  
They are skipped during a regular `rake test` so they will NOT run in CI.  
(Can we run docker --priviledged in github actions?)


cben:
Ooh, exciting!  Yes, it can run in CI.

- [ ] it failed with SSL errors differing from what I expected :tada: 
https://github.com/ManageIQ/kubeclient/runs/5346969200?check_suite_focus=true#step:6:60

- [ ] Should switch to docker.pkg.github.com/k0sproject/k0s/k0s so GitHub actions don't hammer docker hub.

cben:
OK pin-pointed the timeout:
```
KubeclientRealClusterTest#test_real_cluster_verify_none = 1.15 s = .
KubeclientRealClusterTest#test_real_cluster_verify_peer = 
Error: Process completed with exit code 124.
```

Probably `exercise_watcher_with_timeout`, which suggests `.finish` on watches is generally not as reliable as I believed :-(

cben:
@russell @grosser @agrare I still need to tackle the runaway test ^ but please review the general direction.  

I want real end-to-end tests of when TLS verification actually verifies, and the setup here of starting a real cluster works surprisingly fast and opens the door to many more future real-cluster tests...

test/config/update_certs_k0s.rb
```diff
@@ -36,6 +36,14 @@ def sh!(*cmd)
 sh! "#{DOCKER} exec #{CONTAINER} cat /var/lib/k0s/pki/admin.crt  > test/config/external-cert.pem"
 sh! "#{DOCKER} exec #{CONTAINER} cat /var/lib/k0s/pki/admin.key  > test/config/external-key.rsa"
 
-sh! 'bundle exec rake test'
+# Wait for apiserver to be up.  To speed startup, this only retries connection errors;
+# without `--fail-with-body` curl still returns 0 for well-formed 4xx or 5xx responses.
+sleep(1) until sh?('curl --cacert test/config/external-ca.pem ' \
+                        '--key test/config/external-key.rsa ' \
+                        '--cert test/config/external-cert.pem  https://127.0.0.1:6443/healthz')
+
```
grosser, line 44: less funky spacing
```suggestion
sleep(1) until sh?(
  'curl --cacert test/config/external-ca.pem ' \
  '--key test/config/external-key.rsa ' \
  '--cert test/config/external-cert.pem  https://127.0.0.1:6443/healthz'
)

```

test/test_helper.rb
```diff
@@ -12,6 +12,14 @@ def open_test_file(name)
   File.new(File.join(File.dirname(__FILE__), name.split('.').last, name))
 end
 
+# kubeconfig files deviate from above convention.
+# They link to relaved certs etc. with various extensions, all in same dir.
+def config_file(name)
```
grosser, line 17: can avoid global scope with:
```suggestion
MiniTest::Test.class_eval do
  def config_file(name)
```

grosser (approved):
❤️ nice approach

test/test_helper.rb
cben: done

PR #546, opened 2022-02-27
[v4.y] Ensure VCR doesn't leak into other tests

cben:
VCR was interfering with my attempts to add tests connecting to a real cluster (#545), despite `WebMock.enable_net_connect!`.
What's worse, it depended on order that tests executed (errored iff VCR run test run first).

[This won't be needed on master, we dropped the VCR test in #495.]

cc @grosser


grosser (commented):
nice!

Issue #543, opened 2022-02-23
Ruby 3.1 support


cben:
- it's not currently in test matrix (ruby-head is 3.2-to-be but not treating it as blocker)
- Locally with 3.1.0 on linux, `bundle install` works but tests fail with:
  ```
  File does not exist: net/smtp
  ```
  Apparently spun out to 'net-smtp' gem.  Not sure why we depend on it (real fix probably belongs in some library we depend on that imports it).
- Tests print many deprecations.

cben:
Currently the only deprecations message I see in 3.1 runs is:
```
/Users/runner/hostedtoolcache/Ruby/3.1.1/x64/lib/ruby/gems/3.1.0/gems/httpclient-2.8.3/lib/httpclient/ssl_config.rb:370: warning: assigned but unused variable - pathlen
/Users/runner/hostedtoolcache/Ruby/3.1.1/x64/lib/ruby/gems/3.1.0/gems/httpclient-2.8.3/lib/httpclient/ssl_config.rb:51: warning: method redefined; discarding old initialize
/Users/runner/hostedtoolcache/Ruby/3.1.1/x64/lib/ruby/gems/3.1.0/gems/httpclient-2.8.3/lib/httpclient/ssl_config.rb:58: warning: method redefined; discarding old add_cert
/Users/runner/hostedtoolcache/Ruby/3.1.1/x64/lib/ruby/gems/3.1.0/gems/httpclient-2.8.3/lib/httpclient/ssl_config.rb:58: warning: method redefined; discarding old add_file
/Users/runner/hostedtoolcache/Ruby/3.1.1/x64/lib/ruby/gems/3.1.0/gems/httpclient-2.8.3/lib/httpclient/ssl_config.rb:58: warning: method redefined; discarding old add_path
```
which is nothing we can fix here.  I guess I saw more in the past and they got fixed in various gems? :shrug:

PR #548, opened 2022-03-05
[v4.y] Test on Ruby 3.1

cben:
stealing @kjetijor's commit from #537.
Is net-smtp all we need to fix the 3.1 tests #543 ? :tada:
I'm still not sure where `net/smtp` actually gets `require`-d, but pretty certain it's not in kubeclient itself so a dev dependency sounds right.


cben:
Should I also backport #526 + #531 to make ruby-head, truffleruby-head `continue-on-error`?  Well this passed ALL tests, so for now no need :man_dancing:

PR #549, opened 2022-03-05
Test on Ruby 3.1

### LINKED ISSUE #543: Ruby 3.1 support

cben:
stealing @kjetijor's commit 0781a9dd8801227eca1094fbdb2e92e31b74f070 from https://github.com/ManageIQ/kubeclient/pull/537 (again, this time on master branch)
I'm still not sure where net/smtp actually gets require-d, but pretty certain it's not in kubeclient itself so a dev dependency sounds right.

Fixes #543 :tada:

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -14,7 +14,7 @@ jobs:
     continue-on-error: ${{ contains(matrix.ruby, '-head') }}
     strategy:
       matrix:
-        ruby: ['2.5', '2.6', '2.7', '3.0', 'ruby-head', 'truffleruby-head']
+        ruby: ['2.5', '2.6', '2.7', '3.0', '3.1', 'ruby-head', 'truffleruby-head']
         os: ['ubuntu-latest', 'macos-latest']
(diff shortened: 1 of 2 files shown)


cben:
2.5 run was just stuck, which I'm sometimes seeing recently.  minitest starts, and then... nothing for 20min? (because I cancelled manually. saw 5hours in another run!)
```
Run options: --seed 40851

# Running:

Error: The operation was canceled.
```
re-running...

cben:
2.5 stuck again.  But OK, 3.1 and ruby-head (3.2-to-be) passed, merging.

I'll add test timeouts in another PR.

PR #550, opened 2022-03-14
Test VERIFY_PEER / VERIFY_NONE work against real cluster

cben:
Same as #545 but merged into master branch.


cben:
As explained in #545 these are somewhat flaky but provide value, merging.

PR #552, opened 2022-03-22
[v4.y] Load cluster ca certificates using OpenSSL::X509::Store#add_file

### LINKED ISSUE #460: Kubeclient does not support certificate authorities with intermediate certificate

cben:
Backporting #461 (which fixed #460) + tests.


cben:
@jperville @DocX I'm backporting the `add_file` fix here which will allow me to finally release it (sorry for huge delay).

Added tests and CI keeps failing on Mac — any idea if `add_file` might behave differently by OS?

lib/kubeclient/config.rb
```diff
@@ -131,11 +130,16 @@ def fetch_context(context_name)
...
+    end
+
+    def populate_cert_store_from_cluster_ca_data(cluster, cert_store)
       if cluster.key?('certificate-authority')
-        File.read(ext_file_path(cluster['certificate-authority']))
+        cert_store.add_file(ext_file_path(cluster['certificate-authority']))
       elsif cluster.key?('certificate-authority-data')
-        Base64.decode64(cluster['certificate-authority-data'])
+        ca_cert_data = Base64.decode64(cluster['certificate-authority-data'])
+        cert_store.add_cert(OpenSSL::X509::Certificate.new(ca_cert_data))
```
cben, line 142: Umm, looks like the fix only covered `certificate-authority`, but `certificate-authority-data` would still have the bug.
- [ ] need to add test with concatenated inline data.
- [ ] can we fix it?
- [ ] if not, correct CHANGELOG to explain partial fix.

cben:
Oops, maybe the Mac angle was my mistake.

The cert I put in middle of concatenated-ca.pem didn't match the existing certs, so it failed `rake test`.  OTOH linux doesn't run `rake test` on existing certs, it only runs tests inside
`test/config/update_certs_k0s.rb` which regenerates all certs consistently, therefore it passed.

I'm changing CI to complete all tests when one errors to get clearer picture => :tada: mac passed.
(windows — which I never tried in CI before — failed to install gems, I'll split that to separate PR/issue)

Still not handling `certificate-authority-data`.  OK, I'll document and release soon the partial fix — better than nothing.

Issue #555, opened 2022-03-23
`Config` ignores `insecure-skip-tls-verify` field


cben:
Tightly related to #554, also broken ever since `Kubeclient::Config` was added (I'm fixing them together): `insecure-skip-tls-verify` field in kubeconfig was never honored.

The distinction is that #554 is about the default being dangerous, and this ticket is about inability to override the default (either way) by `insecure-skip-tls-verify`.

cben:
Closing, this has been fixed both in 4.9.3 and on master branch.

Issue #554, opened 2022-03-23
CVE-2022-0759 VULNERABILITY: `Config` defaults to `VERIFY_NONE` when kubeconfig doesn't specify custom CA - autoclosed


cben:
# this was assigned [CVE-2022-0759](https://www.cve.org/CVERecord?id=CVE-2022-0759)

Dangerous bug present ever since [`Kubeclient::Config` was created](https://github.com/ManageIQ/kubeclient/pull/127/files#diff-32e70f2f6781a9e9c7b83ae5e7eaf5ffd068a05649077fa38f6789e72f3de837R41-R48):

Whenever kubeconfig did not define custom CA (normal situation for production clusters with public domain and certificate!), `Config` was returning ssl_options[:verify_ssl] hard-coded to `VERIFY_NONE` :-(

Assuming you passed those ssl_options to Kubeclient::Client, this means that instead of checking server's certificate against your system CA store, it would accept ANY certificate, allowing easy man-in-the middle attacks.

This is especially dangerous with user/password or token credentials because MITM attacker could simply steal those credentials to the cluster and do anything you could do on the cluster.

Tightly related to #555, I'm fixing them together. This ticket is about the dangerous default, #555 is about inability to override it.

cben:
This was broken IN ALL RELEASES MADE BEFORE 2022.  I'm releasing a 4.9.3 with fix today.
**Anyone who needs backports to other versions, let me know**

cben:
@lnacshonredhat (or anyone else who knows): can you help describe the impact when using **client-certificate auth**?  My partial understanding:

- An attack on the app using kubeclient is still possible.
- Attacker can still fool kubeclient by returning made-up responses, it doesn't need 2nd connection to real cluster (because the client doens't verify the servers's identity).
  - This way attacker may get from kubeclient content of any resources kubeclient writes, e.g. if kubeclient is used to create/update secrets those will leak :frowning_face:.

- However, the client cert's secret key is not sent, so attacker cannot steal the credentials for later use, only risks are during the active attack.

- The server does verify client's identity! :sweat_smile: 
  What degree of man-in-the-middle attack is possible?
  - Obviously attacker pass through the data byte-for-byte both ways (kubeclient⭤MITM⭤real cluster), but that's what any internet router does, so attacker can't decrypt.
  - Can attacker relay the inner HTTP protocol unmodified AND decrypt it?  Or does decryption require changing the certs in a way that will make the cluster reject the TLS connection?
[...]

PR #556, opened 2022-03-23
[v4.y] SECURITY: Kubeclient::Config: return ssl_options[:verify_ssl] correctly

cben:
Fixing #554 and #555, this branch will be released as 4.9.3.

Split into fix/unit tests/real-cluster integration tests commits for ease of backporting.
(the first fix commit itself is branched directly from ancient #127 that introduced the issue, hence the later merge commits)

cc @grosser @agrare @jperville @lnacshonredhat


cben:
umm I forgot to bump lib/kubeclient/version.rb

PR #557, opened 2022-03-23
SECURITY: Kubeclient::Config: return ssl_options[:verify_ssl] correctly

cben:
Fixing https://github.com/ManageIQ/kubeclient/issues/554 and https://github.com/ManageIQ/kubeclient/issues/555 on master branch.

(also merges tests & changelog added in #552)


cben:
cc @grosser @agrare @jperville @lnacshonredhat

The git history has become hard to read here (due to my attempt to keep it "clean" in the "prefer merges to rebases" sense, LOL)...  
But this is a straight forward merge of #556 into master, only conflicts were CHANGELOG and github actions.yml (over `continue-on-error` — b1824ed8a770ca22aa9520c8e3479c4c84b17f1d).

cben:
:red_circle: truffleruby-head on ubuntu always times out on real-cluster tests, I guess watch `#finish` doesn't work there?  Not new, not merge blocker.

:red_circle: this time [ruby 2.6 on ubuntu also timed out](https://github.com/ManageIQ/kubeclient/runs/5659411390?check_suite_focus=true#step:6:263).  It did pass real-cluster tests :heavy_check_mark:, just was too slow.  Perhaps I should increase 3min timeout :thinking:
Re-running...

.github/workflows/actions.yml
```diff
@@ -10,8 +10,8 @@ on:
     - '**'
 jobs:
   build:
+    continue-on-error: true
```
grosser, line 13: I think this is wrong and will result in commits marked as passing when they failed CI

README.md
```diff
@@ -10,6 +10,12 @@ The client supports GET, POST, PUT, DELETE on all the entities available in kube
 The client currently supports Kubernetes REST api version v1.
 To learn more about groups and versions in kubernetes refer to [k8s docs](https://kubernetes.io/docs/api/)
 
+## VULNERABILITY❗
+
+If you use `Kubeclient::Config`, all gem versions released before 2022 could return incorrect `ssl_options[:verify_ssl]`,
```
grosser, line 15: should say `before va.b.c` ?
... ideally also change heading to `## VULNERABILITY in <va.b.c` so I can quickly ignore the heading

grosser (commented):
thx for the fix and great summary :)

README.md
grosser: see https://github.com/ManageIQ/kubeclient/pull/560

.github/workflows/actions.yml
cben: I'm pretty certain the overall status is still failed 🔴 (as on all recent PRs where truffleruby fails), its just that individual runs continue till the end.

PR #566, opened 2022-07-11
[4.y] Refresh token on every request

harshit-splunk:
https://github.com/ManageIQ/kubeclient/issues/561
It is not an optimal solution since it reads token on every request.
As an enhancement, we can watch changes in the token file and update the token when it detects any changes.

### DIFF
--- a/lib/kubeclient/common.rb
+++ b/lib/kubeclient/common.rb
@@ -36,6 +36,8 @@ module Kubeclient
     DEFAULT_HTTP_PROXY_URI = nil
     DEFAULT_HTTP_MAX_REDIRECTS = 10
 
+    DEFAULT_BEARER_TOKEN_FILE = '/var/run/secrets/kubernetes.io/serviceaccount/token'
+
     SEARCH_ARGUMENTS = {
       'labelSelector'   => :label_selector,
       'fieldSelector'   => :field_selector,
@@ -78,7 +80,7 @@ module Kubeclient
       @api_version = version
       @headers = {}
       @ssl_options = ssl_options
-      @auth_options = auth_options
+      @auth_options = Marshal.load(Marshal.dump(auth_options))
       @socket_options = socket_options
       # Allow passing partial timeouts hash, without unspecified
       # @timeouts[:foo] == nil resulting in infinite timeout.
@@ -87,11 +89,25 @@ module Kubeclient
       @http_max_redirects = http_max_redirects
       @as = as
 
+      @log = Logger.new(STDOUT)
+      @log.formatter = proc do |severity, datetime, progname, msg|
+          "#{datetime} [#{severity}]: #{msg}\n"
+      end
+
       if auth_options[:bearer_token]
         bearer_token(@auth_options[:bearer_token])
-      elsif auth_options[:bearer_token_file]
+      end
+      if auth_options[:bearer_token_file]
         validate_bearer_token_file
+        @log.info("Reading bearer token from #{@auth_options[:bearer_token_file]}")
         bearer_token(File.read(@auth_options[:bearer_token_file]))
+      elsif File.file?(DEFAULT_BEARER_TOKEN_FILE)
+        @auth_options[:bearer_token_file] = DEFAULT_BEARER_TOKEN_FILE
+        validate_bearer_token_file
+        @log.info("Reading bearer token from default location #{DEFAULT_BEARER_TOKEN_FILE}")
+        bearer_token(File.read(@auth_options[:bearer_token_file]))
+      else
+        @log.info("bearer_token_file path not provided. Kubeclient will not be able to refresh the token if it expires")
       end
     end
 
@@ -134,6 +150,11 @@ module Kubeclient
       load_entities
       define_entity_methods
       @discovered = true
+    end
+
+    def get_headers
+      bearer_token(File.read(@auth_options[:bearer_token_file])) if @auth_options[:bearer_token_file]
+      @headers
     end
 
     def self.parse_definition(kind, name)
@@ -349,7 +370,7 @@ module Kubeclient
       ns_prefix = build_namespace_prefix(options[:namespace])
       response = handle_exception do
         rest_client[ns_prefix + resource_name]
-          .get({ 'params' => params }.merge(@headers))
+          .get({ 'params' => params }.merge(get_headers))
       end
       format_response(options[:as] || @as, response.body, entity_type)
     end
@@ -362,7 +383,7 @@ module Kubeclient
       ns_prefix = build_namespace_prefix(namespace)
       response = handle_exception do
         rest_client[ns_prefix + resource_name + "/#{name}"]
-          .get(@headers)
+          .get(get_headers)
       end
       format_response(options[:as] || @as, response.body)
     end
@@ -378,7 +399,7 @@ module Kubeclient
           rs.options.merge(
             method: :delete,
             url: rs.url,
-            headers: { 'Content-Type' => 'application/json' }.merge(@headers),
+            headers: { 'Content-Type' => 'application/json' }.merge(get_headers),
             payload: payload
           )
         )
@@ -400,7 +421,7 @@ module Kubeclient
       hash[:apiVersion] = @api_group + @api_version
       response = handle_exception do
         rest_client[ns_prefix + resource_name]
-          .post(hash.to_json, { 'Content-Type' => 'application/json' }.merge(@headers))
+          .post(hash.to_json, { 'Content-Type' => 'application/json' }.merge(get_headers))
       end
       format_response(@as, response.body)
     end
@@ -410,7 +431,7 @@ module Kubeclient
       ns_prefix = build_namespace_prefix(entity_config[:metadata][:namespace])
       response = handle_exception do
         rest_client[ns_prefix + resource_name + "/#{name}"]
-          .put(entity_config.to_h.to_json, { 'Content-Type' => 'application/json' }.merge(@headers))
+          .put(entity_config.to_h.to_json, { 'Content-Type' => 'application/json' }.merge(get_headers))
       end
       format_response(@as, response.body)
     end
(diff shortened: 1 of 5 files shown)


lib/kubeclient/common.rb
```diff
@@ -78,7 +80,7 @@ def initialize_client(
       @api_version = version
       @headers = {}
       @ssl_options = ssl_options
-      @auth_options = auth_options
+      @auth_options = Marshal.load(Marshal.dump(auth_options))
```
cben, line 83: I presume the goal here is deep copy?

I'm not very familiar with `Marshal`, trying to understand whether this is always safe...
A particular scenario to consider: auth_options will frequently come from a parsed kubeconfig e.g.:
```
config = Kubeclient::Config.read(ENV['KUBECONFIG'] || '/path/to/.kube/config')
context = config.context
Kubeclient::Client.new(
  context.api_endpoint,
  'v1',
  ssl_options: context.ssl_options,
  auth_options: context.auth_options  # <-- under external control?
)
```
[Marshal docs warn against de-serializing untrusted input](https://ruby-doc.org/core-3.1.2/Marshal.html#module-Marshal-label-Security+considerations).  But here we don't quite have untrusted input. The kubeconfig can be parsed into "safe_load" types — arbitrary mix of hashes, arrays, strings, numbers, booleans and dates -> Marshal.dump -> Marshal.load.
Probably it can only deserialize into same "safe" types?

lib/kubeclient/common.rb
```diff
@@ -87,11 +89,25 @@ def initialize_client(
...
 
+      @log = Logger.new(STDOUT)
+      @log.formatter = proc do |severity, datetime, progname, msg|
+          "#{datetime} [#{severity}]: #{msg}\n"
+      end
+
       if auth_options[:bearer_token]
         bearer_token(@auth_options[:bearer_token])
-      elsif auth_options[:bearer_token_file]
+      end
```
cben, line 99: I wonder whether this `end` should be `elsif`.  Specifically, if one runs inside a cluster and passes an explicit `bearer_token: ...`, the auto-detected `DEFAULT_BEARER_TOKEN_FILE` will override it?

lib/kubeclient/common.rb
```diff
@@ -87,11 +89,25 @@ def initialize_client(
...
+      if auth_options[:bearer_token_file]
+        validate_bearer_token_file
+        @log.info("Reading bearer token from #{@auth_options[:bearer_token_file]}")
+        bearer_token(File.read(@auth_options[:bearer_token_file]))
+      elsif File.file?(DEFAULT_BEARER_TOKEN_FILE)
```
cben, line 104: I'd prefer to drop DEFAULT_BEARER_TOKEN_FILE from this PR.  This PR can support re-loading token but still require user to pass an explicit `:bearer_token_file`.

Why? Built-in support for standard in-cluster config is handy (it's silly that to this day it remains a recipe in README) but:

1. I'd prefer not to add functionality to v4.y branch that we don't have on master (unless you want to do both).

2. This only detects the token part, doesn't cover `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`.  I'm not sure if best way to support in-cluster config is auto-detection, vs. a separate factory e.g. `Kubeclient::Client.in_cluster()`.  If we don't need to decide these aspects here, we can land the re-loading part faster.

cben (commented):
Thanks, and sorry for slow review!  I haven't tried this yet.
Some questions/nitpicts, but overall, the approach makes sense :+1:

Can you add any tests?
- one approach is use webmock to assert on headers?
- another is extend `test/test_real_cluster.rb` which runs as part of `test/config/update_certs_k0s.rb` script.  
  Something like create a client with `:bearer_token_file`, make a successful request, rewrite file with bad token, assert next request fails, rewrite file back with good token, assert next request succeeds.
  (Pro: makes actual requests against real k0s cluster! Con: only runs on linux.)

lib/kubeclient/common.rb
cben: I googled more and this appears to be [the idiomatic way](https://stackoverflow.com/a/17913436/239657) to deep copy using ruby stdlib.  :heavy_check_mark: 

The other "obvious" contender is ActiveSupport [.deep_dup](https://guides.rubyonrails.org/active_support_core_extensions.html#deep-dup) but kubeclient being a rails-independent library probably shouldn't pull in ActiveSupport's monkey patches.

Finally, a shallow copy might suffice here?

cben:
Tests failed on ruby 2.6 & 2.7 are unrelated to this PR, #569

cben:
> It is not an optimal solution since it reads token on every request.
> As an enhancement, we can watch changes in the token file and update the token when it detects any changes.

That's fine for now, reload on every request is also what got implemented on master branch.

cben:
For tests, #567 that just merged to master contains 2 unit tests for watching that can probably be easily backported.  Plus should cover regular non-watch actions (get, delete etc.).

lib/kubeclient/common.rb
harshit-splunk: I was following the [the idiomatic way](https://stackoverflow.com/a/17913436/239657) only. I can switch it to shallow copy

lib/kubeclient/common.rb
harshit-splunk: Makes sense!
I'll remove DEFAULT_BEARER_TOKEN_FILE from this PR.

lib/kubeclient/common.rb
harshit-splunk: Yes, it would override it. I have removed auto-detection so users need to manually provide path to enable refreshing the token

harshit-splunk:
Sorry for the late response. I have added unit tests and removed `DEFAULT_BEARER_TOKEN_FILE`

cben:
Sorry for silence. We're on a long family trip; I thought I'll have lots of time for hacking but I had zero :rofl:  
Looking into this PR today.  The code all LGTM, just need to see what's going on with the CI errors...

harshit-splunk:
@cben Is there any release planned for this?

cben:
Yes, definitely.  But at i said, i want to diagnose #569 , any help welcome.

lib/kubeclient/common.rb
```diff
@@ -87,11 +87,18 @@ def initialize_client(
...
+          "#{datetime} [#{severity}]: #{msg}\n"
+      end
+
+      if auth_options[:bearer_token_file]
         validate_bearer_token_file
+        @log.info("Reading bearer token from #{@auth_options[:bearer_token_file]}")
         bearer_token(File.read(@auth_options[:bearer_token_file]))
+      elsif auth_options[:bearer_token]
+        bearer_token(@auth_options[:bearer_token])
+        @log.info("bearer_token_file path not provided. Kubeclient will not be able to refresh the token if it expires")
```
rvignesh89, line 101: Is there any strong reason to add this log line? We are using kubeclient in a CLI app where the bearer token is refreshed through a process outside kubeclient. Having this line messes up our CLI logging now. Does it make sense to make it atleast a debug level?

cben: Good point.  
It seems this is first place where kubeclient is logging at all on v4.y branch.
The similar-but-different implementation on master branch won't have these logs (though it has some other logging in new Informer class, including error level).

=> I tend to agree by default these should not be visible.  
=> Downgrading both to .debug level sounds fine to me.  
=> Other options to control logging, e.g. injecting a Logger object, are also possible but I think we can defer such over-thinking to master branch and future 5.y releases...

@harshit-splunk do you have an opinion on this?  
@rvignesh89 Could you send a PR?

cben: Hmm I see debug is still visible by default:
```
$ ruby -e 'require "logger"; Logger.new(STDOUT).debug("foo")'
D, [2022-09-04T15:15:12.509754 #747358] DEBUG -- : foo
```
I guess it's better for back-compat (at least on 4.y versions) that logging be opt-in, and by default kubecliet remain silent.  WDYT?

In this case the logging is merely giving (useful) advice on parameters passed to constructor.  _Technically_, if we just dropped them, the caller could easily implement similar functionality.  

Though once they become aware of the idea these messages suggest, they don't even need them logged.  It's similar to a deprecation message (which suggests it should be at most printed once), except that we're not really deprecating `:bearer_token`, it remains useful.

cben: Hmm, I see we forgot to mention the new refresh functionality in README.md.  
=> I'm thinking we could drop these messages altogether and replace them with explanation of the benefit in README — but would like to hear @harshit-splunk's opinion.

harshit-splunk: I think we should remove the logs since other components don't log. I only added those to debug this functionality. I should have removed after testing.

rngtng: I'd vote for removing logging or at least have the opportunity to inject/config the logger. In my case'd like to silence the log as it 'pollutes' STDOUT which I need for other reasons

rngtng: btw. my (hopefully temporary) hack is to monkey patch for a local`STDOUT` logging into the void....

```
module Kubeclient
  module ClientMixin
    STDOUT = StringIO.new
  end
end
```

joshbranham: Is there open work to remove this line? I agree it should either not be in stdout or be classified as debug logging.

harshit-splunk: I have created a PR(https://github.com/ManageIQ/kubeclient/pull/577)

cben: ^ Thanks. Merged and released 4.10.1.  Also added some words in the README about `bearer_token` vs `bearer_token_file` (#579).

joshbranham: @cben I see 4.10.1 in Rubygems but not reflected here, is there a reason why? https://github.com/ManageIQ/kubeclient/blob/master/lib/kubeclient/version.rb

joshbranham: Nevermind I see there is a long running branch for the 4.x release here https://github.com/ManageIQ/kubeclient/blob/v4.y/CHANGELOG.md

Might be worth adding a note to the readme so folks know where to find the latest release notes

PR #567, opened 2022-07-18
fix: move reading bearer token file to WatchStream

sumo-drosiek:
WatchStream uses old bearer token, because it is not re-readed by `WatchStream`

### DIFF
--- a/test/test_watch.rb
+++ b/test/test_watch.rb
@@ -110,6 +110,7 @@ class TestWatch < MiniTest::Test
       'http://localhost:8080/api/', 'v1',
       auth_options: { bearer_token_file: file.path }
     )
+    watcher = client.watch_pods(as: :raw)
 
     begin
       file.write("valid_token")
@@ -119,7 +120,7 @@ class TestWatch < MiniTest::Test
         .to_return(body: open_test_file('watch_stream.json'), status: 200)
 
       got = nil
-      client.watch_pods(as: :raw).each { |notice| got = notice }
+      watcher.each { |notice| got = notice }
       assert_match(/\A{"type":"DELETED"/, got)
       remove_request_stub(stub_token)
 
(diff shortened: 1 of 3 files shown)


cben:
This matters when there is significant delay between
```
watcher = kclient.watch_foo(...)
```

cben:
IIUC this matters when there is significant delay between creating instance:
```
watcher = kclient.watch_foo(...)
```
and actually starting the watch:
```
watcher.each do
  ...
end
```

Thanks for the fix!  We've already had a series of several PRs fixing reloading-tokens-in-watches area, and looks like we still missed this.
Could add a test to cover this? :pray:   
(See some testing suggestions on https://github.com/ManageIQ/kubeclient/pull/566#discussion_r928234444, that PR is on v4.y branch but the desired functionality is same, so hopefully a similar test could work on both)

- Tests failed on ruby 2.6 & 2.7 are unrelated to this PR, #569

sumo-drosiek:
> See some testing suggestions on https://github.com/ManageIQ/kubeclient/pull/566#discussion_r928234444

Didn't catch any testing suggestion there. However I added some to this PR which I believe covers the functionality

test/test_watch.rb
```diff
@@ -102,6 +102,42 @@ def test_watch_pod_max_redirect
...
+
+    begin
+      file.write("valid_token")
+      file.rewind
+      stub_token = stub_request(:get, %r{/watch/pods})
+        .with(headers: { Authorization: 'Bearer valid_token' })
+        .to_return(body: open_test_file('watch_stream.json'), status: 200)
+
+      got = nil
+      client.watch_pods(as: :raw).each { |notice| got = notice }
```
cben, line 122: Beautiful tests overall :clap:  

However, I've checked your branch locally, re-ordered the commits, and confirmed the new tests also pass before your fix.  
This one-liner creates a WatchStream and calls `.each` at same time.  
However, if you extract `watcher = client.watch_pods(as: :raw)` to the top, then call `watcher.each { ... }` after the file writes, it fails without the fix and passes with it.

sumo-drosiek: o wow, thanks for testing it. There are some nuances which I missed :sweat_smile:

sumo-drosiek: Should be better now

cben:
LGTM :100:

PR #571, opened 2022-08-04
Update HTTP gem requirement to allow version 5

nielsslot:
I wanted to use kubeclient in a project and noticed it required me to downgrade the version of the http gem to version 4. Version 5.0.0 of the http gem was released more than a year ago. Their [changelog](https://github.com/httprb/http/blob/main/CHANGES.md) notes some breaking changes, but none of them seem to impact the usage in kubeclient.

This pull request also updates the rake development dependency to version 13. It looks like version 13 was [released almost three years ago](https://github.com/ruby/rake/releases/tag/v13.0.0). Updating it was the easiest way to use version 5 of the http gem in kubeclient since it is dependent on the llhttp-ffi gem, which apparently is dependent on rake 13. I very much doubt that llhttp-ffi actually needs rake at runtime, but it is what it is. And actually updating rake here does make sense after three years.

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.5.0'
 
   spec.add_development_dependency 'bundler', '>= 1.6'
-  spec.add_development_dependency 'rake', '~> 12.0'
+  spec.add_development_dependency 'rake', '~> 13.0'
   spec.add_development_dependency 'minitest'
   spec.add_development_dependency 'minitest-rg'
   spec.add_development_dependency 'webmock', '~> 3.0'
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -34,5 +34,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'faraday', '~> 1.1'
   spec.add_dependency 'faraday_middleware', '~> 1.0'
   spec.add_dependency 'recursive-open-struct', '~> 1.1', '>= 1.1.1'
-  spec.add_dependency 'http', '>= 3.0', '< 5.0'
+  spec.add_dependency 'http', '>= 3.0', '< 6.0'
 end


timflapper:
> @cben Is there anything we can do to help get this merged?

I'd also be interested to know if there's anything that can be done to get this one released.

cben:
Oh sorry, i totally dropped the ball on this.  I have family plans today but hope to get it by Sunday.

cben:
Resolved trivial (change in adjacent line) merge conflict.

Going over the breaking changes in [http's changelog](https://github.com/httprb/http/blob/main/CHANGES.md)...

- [ ] https://github.com/httprb/http/pull/576 https://github.com/httprb/http/issues/524 BREAKING CHANGE Preserve header names casing. ([@joshuaflanagan](https://github.com/joshuaflanagan))
- [ ] https://github.com/httprb/http/pull/540 https://github.com/httprb/http/issues/538 BREAKING CHANGE Require explicit MIME type for Response#parse ([@ixti](https://github.com/ixti))
- [ ] https://github.com/httprb/http/pull/546 BREAKING CHANGE Provide initiating HTTP::Request object on HTTP::Response. ([@joshuaflanagan](https://github.com/joshuaflanagan))

cben:
- [ ] ruby 2.5 [tests timed out](https://github.com/ManageIQ/kubeclient/actions/runs/3359829904/jobs/5568290089), sporadically times out (so far 2 times, passed 1 time) in `RetryTest#test_can_watch_watches` — maybe related as we use `http` for watches?

  - [x] http 5.1.0 dropped ruby 2.5 support, but bundler correctly installed http 5.0.4 here.
  
  (I really should just drop EOL 2.5 from master branch, but I kinda still want it to work on v4.y branch)

cben:
## Compatibility

Well, there is one mild change, but it's to undocumented functionality :heavy_check_mark: 

Kubeclient::Client instances have a `.headers` map that starts out initialized to `{}`.  AFAICT we never change it, and there is no documentation that users may access/mutate it.  

But if a user does put some extra headers there `client["fOO_bAR"] = "baz"`, specifically _if_ the key is a string, http 5.0 no longer normalizes the header name but will pass it to server exactly as spelled.  (symbol keys are still normalized) https://github.com/httprb/http/pull/576

=> Since it's undocumented, I'm fine with this don't consider it a breaking change.  
Also the long-term result is less "magic" happening, user will get what they asked for (that was the change's motivation in http) :+1:

**[Users: Don't take this as justification to mutate undocumented .headers.  If you need it, open an issue to publicly expose the functionality.]**

cben:
(Oh I forgot to mention long-term we want to get off http gem entirely #488, blocked on question about `WatchStream#finish`)

cben:
Hmm, the ruby 2.5 sporadic hanging in `RetryTest#test_can_watch_watches` does reproduce locally.  I don't know yet if it's new with this PR, might well have existed before.

A recipe you can leave running until it hangs:
```sh
while timeout --verbose 3m env TESTOPTS="--verbose" bundle exec rake test; do 
  ruby -e 'puts "-"*200'
done
```
but that doesn't show traceback of where it was hanging.

vindvaki:
> Hmm, the ruby 2.5 sporadic hanging in RetryTest#test_can_watch_watches does reproduce locally. I don't know yet if it's new with this PR, might well have existed before.

@cben Regardless of the cause, Ruby 2.5 (and even 2.6), are [both EOL](https://www.ruby-lang.org/en/downloads/branches/). If causing troubles in CI, maybe it's time to drop support?

cben:
You're right => dropped 2.5 & 2.6 in #583.  I also didn't notice that specific test is for `Informer` functionality doesn't exist on master branch, so my whole "but I _maybe_ still care about 2.5 working there" argument was spurious.

It's just that my chronic deficit of time for kubeclient pushes me rely more on CI, and any silly thing like this is a distraction that ends up postponing progress :-(

Over these days I also saw various failures of this test on newer rubies, including without this PR.  Opened #584 for that.

vindvaki:
Thanks for following up on this. Much appreciated 🙏

dhstewart:
@cben thank you for getting this merged in! 🎉 
 
Do you plan to bump the version ?

PR #576, opened 2022-09-23
use openid_connect gem's jwks caching feature

nov:
same with https://github.com/omniauth/omniauth_openid_connect/pull/124

when passing `OpenIDConnect::Discovery::Provider::Config::Response` instance to `OpenIDConnect::ResponseObject::IdToken.decode`, it fetches JWK Set using `JSON::JWK::Set::Fetcher`.

`JSON::JWK::Set::Fetcher` tries to cache JWKS by given `kid` when `JSON::JWK::Set::Fetcher.cache` is setup like below.

```ruby
JSON::JWK::Set::Fetcher.cache = Rails.cache
```

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rubocop', '~> 1.3.0' # locked to minor so new cops don't slip in
   spec.add_development_dependency 'googleauth', '~> 0.5'
   spec.add_development_dependency('mocha', '~> 1.5')
-  spec.add_development_dependency 'openid_connect', '~> 1.1'
+  spec.add_development_dependency 'openid_connect', '~> 1.4'
   spec.add_development_dependency 'net-smtp'
 
   spec.add_dependency 'faraday', '~> 1.1'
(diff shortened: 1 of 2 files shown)


cben:
@rhysm @benlangfeld you know more than me about OIDC, could you review?

@nov Does this deserve any additions to README?  Or is it "if you've set up a JWK cache it'll automagically do the obvious Right Thing, and if not you don't care about this?"

nov:
If OIDC login flow is executed very often, it deserve to be added to README too.

cben:
OK, I don't know how to review this myself, but I shouldn't have blocked this either.  Superficially, makes sense :+1:  
[I'm just getting too little time for kubeclient, so trying to outsource...  nagging is welcome!]

@nov please add at least a CHANGELOG.md entry (as I'm not sure how to explain this change well), and I'll merge.

cben:
@nov friendly ping — the only thing blocking merge is I don't know how to explain this change in CHANGELOG.md, please add an entry there :pray: 

FYI, #606 is looking to auto-renew credentials, and intends to call `OIDCAuthProvider.token` _on every kubeclient request_.  We need more eyes on that.

PR #579, opened 2022-10-01
Changelog & bump for 4.10.1 release

cben:
dropped unwanted logging #577


cben:
(Marked "invalid" for [Hacktoberfest](https://hacktoberfest.com/) purposes as I don't think my own PR that I alone reviewed, should count, especially an "administrative" one like this.
That doesn't mean somebody else contributing changelog or docs would be invalid — in fact any help back-filling the changelog on master branch would be very welcome contribution and would advance us towards 5.0 release.)

PR #583, opened 2022-11-23
Drop EOL ruby 2.5, 2.6 from CI on master branch.

cben:
Also:
- Fix rubocop: Apparently for a while I had a broken github actions config so Rubocop
wasn't actually running in CI... :man_facepalming: 
  Plus adjust to some new linting from bumping rubocop's TargetRubyVersion.

- Fix test_watch_pod_api_bearer_token_success leaking a temp file.


cben:
So, the ruby 3.0 failure reinforces that `test_informer.rb` is sporadically failing, even on master — but that doesn't preclude merging this.

PR #586, opened 2022-11-24
informer test flakyness: try reporting problems earlier

grosser:
https://github.com/ManageIQ/kubeclient/issues/584 describes a few cases where the test got stuck and one looked like it watched twice ... one way that could happen if we retry the loop and log errors ... so let's make that explode instead so we see what the real underlying error is
@cben

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -8,6 +8,8 @@ require 'mocha/minitest'
 require 'json'
 require_relative '../lib/kubeclient'
 
+Thread.abort_on_exception = true
+
 MiniTest::Test.class_eval do
   # Assumes test files will be in a subdirectory with the same name as the
   # file suffix.  e.g. a file named foo.json would be a "json" subdirectory.
--- a/test/test_informer.rb
+++ b/test/test_informer.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require_relative 'helper'
+require 'stringio'
+require 'logger'
 
 # tests with_retries in kubeclient.rb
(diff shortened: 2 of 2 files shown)


cben:
one failure in CI but not the one you modified here:

https://github.com/ManageIQ/kubeclient/actions/runs/3542843672/jobs/5948771586#step:6:247
```
   7) Failure:
RetryTest#test_restarts_on_error [/Users/runner/work/kubeclient/kubeclient/test/test_informer.rb:98]:
The request GET /\/v1\/pods/ was expected to execute at least 2 times but it executed 1 time

The following requests were made:

GET http://localhost:8080/api/v1/pods?resourceVersion=0 with headers {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'User-Agent'=>'Faraday v1.10.2'} was made 1 time
GET http://localhost:8080/api/v1/watch/pods?resourceVersion=1 with headers {'Connection'=>'close', 'Host'=>'localhost:8080', 'User-Agent'=>'http.rb/5.1.0'} was made 1 time
```

cben:
I'm running this locally...
But I think it's also looks good to merge.

cben:
another err seen locally:
```
  1) Failure:
RetryTest#test_can_watch_watches [/home/beni/kubeclient/test/test_informer.rb:121]:
Expected: [["ADDED"], ["ADDED"]]
  Actual: [[], ["ADDED"]]
```

OK, meanwhile I'll merge this and sooner or later it'll help get more info :shrug:

PR #587, opened 2022-11-27
informer test: reduce flakyness by waiting for correct state

grosser:
hopefully improving on https://github.com/ManageIQ/kubeclient/pull/586#issuecomment-1328202844
before we waited a set amount of time and now we wait until the watcher is ready
... which will stop and fail in that spot if it never got ready

### DIFF
--- a/lib/kubeclient/informer.rb
+++ b/lib/kubeclient/informer.rb
@@ -65,17 +65,17 @@ module Kubeclient
     end
 
     def watch_to_update_cache
-      watcher = @client.watch_entities(@resource_name, watch: true, resource_version: @started)
+      @watcher = @client.watch_entities(@resource_name, watch: true, resource_version: @started)
       stop_reason = 'disconnect'
 
       # stop watcher without using timeout
       Thread.new do
         sleep(@reconcile_timeout)
         stop_reason = 'reconcile'
-        watcher.finish
+        @watcher.finish
       end
 
-      watcher.each do |notice|
+      @watcher.each do |notice|
         case notice[:type]
         when 'ADDED', 'MODIFIED' then @cache[cache_key(notice[:object])] = notice[:object]
         when 'DELETED' then @cache.delete(cache_key(notice[:object]))
--- a/test/test_informer.rb
+++ b/test/test_informer.rb
@@ -141,7 +141,8 @@ class RetryTest < MiniTest::Test
 
   def with_worker
     informer.start_worker
-    sleep(0.03) # wait for worker to watch
+    sleep(0.01) until informer.instance_variable_get(:@watcher) # wait for worker to watch
+    sleep(0.01) # give some time for threads to store
     yield
   ensure
     informer.stop_worker


lib/kubeclient/informer.rb
```diff
@@ -65,17 +65,17 @@ def fill_cache
     end
 
     def watch_to_update_cache
-      watcher = @client.watch_entities(@resource_name, watch: true, resource_version: @started)
+      @watcher = @client.watch_entities(@resource_name, watch: true, resource_version: @started)
```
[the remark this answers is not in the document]
cben: Similar question for `@worker` which is shared across 2 methods, so there I'd say worth "announcing".

lib/kubeclient/informer.rb
```diff
@@ -65,17 +65,17 @@ def fill_cache
...
       stop_reason = 'disconnect'
 
       # stop watcher without using timeout
       Thread.new do
         sleep(@reconcile_timeout)
         stop_reason = 'reconcile'
-        watcher.finish
+        @watcher.finish
```
cben, line 75: Consider adding `@watcher = nil` to release the watcher instance to GC after this method exits? (Possibly in `ensure` since this runs in worker thread that can be `.kill`ed)

Not sure if this matters for anything.  A stopped WatchStream instance is not using much resources, and at most 1 could remain on an inactive informer.

What might matter more is another scenario, unrelated to this PR: if informer worker thread is killed, this `@watcher.finish` might not run, possibly leaking an open watch socket?  Does that need an `ensure` somewhere?  Or maybe cleaner to add `@watcher&.finish` in `#stop_worker` (possible now that you made it an instance variable)

cben: This later scenario might be exactly what happened in this CI run, .finish wasn't called:
https://github.com/ManageIQ/kubeclient/actions/runs/3560257866/jobs/5980172236#step:6:223

cben: Looking at `WatchStream` implementation, I don't think releasing `@watcher` instance is important.  It in turn holds an `HTTP::Client` instance, but .finish force-closes it, so there should be no issue.

cben:
just for the record, this still had a CI failure https://github.com/ManageIQ/kubeclient/actions/runs/3560257866/jobs/5980172236#step:6:224
```
RetryTest#test_timeout [/Users/runner/work/kubeclient/kubeclient/test/test_informer.rb:135]:
not all expectations were satisfied
unsatisfied expectations:
- expected exactly once, invoked never: #<AnyInstance:Kubeclient::Common::WatchStream>.finish(any_parameters)
satisfied expectations:
- expected never, invoked never: #<Logger:0x12188>.error()
```

cben:
@grosser whenever I "Re-run all jobs", I'm getting 1–2 failures...
Anyway this makes the test more deterministic, so I'm merging.

grosser:
maybe this helps https://github.com/ManageIQ/kubeclient/pull/593

PR #589, opened 2022-12-19
V4.y dependency update

dhstewart:
### Why

I'd like to be able to get the updated `spec.add_dependency 'http', '>= 3.0', '< 6.0'` that is currently on `master`, but with backwards compatibility for the `v4.y` version.

### What
- update the relevant dependencies
- make the minimum required version of ruby 2.7 in the gemspec
- drop two EOL ruby versions from github actions
- get rubocop running/green again

### RuboCop approach

I tried running autofix on the codebase and manually fixing what couldn't be autofixed. This resulted in some failing tests and a broad diff throughout the codebase. I wasn't comfortable with this approach and abandoned that branch (though I still have it locally) in favor of this PR.

In this PR, I'm optimizing for keeping the diff small and reviewable. I have disabled every cop that has an active offence in the codebase and moved them under an explanatory comment. If we want to bring the codebase back into compliance, I think it would be better to handle that in separate Pull Requests to make reviewing easier and to split the work out. Some offences are easy but tedious to fix while others require a refactor and I'm not sure what the level of effort would be on those yet.

Also, I have added all new cops that need to be configured under a separate comment. This gets rid of a warning in the command line, and could be configured as desired in a follow up.

Let me know what you think! Thank you!

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -27,7 +27,7 @@ jobs:
         include:
         # run rubocop against lowest supported ruby
         - os: ubuntu-latest
-          ruby: '2.5'
+          ruby: '2.7'
           command: 'bundle exec rake rubocop'
     name: ${{ matrix.os_and_command.os }} ${{ matrix.ruby }} rake ${{ matrix.os_and_command.command }}
     steps:
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os_and_command.os }}
     strategy:
       matrix:
-        ruby: [ '2.5', '2.6', '2.7', '3.0', '3.1', 'ruby-head', 'truffleruby-head' ]
+        ruby: [ '2.7', '3.0', '3.1', 'ruby-head', 'truffleruby-head' ]
         os_and_command:
         - os: macos-latest
           command: 'env TESTOPTS="--verbose" bundle exec rake test'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,6 @@
 AllCops:
   DisplayCopNames: true
-  TargetRubyVersion: 2.2 # Oldest version kubeclient supports
+  TargetRubyVersion: 2.7 # Oldest version kubeclient supports
 MethodLength:
   Enabled: false
 ClassLength:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,29 +10,30 @@ Metrics/AbcSize:
 Metrics/ParameterLists:
   Max: 5
   CountKeywordArgs: false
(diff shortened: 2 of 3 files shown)


cben:
Thanks!  Wow, I didn't expect this to become tricky, I forgot about the ruby version affecting rubocop.

In theory, you might be able to leave 2.5 or 2.6 "supported" — people using ruby 2.5 will automatically get http 4.y.z (as latest that supports their ruby), while people with 2.6+ will be able to enjoy http 5.  
But, that's up to you.  I was reluctant for a long while to drop any ruby support from kubeclient (prior to major version bump), but I agree it's OK, no need to support long-EOL versions _if_ it holds us back.

> If we want to bring the codebase back into compliance, I think it would be better to handle that in separate Pull Requests to make reviewing easier and to split the work out.

I agree 100% :+1:  This being a backport/maintenance branch, I don't think we care much about stylistic nitpicks, reducing code divergence is more valuable.

cben:
CI hasn't run, says:
>  Error when evaluating 'runs-on' for job 'build'. .github/workflows/actions.yml (Line: 14, Col: 14): Unexpected type of value '', expected type: OneOf.

:confused: Maybe the `include` for rubocop should also have `os_and_command:` structure?

.github/workflows/actions.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         include:
         # run rubocop against lowest supported ruby
         - os: ubuntu-latest
-          ruby: '2.5'
+          ruby: '2.7'
```
dhstewart, line 30: I doubt this fixes the CI build error but just noticed it as I was trying to debug.

cben:
So far all changes LGTM :+1:

.github/workflows/actions.yml
dhstewart: This does seem to have fixed the error:

>Error when evaluating 'runs-on' for job 'build'. .github/workflows/actions.yml (Line: 14, Col: 14): Unexpected type of value '', expected type: OneOf.


But now there's an error saying the windows truffle ruby download was a 404.

https://github.com/ManageIQ/kubeclient/actions/runs/3740153807/jobs/6348284836

>Error: Error: Unexpected HTTP response: 404
    at D:\a\_actions\ruby\setup-ruby\v1\dist\index.js:8075:25
    at Generator.next (<anonymous>)
    at fulfilled (D:\a\_actions\ruby\setup-ruby\v1\dist\index.js:7985:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

Not sure how to fix this one but doing a little research to see.

cben: That's ok, Truffleruby doesn't exist for windows.
I've tried to exclude it from the matrix but gave up for now.

cben:
released 4.11.0

PR #590, opened 2022-12-20
Update actions/checkout

dhstewart:
### Why
Related to https://github.com/ManageIQ/kubeclient/issues/582 

We're getting deprecation warnings in CI due to node12.

### What
Moving to [actions/checkout@v3](https://github.com/actions/checkout#whats-new) switches us from node12 to the node16 runtime default.


cben:
Thanks!

PR #591, opened 2022-12-20
Update googleauth dependency

dhstewart:
### Why

I did some research on this issue https://github.com/ManageIQ/kubeclient/issues/564

It looks like to get openid connect up to date we'll need to get to Faraday v2, but our current version of googleauth constrains below that.

So it seems that getting the googleauth dependency into v1 is a necessary first step to addressing the issue.

### What
- update the googleauth dependency (here's the [the changelog](https://github.com/googleapis/google-auth-library-ruby/blob/main/CHANGELOG.md))
- ran tests locally, all passed 
 
If a more incremental approach to dependency updates are desired, I'm happy to approach this differently. 

Let me know what you think 👍

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'minitest-rg'
   spec.add_development_dependency 'webmock', '~> 3.0'
   spec.add_development_dependency 'rubocop', '~> 1.3.0' # locked to minor so new cops don't slip in
-  spec.add_development_dependency 'googleauth', '~> 0.5'
+  spec.add_development_dependency 'googleauth', '~> 1.3'
   spec.add_development_dependency('mocha', '~> 1.5')
   spec.add_development_dependency 'openid_connect', '~> 1.1'
   spec.add_development_dependency 'net-smtp'


cben:
FYI, test_informer.rb is known flaky, we have a tracking issue for that, but for now its failures are not merge blockers.

cben:
Thanks!  Sorry, forgot this for no good reason :-(

LGTM and reading their changelog, sounds "probably non-breaking".  But I really don't know almost anything about google auth, and that's some ~30 releases they've had (most before adopting semver)...  
@dhstewart have you done any actual testing against GCP?

@lucasmazza @timothysmith0609 @jeremywadsack can any of you comment on the safety of this upgrade, and/or test it?

jeremywadsack:
@cben The latest version of `googleauth` we are using in production is 0.17.1 right now and that works fine with `kubeclient`. I won't have time to test this in a real GKE cluster but that's certainly something that I'd expect before merging this, especially with a major version change. 

@dhstewart If you just need Faraday v2, can you use the spec `'~> 1.1', '>= 1.1.2'`? That way it's not forcing potential updates on other libraries that might be pinned below 1.3 (for whatever reason).

hmcguire-shopify:
> can any of you comment on the safety of this upgrade, and/or test it?

We've been using `googleauth` 1.3.0 with the master branch of Kubeclient and haven't seen any issues.

cben:
I think I'm gonna merge this and see if anybody complains.  Could relax the constaint to 1.1+ later if needed.

#606 is going towards some "how exactly do these libraries cache/expire tokens" questions, and reliably answering that for a wide range of versions is harder; if we can get away with only supporting recent ones that's simpler :sweat_smile:

PR #593, opened 2022-12-22
use forking-test-runner to ensure no test pollution between different…

grosser:
… test files

maybe this cuts down on the random errors
... and even if not it should reduce wild goose chases if one test messes with global state

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,11 +1,20 @@
 # frozen_string_literal: true
 
 require 'bundler/gem_tasks'
-require 'rake/testtask'
 require 'rubocop/rake_task'
 require 'yaml'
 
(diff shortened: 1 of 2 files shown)


grosser:
welp ... truffle-ruby does not have fork 😞

grosser:
fixed that ... still errors seems to be 98 and 135 mostly :/

grosser:
fixed the errors in https://github.com/ManageIQ/kubeclient/pull/594 ... this PR needs a rebase once that's merged

cben:
All green now with #594.  You still consider this desired, right?

Oh, it's your gem.  Looks very nice! :clap:

PR #594, opened 2022-12-22
improve informer test

grosser:
making all the threads terminate and no longer killing things ... saw 1 green run already 🤞


lib/kubeclient/informer.rb
```diff
@@ -88,6 +101,13 @@ def watch_to_update_cache
         @watching.each { |q| q << notice }
       end
       @logger&.info("watch restarted: #{stop_reason}")
+
+      # wake the waiter unless it's dead so it does not hang around
+      begin
+        @waiter.run
+      rescue ThreadError # rubocop:disable Lint/SuppressedException
+      end
+      @waiter.join
```
cben, line 110: Oh I think I see — previously, leftover waiter thread(s) could execute `@watcher.finish` at any time, possibly interrupting unrelated watchers from later tests, right?

But if we somehow (e.g. 'ERROR') come here early during `sleep(@reconcile_timeout)`, we might block for a long time — default 15min right?  
Here, what we're potentially blocking is the existing worker thread's loop, which won't be directly felt by the app but can cause significant gaps in the watching?

lib/kubeclient/informer.rb
```diff
@@ -21,22 +22,34 @@ def watch(&block)
...
     def stop_worker
-      @worker&.kill # TODO: be nicer ?
+      @stopped = true
+      [@waiter, @worker].compact.each do |thread|
+        begin
+          thread.run # cancel sleep so either the loop sleep or the timeout sleep are interrupted
+        rescue ThreadError
+          # thread was already dead
+        end
+        thread.join
```
cben, line 51: Similar question about `@waiter` possibly blocking for many minutes.  
Here it'd block the app calling `stop_worker` right?

- I'm thinking there is a subordinate relationship between worker thread -> single `watch_to_update_cache` run -> waiter thread. :thinking:  
  What if waiter threads did not refer to current instance variable `@watcher.finish` but closed over a lexically scoped reference to _their_ watcher?  Would then it be safe to leak unfinished waiter threads without .join-ing them?  Would it be a good idea?

- Waiter threads don't do much — are they safe to .kill?  Or is there another way to interrupt a sleep() early?

cben:
@agrare I don't know if you have been following the Kubeclient::Informer class on master branch, but perhaps it's interesting for future use in manageiq (?).  Either way, you may have useful experience to share here from how manageiq is killing & restarting watches...  That's a long way of saying "please review" :wink: 

[P.S. entirely out-of-scope here, but there is the unfinished business of #488.  The blocker there was we didn't know how to implement watch `.finish` with Faraday :shrug:  @agrare I was considering maybe dropping `.finish` in 5.0 and letting you figure out how to bring it back when you want to upgrade ManageIQ :stuck_out_tongue_closed_eyes:  But now I realize `Informer` relies on it deeply, and I suspect so would any advanced usage of watches?]

lib/kubeclient/informer.rb
agrare: > Or is there another way to interrupt a sleep() early?

You can use `Concurrent::Event` with a wait timeout to implement an interruptible sleep.

lib/kubeclient/informer.rb
grosser: it would have killed a random new waiter before (not really in tests since they all build their own informer)

added a `Thread.pass` to fix the race condition (don't think it really happens since the watch is a http request so it will take some time, but it's cheap so 🤷 )

lib/kubeclient/informer.rb
grosser: before we killed them, but that made the tests brittle, so I like the joining since that makes anything that goes wrong more obvious (and re-raises exceptions too)

grosser:
I think this is a good step forward, so prefer to merge and then iterate further if there are still open issues.

cben:
Sorry, missed your reply entirely.

FWIW, `Thread.pass` is advisory, so not sure it can really "fix" things:

> Give the thread scheduler a hint to pass execution to another thread. A running thread may or may not switch, it depends on OS and processor.

I'm still somewhat concerned a call to `stop_worker` might get stuck for a long time.  
But yes, previous behavior was chaotic, so I guess it's an advance. :+1:

And I'm certainly happy about CI looking reliably green now :tada:  Merging.

lib/kubeclient/informer.rb
```diff
@@ -21,22 +22,34 @@ def watch(&block)
...
+          begin
+            fill_cache
+            watch_to_update_cache
+          rescue StandardError => e
+            # need to keep retrying since we work in the background
+            @logger&.error("ignoring error during background work #{e}")
+          ensure
+            sleep(1) # do not overwhelm the api-server if we are somehow broken
+          end
+          break if @stopped
```
DocX, line 37: why not change the `loop` to `until @stopped` then?

lib/kubeclient/informer.rb
DocX: From my testing `@watcher.each` can get blocked on the response stream. Even closing the `http_client` will not unblock the response body stream.

In that case `join` on the waiter thread that runs the `@watcher.each` internally will never return

PR #597, opened 2023-01-04
fix: handle informer.stop_worker race conditions

DocX:
Testing the new informer revealed few edge case race condition resulting in deadlock.

When `stop_worker` is called immediately after `start_worker`, for example when `list` is evaluated as
there is no need for further watching (e.g. all jobs are completed). This can result in that
the original `stop_worker` would "resume" the thread that is supposed to call "finish" on the watcher. However it either is called when the thread is not yet created, or after the thread is created, but before the `watcher.each`
has been entered.

This then results in that the `watch_to_update_cache` would never return (because the watcher stop attempt was
abandoned before the watch started). And this would block the "worker.join" call.

This PR adds tests for those race conditions and implements fixes

### DIFF
--- a/lib/kubeclient/informer.rb
+++ b/lib/kubeclient/informer.rb
@@ -25,32 +25,26 @@ module Kubeclient
     def start_worker
       @stopped = false
       @worker = Thread.new do
-        loop do
+        until @stopped
           begin
             fill_cache
-            watch_to_update_cache
+            stop_reason = watch_to_update_cache
+            @logger&.info("watch restarted: #{stop_reason}")
           rescue StandardError => e
             # need to keep retrying since we work in the background
             @logger&.error("ignoring error during background work #{e}")
           ensure
             sleep(1) # do not overwhelm the api-server if we are somehow broken
           end
-          break if @stopped
         end
       end
       sleep(0.01) until @cache || @stopped
     end
 
     def stop_worker
-      @stopped = true
-      [@waiter, @worker].compact.each do |thread|
-        begin
-          thread.run # cancel sleep so either the loop sleep or the timeout sleep are interrupted
-        rescue ThreadError
-          # thread was already dead
-        end
-        thread.join
-      end
+      @stopped = true # mark that any threads should be stopped if it is running
+      pass_and_run(@worker) # skip the end loop sleep
+      @worker.join
     end
 
     private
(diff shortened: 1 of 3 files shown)


lib/kubeclient/informer.rb
```diff
@@ -34,22 +34,20 @@ def start_worker
...
-        begin
-          thread.run # cancel sleep so either the loop sleep or the timeout sleep are interrupted
-        rescue ThreadError
-          # thread was already dead
-        end
-        thread.join
+      @stopped = true # mark that any threads should be stopped if it is running
+      begin
+        Thread.pass
+        @worker.run # interrupt potential loop sleep
```
DocX, line 46: This should not be necessary, since at most it will avoid the `sleep 1` in the loop. Condition on `@stopped` is suficient.

lib/kubeclient/informer.rb
```diff
@@ -78,37 +78,45 @@ def fill_cache
...
-        Thread.pass # make sure we get into the sleep state of the waiter
-        @waiter.run
-      rescue ThreadError # rubocop:disable Lint/SuppressedException
+    def watcher_with_timeout
+      @watcher = @client.watch_entities(@resource_name, watch: true, resource_version: @started)
+      timeout_deadline = Time.now + @reconcile_timeout
+      watcher_finished = false
+
+      finisher_thread = Thread.new do
+        sleep 0.5 until @stopped || watcher_finished || Time.now > timeout_deadline
```
DocX, line 107: instead of using sleep as the condition for continuing (and using run as signalling), evaluate the conditions explicitly. this allows to wait for any of the different conditions - either stopped is triggered, the method ended, or the timeout expired.

test/test_watch.rb
```diff
@@ -238,4 +238,82 @@ def test_watch_with_finish_and_ebadf
 
     assert_requested(:get, "#{api_host}/v1/watch/events", times: 1)
   end
+
+  def test_watch_finish_when_response_connection_open
```
DocX, line 242: This is actually unrelated to the issue. I added it as that was my initial guess on where the issue is. Maybe it is still useful to have it here? But happy to revert this file changes.

cben:
cc @grosser

lib/kubeclient/informer.rb
```diff
@@ -24,32 +24,32 @@ def watch(&block)
...
       @stopped = false
       @worker = Thread.new do
-        loop do
+        until @stopped do
           begin
             fill_cache
-            watch_to_update_cache
+            watch_to_update_cache.tap do |stop_reason|
+              @logger&.info("watch restarted: #{stop_reason}")
+            end
```
grosser, line 32:
```suggestion
            stop_reason = watch_to_update_cache
            @logger&.info("watch restarted: #{stop_reason}")
```

lib/kubeclient/informer.rb
grosser: avoiding a `sleep 1` could be useful to make the tests finish quickly ... unless they stub all sleeps

test/test_watch.rb
grosser: good to test weird edge-cases 👍

grosser (approved):
extracting watcher_with_timeout is great :)

grosser (approved):

lib/kubeclient/informer.rb
```diff
@@ -78,37 +72,53 @@ def fill_cache
...
-      @waiter = Thread.new do
-        sleep(@reconcile_timeout)
-        stop_reason = 'reconcile'
+      finisher_thread = Thread.new do
+        sleep(0.5) until @stopped || watcher_finished || Time.now > timeout_deadline
+        # loop calling finish until the actual method has
+        # exited, since watcher.each may be called after the
+        # finish in this thread is called
         @watcher.finish
+        @watcher.finish until watcher_finished
```
Mins, line 106: might want to add a small delay here

cben:
This is excellent!  Good refactor, good tests, very pragmatic `sleep 0.5 until cond1 || cond2...` approach :clap:

However, here is a relevant failure from CI https://github.com/ManageIQ/kubeclient/actions/runs/3845504254/jobs/6706354722#step:6:166 :
```
1) Failure:
TestInformer#test_restarts_on_error [test/test_informer.rb:96]:
[0.01, 1, 1]
```

lib/kubeclient/informer.rb
cben: Indeed, might even let the watch thread exit faster than busy-waiting?  Even if not, it's polite to other threads & processes to limit how much CPU this can eat.

DocX:
@cben thanks. the error looks like it was flake, however i've updated the unit test code anyway to be more robust and less time-dependent. hopefully it should help.

cben:
Fix makes sense :+1:, don't care how many times exactly Informer called sleep(), especially when we're stubbing these sleeps to be short.  LGTM :100: 

Please resolve conflict, I just merged #598.

cben (approved):

DocX:
@cben thanks for merging the other PR. resolved the conflict and rebased the commits to one.

PR #598, opened 2023-01-05
feat: allow passing options to Informer

DocX:
Adding option to pass `options` to Informer so it is possible to make the get and watch requests with additional parameters
like `namespace` or `labelSelector`

### DIFF
--- a/lib/kubeclient/informer.rb
+++ b/lib/kubeclient/informer.rb
@@ -1,7 +1,7 @@
 module Kubeclient
   # caches results for multiple consumers to share and keeps them updated with a watch
   class Informer
-    def initialize(client, resource_name, reconcile_timeout: 15 * 60, logger: nil)
+    def initialize(client, resource_name, options: {}, reconcile_timeout: 15 * 60, logger: nil)
       @client = client
       @resource_name = resource_name
       @reconcile_timeout = reconcile_timeout
@@ -10,6 +10,7 @@ module Kubeclient
       @started = nil
       @stopped = false
       @watching = []
+      @options = options
     end
 
     def list
@@ -70,7 +71,8 @@ module Kubeclient
     end
 
     def fill_cache
-      reply = @client.get_entities(nil, @resource_name, raw: true, resource_version: '0')
+      get_options = @options.merge(raw: true, resource_version: '0')
+      reply = @client.get_entities(nil, @resource_name, get_options)
       @cache = reply[:items].each_with_object({}) do |item, h|
         h[cache_key(item)] = item
       end
@@ -78,7 +80,8 @@ module Kubeclient
     end
 
     def watch_to_update_cache
-      @watcher = @client.watch_entities(@resource_name, watch: true, resource_version: @started)
+      watch_options = @options.merge(watch: true, resource_version: @started)
+      @watcher = @client.watch_entities(@resource_name, watch_options)
       stop_reason = 'disconnect'
 
       # stop watcher without using timeout
--- a/README.md
+++ b/README.md
@@ -893,6 +893,19 @@ informer.list # all current pods
 informer.watch { |notice|  } # watch for changes (hides restarts and errors)
 ```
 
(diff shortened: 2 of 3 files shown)


grosser (approved):
🎉

lib/kubeclient/informer.rb
```diff
@@ -1,7 +1,7 @@
 module Kubeclient
   # caches results for multiple consumers to share and keeps them updated with a watch
   class Informer
-    def initialize(client, resource_name, reconcile_timeout: 15 * 60, logger: nil)
+    def initialize(client, resource_name, options: {}, reconcile_timeout: 15 * 60, logger: nil)
```
Mins, line 4: What do you think about splitting this to 2 options? It seems that the `watch` and the `get` APIs accept different options

DocX: Seems overkill. I would leave it this way until there is strong use case for different options. Options that control the result of the requests are same.

lib/kubeclient/informer.rb
```diff
@@ -70,15 +71,17 @@ def cache_key(resource)
...
+      reply = @client.get_entities(nil, @resource_name, get_options)
       @cache = reply[:items].each_with_object({}) do |item, h|
         h[cache_key(item)] = item
       end
       @started = reply.dig(:metadata, :resourceVersion)
     end
 
     def watch_to_update_cache
-      @watcher = @client.watch_entities(@resource_name, watch: true, resource_version: @started)
+      watch_options = @options.merge(watch: true, resource_version: @started)
```
cben, line 83: Off-topic, but I wonder if `watch: true` does anything.  I don't see `watch_entities` looking at it.  @grosser do you remember why you wrote that in the first place?

lib/kubeclient/informer.rb
```diff
@@ -70,15 +71,17 @@ def cache_key(resource)
     end
 
     def fill_cache
-      reply = @client.get_entities(nil, @resource_name, raw: true, resource_version: '0')
+      get_options = @options.merge(raw: true, resource_version: '0')
```
cben, line 74: @grosser similar question on the origin of `raw: true` here, I don't see it doing anything.
Did you intend `as: :raw`?  Though that means totally unparsed string, I suspect you want `as: :parsed_symbolized` and the current code works thanks to default RecursiveOpenStruct's leniency (?)

lib/kubeclient/informer.rb
cben: A good question, let's see...  
#### Shared options:

- :namespace (string)
- :label_selector (string)
- :field_selector (string)
- :resource_version (string) — but informer overrides this in both get & watch
- :as — informer might break if you set this in `options`, or if you've customized `client`'s `@as`?  Probably informer SHOULD override this?  However should do `format_response` processing in `@watching.each { |q| q << notice }`?  Or should it be an error?  => You know what, I'll offer a separate issue about this.

#### Get-only options:

- :limit (integer)
- :continue (string) — these two are maybe worth using inside `fill_cache` (to limit apiserver RAM usage on huge collections), but make no sense to set externally.

#### Watch-only options:

- :allow_watch_bookmarks (bool) — Current informer code would log "Unsupported event type" but wouldn't benefit.  Would be nice to use internally by informer, but if we do that I think it can be unconditional, makes no sense to set externally IMHO.
- :name (string) — Informer does not support this currently, would need to switch to `get_entity`.  But that's implementation detail; if we support this, I wouldn't want caller to have to pass name in two different ways, a shared `options` interface with special treatment for `:name` remains good.

=> So I think @DocX is right here.

cben (approved):

cben:
LGTM.  Asked 2 off-topic questions here while we're looking at this code, but dropping those `watch: true` and `raw: true` — if indeed they do nothing — wouldn't belong here.  Merging.

grosser:
both just copy-paste afaik ... if the tests still work then it can go away :)

PR #600, opened 2023-01-20
support impersonation via auth_options and in kubectl config

DocX:
This PR is rebase and update of https://github.com/ManageIQ/kubeclient/pull/524 to current master version

We tested it and it works as expected

Ruby http clients (including adapters for Faraday) are not able to handle multiple headers of same name, which is how the kube API expects to receive the list of multiple groups or extra fields. To avoid any confusion, added explicit error when trying to configure client with multivalue groups. **This may brake some applications, if user is not aware of the impersonation setting and were ignoring it until now**

### DIFF
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -137,6 +137,7 @@ module Kubeclient
       @as = as
 
       validate_bearer_token_file
+      configure_impersonation_headers
     end
 
     def configure_faraday(&block)
@@ -742,6 +743,21 @@ module Kubeclient
       return unless (file = @auth_options[:bearer_token_file])
       raise ArgumentError, "Token file #{file} does not exist" unless File.file?(file)
       raise ArgumentError, "Token file #{file} cannot be read" unless File.readable?(file)
+    end
+
+    # following https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation
+    def configure_impersonation_headers
+      return unless (auth_as = @auth_options[:as])
+      @headers[:'Impersonate-User'] = auth_as
+      if (auth_as_groups = @auth_options[:as_groups])
+        @headers[:'Impersonate-Group'] = Array(auth_as_groups).join
+      end
+      if (auth_as_uid = @auth_options[:as_uid])
+        @headers[:'Impersonate-Uid'] = auth_as_uid
+      end
+      @auth_options[:as_user_extra]&.each do |k, v|
+        @headers[:"Impersonate-Extra-#{k}"] = Array(v).join
+      end
     end
 
     def return_or_yield_to_watcher(watcher, &block)
--- a/lib/kubeclient.rb
+++ b/lib/kubeclient.rb
@@ -750,14 +750,18 @@ module Kubeclient
     def configure_impersonation_headers
       return unless (auth_as = @auth_options[:as])
       @headers[:'Impersonate-User'] = auth_as
-      if (auth_as_groups = @auth_options[:as_groups])
-        @headers[:'Impersonate-Group'] = Array(auth_as_groups).join
+      if (as_groups = @auth_options[:as_groups])
+        # Faraday joins multi-value headers with commas, which is not same as having
+        # multiple headers with the same name, as required by the k8s API
+        raise ArgumentError, 'Multiple as_groups are not supported' if as_groups.count > 1
+        @headers[:'Impersonate-Group'] = as_groups[0]
       end
-      if (auth_as_uid = @auth_options[:as_uid])
-        @headers[:'Impersonate-Uid'] = auth_as_uid
+      if (as_uid = @auth_options[:as_uid])
+        @headers[:'Impersonate-Uid'] = as_uid
       end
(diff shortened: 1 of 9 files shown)


lib/kubeclient.rb
```diff
@@ -744,6 +746,25 @@ def validate_bearer_token_file
...
+        # Faraday joins multi-value headers with commas, which is not same as having
+        # multiple headers with same name, as required by the k8s API
+        raise ArgumentError, "Kubeclient does not support multiple impersonate groups" if auth_as_groups.count > 1
+        @headers[:'Impersonate-Group'] = auth_as_groups[0]
+      end
+      if (auth_as_uid = @auth_options[:as_uid])
+        @headers[:'Impersonate-Uid'] = auth_as_uid
+      end
+      @auth_options[:as_user_extra]&.each do |k, v|
+        raise ArgumentError, "Kubeclient does not support impesonate extra field with multiple values" if v.count > 1
```
DocX, line 763: @cben what do you think about this error? this may introduce regression on users side. Should we print just some warning instead and fallback to trimming arrays with > 1 value to just the first value?

DocX: Just for context, as mentioned in the PR summary, this is because Faraday does not support multiple headers with same name, however the Kube API is expecting those to be requested as duplicate headers (with same name, and each different value). It does not recognize joining the value as comma separated lists.

cben: yikes.  doesn't [the RFC](https://www.rfc-editor.org/rfc/rfc9110#section-5.3) require "aside from the well-known exception noted below, a sender MUST NOT generate multiple field lines with the same name in a message (whether in the headers or trailers) or append a field line when a field line of the same name already exists in the message, unless that field's definition allows multiple field line values to be recombined as a comma-separated list" ?  (with the grandfathered exception being `Set-Cookie`)

Is there any plan to fix that in kubernetes API?  (not that this helps us much here)

Is this supported better in Faraday 2?

DocX: @cben I've opened [issue](https://github.com/lostisland/faraday/issues/1486) in faraday and actually it is down at the http libraries. Neither ruby's `net/http` nor `exconn` supports that. Maybe other client library does (my tip would be typhoeues, but did not test). Anyway I guess as kubeclient we don't want to impose the faraday adapter used on the user? Since user may set global adapter for Faraday that would also be used by kubeclient.

So I think there is nothing to do, except implementing own client? Or maybe forcing specific faraday adapter in kubeclient requests.

DocX: There is also some [issue in kubernetes](https://github.com/kubernetes/kubernetes/issues/82468) around this problem, from using http proxies, that often reconstruct headers by joining them with comma, which breaks the semantic. But no fix seems to be on the way.

cben: I see, thanks for the pointers.  It's been open for several years, and even if k8s add a new interface in future, it's useful for kubeclient to be able to function against existing k8s versions. :+1:

Compatibility on existing kubeconfig files is a good question, but given that previous behavior was a bug (impersonation settings were ignored), I'm not worried much about it.

Also, we're in the weird state that `master` branch has breaking changes since 4.y releases, but no 5.y releases were made yet, sparing the question of how exactly to semver it :wink:

* I like that `Config` supports multiple values but only `Client` will complain :heavy_check_mark:

cben:
- [ ] Please add an entry in CHANGELOG.md.  We grew a huge "changelog debt" :-(, which I hope to make up one day, but since this _might_ break somebody best document it right away.

## What about watches?
Watches still use https://github.com/httprb/http gem (#488 having some unresolved issues).  
Does this implementation affect the headers we send there too?  I _think_ so, `http_options` method takes `@headers`.
- [ ] Please add a unit test for watch.  As long as we have 2 easy-to-diverge implementations, I prefer not to make HTTP-level changes without test coverage for both.

Otherwise LGTM :rocket:

lib/kubeclient.rb
cben: > Should we print just some warning instead and fallback to trimming arrays with > 1 value to just the first value?

Not now.  It would be incorrect & result will depend on order listed in kubeconfig.  
If we do this invisibly, people may keep depending on it, making it hard to revert such a workaround...

I prefer merging with it correctly raising an error, and waiting for people to complain.  _Then_ they are in position to change their kubeconfigs, OR add custom trimming in their code between `Config` and `Client` — OR convince us the hassle is sufficient to bake the  trimming kludge into kubeclient...

cben (approved):
Excellent :100:

DocX:
Thanks @cben. Is there any plan to publish new release with the update?

cben:
I need to make progress on my 5.0 changelog backlog to get master branch releasable...
I'm sorry, I can't give time estimate. But nagging is welcome ;-|

Alternatively, do you want to work on backporting this to `v4.y` branch which uses rest-client?
It'd require some changes, not sure it's worth the effort.

DocX:
Ok, no problem. Just asking. It is not super critical for us, so we can wait (and we can always install gem from git if we need to)

PR #606, opened 2023-02-28
Enable refreshing of auth_provider bearer_token

hmcguire-shopify:
Ref: #393

Previously, once a Kubeclient::Context was created, auth_provider bearer_tokens would be set and never refreshed.

This commit enables automatic refreshing of auth_provider bearer_tokens by returning callables instead of static tokens. When used with the faraday client, the callable is automatically called for each request, and for the non-faraday client the callable is called before setting the header manually.

Tests have been updated to cover this functionality for both client types, and an additional test was added for refreshing bearer_token_file because it did not appear to be covered before.

### DIFF
--- a/lib/kubeclient/config.rb
+++ b/lib/kubeclient/config.rb
@@ -199,9 +199,9 @@ module Kubeclient
       case auth_provider['name']
       when 'gcp'
         config = expand_command_option(auth_provider['config'], 'cmd-path')
-        Kubeclient::GCPAuthProvider.token(config)
+        -> { Kubeclient::GCPAuthProvider.token(config) }
       when 'oidc'
-        Kubeclient::OIDCAuthProvider.token(auth_provider['config'])
+        -> { Kubeclient::OIDCAuthProvider.token(auth_provider['config']) }
       end
     end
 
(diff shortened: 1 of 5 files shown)


hmcguire-shopify:
Test failures appear to be because SSL certs need to be regenerated

cben:
cc @benlangfeld @jeremywadsack @timothysmith0609 @nov @dhstewart: I dumped some questions about this PR above, but you all know more about auth than I do, please help :))

With auth knowledge (and ability to test) being fragmented by provider, it's hard to make progress...  
I don't want my "are we covering all use cases?" perspective to stall this but I suspect it's time to break out of the existing API mold.

----

Ooh one more point: current README recommends renewal by re-creating _everything_, even a new `Config` instance :fire:.  (No idea if users actually followed this extreme advice.)  I pushed for that in name of "future flexibility", as I didn't want to commit to things like "every `.context` calls obtains new auth".  But now it occurs to me that's very bad wrt. #408 because it leaves no objects where refresh tokens from previous time could be saved :worried: 

* From OOP perspective, I'd hate to introduce some kind of invisible global cache; my experience is that all libraries starting from that eventually need to give users some kind of explicit "Session" objects or other caching controls...
* But the big picture goes beyond OOP: some providers already cache tokens outside our process!  And IIUC the ultimate goal of #408 is persisting back to the kubeconfig file (same way `kubectl` does)?

[I've been saying "caching" but that's imprecise word for "keep this and don't come back without it"]

PR #614, opened 2023-05-15
Add test coverage for Ruby 3.2

dosuken123:
GitLab is planning to upgrade ruby version to 3.1 first, and later to 3.2. For the preparation, we're checking if dependencies are compatible with these versions. (See https://gitlab.com/gitlab-org/gitlab/-/issues/404750 for more information)

This PR adds the test coverage for Ruby 3.2.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os_and_command.os }}
     strategy:
       matrix:
-        ruby: [ '2.7', '3.0', '3.1', 'ruby-head', 'truffleruby-head' ]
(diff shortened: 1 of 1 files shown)


cben:
Thanks!  Also cherry-picked to v4.y branch (#615) from which releases are made, and 3.2 passed there.  

Last [4.11.0 release](https://rubygems.org/gems/kubeclient/versions/4.11.0) has "Required Ruby Version: >= 2.7.0", no upper limit, so I assume no new release is needed at this point?

dosuken123:
@cben Thanks for quick merge and cherry-picking it to v4! Yes, I think we don't need a new release.

PR #630, opened 2024-03-27
Assume role support for aws eks credentials

alauddinulaganathan:
1. Describe what this Pull Request does and why you think it is needed.
   We are using this kubeclient to perform API operations against the AWS EKS (kubernetes) clusters (i.e. the Kubernetes API Server). But we are not using access_key & secret_key directly as it is generally considered insecure. Hence we use - role based access control along with AWS STS to perform authentication (i.e. short lived temporary credentials). But the support for such Credentials Provider is not there in this Client (only supports - static Credentials & profile based static credentials). 

I tried a small change to support the AssumeRoleCredentials Provider and it worked nicely. Hence I thought this would be a nice feature enhancement for this repo to users like me who want to use role based access with STS assumed role credentials.

2. If this fixes an existing issue, please specify in `Fixes #<id>` format
   Not a bug fix. It is a feature enhancement.


alauddinulaganathan:
@miq-bot add-label enhancement
@miq-bot add-label aws
@miq-bot add-label eks
@miq-bot add-label aws_eks

README.md
```diff
@@ -331,9 +331,19 @@ credentials = Aws::Credentials.new(access_key, secret_key)
 # Or a profile
 credentials = Aws::SharedCredentials.new(profile_name: 'default').credentials
 
+# Auth Options for a direct credential
 auth_options = {
   bearer_token: Kubeclient::AmazonEksCredentials.token(credentials, eks_cluster_name)
 }
+
+# Or for an STS Assumed Role Credentials or any other credential Provider other than Static Credentials
+credentials = Aws::AssumeRoleCredentials.new({ client: sts_client, role_arn: role_arn, role_session_name: session-name })
```
agrare, line 340: Really minor:
```suggestion
credentials = Aws::AssumeRoleCredentials.new({ client: sts_client, role_arn: role_arn, role_session_name: session_name })
```

lib/kubeclient/aws_eks_credentials.rb
```diff
@@ -20,18 +20,26 @@ def token(credentials, eks_cluster, region: 'us-east-1')
         end
         # https://github.com/aws/aws-sdk-ruby/pull/1848
         # Get a signer
-        signer = Aws::Sigv4::Signer.new(
-          service: 'sts',
-          region: region,
-          credentials: credentials
-        )
+        if is_credentials_provider
```
agrare, line 23: From `Aws::Sigv4::Signer`:

> A credential provider is any object that responds to `#credentials`


This got me thinking, what if instead of adding another parameter we did
```suggestion
        if credentials.respond_to?(:credentials)
```
just to make it a little easier on the end user.
Unfortunately AWS use modules rather than a `Aws::CredentialProvider` base class so we can't check `kind_of?`.

alauddinulaganathan: Nice Suggestion. Modified and committed the suggested changes and updated the readme file accordingly.

README.md
alauddinulaganathan: Committed the suggestion.

agrare (approved):
LGTM, test failures look unrelated to this change I'll fix those in another PR

agrare:
Test failures should be resolved by #631 and #632

PR #632, opened 2024-04-02
Regenerated expired test TLS certs

agrare:
Regenerate expired TLS certs by running `test/config/update_certs_k0s.rb`


agrare:
@Fryguy recommend merging https://github.com/ManageIQ/kubeclient/pull/631 on red to fix the bundler version issue, then I'll rebase this one and it should go green.

agrare:
Okay now only the 2.7 tests are failing which will be fixed by https://github.com/ManageIQ/kubeclient/pull/631

kbrock (approved):
thanks for the change

LGTM

agrare:
Thanks @Fryguy !

PR #640, opened 2024-10-25
Try bumping recursive-open-struct

smoser:
recursive-open-struct recently had a 2.0 release which included an ABI breaking change.  I'm not sure if the change affects kubeclient or not.  If not, it would be nice to be able to use the newer release.

I have only tried import test, I was hoping that c-i would give a better sniff.

https://github.com/wolfi-dev/os/issues/31938

### DIFF
--- a/kubeclient.gemspec
+++ b/kubeclient.gemspec
@@ -34,6 +34,6 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'faraday', '>= 1.1', '< 3.0'
   spec.add_dependency 'faraday-follow_redirects', '>= 0.3.0'
-  spec.add_dependency 'recursive-open-struct', '~> 1.1', '>= 1.1.1'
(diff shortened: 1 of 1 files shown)


xnox:
@smoser I see workflows awaiting to run. Maybe you can turn on actions in your fork, modify workflow file and manually trigger them there? Not sure if any of them are possible to run in forks without secrets.

smoser:
> @smoser I see workflows awaiting to run. Maybe you can turn on actions in your fork, modify workflow file and manually trigger them there? Not sure if any of them are possible to run in forks without secrets.

We'll see how it does: https://github.com/smoser/kubeclient/actions/runs/11525469999

agrare (approved):
I don't see us using marshal_dump anywhere, and I've run a couple of tests against our vcr cassettes for a kubernetes and an openshift cluster and they all pass.

:+1: LGTM thanks @smoser !

PR #671, opened 2025-10-15
[v4.y] CI: update certs + add ruby 3.3, 3.4

cben:
I noticed v4.y branch has a big gap between 3.2 and "head" versions (which are already at 3.5 prerelease).  
Master branch already covers 3.3 and 3.4.


grosser:
can we remove ruby-head for now to get it green and then open another PR to bring it back ?

agrare (approved):

PR #675, opened 2025-10-15
[v4.y] Exclude windows-latest from truffleruby testing

agrare:
Truffleruby doesn't support windows so drop it from the test matrix.

Fixes: https://github.com/ManageIQ/kubeclient/actions/runs/18524932757/job/52793370757
`Error: Error: TruffleRuby does not currently support Windows.`

Re implementation of https://github.com/ManageIQ/kubeclient/pull/673


agrare:
This has been released in v4.13.0
https://github.com/ManageIQ/kubeclient/releases/tag/v4.13.0

PR #677, opened 2025-10-29
feat: parse kubeconfig for bearer_token_file

jperville:
As an user of the kubeclient gem
I want to configure my Kubeconfig::Client automatically to use a tokenFile from kubeconfig
So that my token is automatically refreshed and my client does not expire.

This PR adds parsing for [tokenFile in the kubeconfig authinfo](https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#AuthInfo) and automatically sets `auth_config[:bearer_token_file]`  in the client library. tokenFile will be prioritary over token entries if both are found.

This feature makes the existing support for bearer_token_file (which allows for refreshing tokens) in this library easier to use.

### DIFF
--- a/test/config/userauth.kubeconfig
+++ b/test/config/userauth.kubeconfig
@@ -12,6 +12,11 @@ contexts:
 - context:
     cluster: localhost:6443
     namespace: default
+    user: system:admin:token-file
+  name: localhost/system:admin:token-file
+- context:
+    cluster: localhost:6443
+    namespace: default
     user: system:admin:userpass
   name: localhost/system:admin:userpass
 current-context: localhost/system:admin:token
@@ -21,6 +26,9 @@ users:
 - name: system:admin:token
   user:
     token: 0123456789ABCDEF0123456789ABCDEF
+- name: system:admin:token-file
+  user:
+    tokenFile: /path/to/secret/token
 - name: system:admin:userpass
   user:
     username: admin
--- a/lib/kubeclient/config.rb
+++ b/lib/kubeclient/config.rb
@@ -173,7 +173,9 @@ module Kubeclient
 
     def fetch_user_auth_options(user)
       options = {}
-      if user.key?('token')
+      if user.key?('tokenFile')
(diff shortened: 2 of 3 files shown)


jperville:
@Fryguy  could you have a look at this PR? thanks a lot for your time.

Fryguy:
@jperville Apologies for the delay - for some reason I never got a notifcation for this. @agrare Can you also take a look?

agrare:
I'm just going to kick the tests on this one since it has been a while

test/config/userauth.kubeconfig
```diff
@@ -21,6 +26,9 @@ users:
 - name: system:admin:token
   user:
     token: 0123456789ABCDEF0123456789ABCDEF
+- name: system:admin:token-file
+  user:
+    tokenFile: /path/to/secret/token
```
Fryguy, line 31: So strange that Kube decided to use CameCase for this key, but dash-cased for the rest 🤦  Verfied at https://kubernetes.io/docs/reference/config-api/kubeconfig.v1/#AuthInfo

agrare (approved):
Ran a live test against my CRC cluster and this works perfectly, LGTM

jperville:
Hello, just found about the exchange on my own PR.
Thanks a lot for merging it!

test/config/userauth.kubeconfig
jperville: I was surprised too, but the schema is the law :-)

cben:
Other fields referring to files supported:
1. relative paths (resolved relative to `kcfg_path`);
2. Disabling filesystem access by `Config.new(data, nil)`. This is a (limited but better than nothing) security feature for situations you don't trust the config.

Documented [in README](https://github.com/ManageIQ/kubeclient#kubeclientconfig).
3 relevant helpers: https://github.com/ManageIQ/kubeclient/blob/6d9baa8edeb835cbfa6033ef974925725bdf04e4/lib/kubeclient/config.rb#L92-L117

Does this PR preserve that? Should the `*_nopath` test cases be expanded?

